PR #10, opened 2018-07-10
Add curried function for display

remo5000:
Must have accidentally changed it when incorporating externalBuiltIns.

### Features
- Fixed display
- Added externalContext parameter for all externalBuiltIns
- Added alert, visualiseList function
- Fix typescript errors on exporting a function, but not exporting a related type that it uses

### DIFF
--- a/src/createContext.ts
+++ b/src/createContext.ts
@@ -118,13 +118,8 @@ export const importBuiltins = (context: Context, externalBuiltIns: CustomBuiltIn
     defineSymbol(context, 'timed', (f: Function) => misc.timed(context, f, context.externalContext, externalBuiltIns.display))
     // previously week 5
     defineSymbol(context, 'assoc', list.assoc)
-    if (window.hasOwnProperty('ListVisualizer')) {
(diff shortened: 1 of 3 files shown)


src/createContext.ts
```diff
@@ -60,7 +60,7 @@ export const importBuiltins = (context: Context, externalBuiltIns: CustomBuiltIn
 
   if (context.chapter >= 1) {
     defineSymbol(context, 'runtime', misc.runtime)
-    defineSymbol(context, 'display', externalBuiltIns.display)
+    defineSymbol(context, 'display', (v: Value) => externalBuiltIns.display(v, context.externalContext))
     defineSymbol(context, 'error', misc.error_message)
     defineSymbol(context, 'prompt', externalBuiltIns.prompt)
```
ningyuansg: Prompt should also be curried with `externalBuitIns`, iirc? It's definition is something like `promptString: string => void`

ningyuansg: `alert` should also be curried as well. It doesn't exist on the node environment, and so will cause a ReferenceError if a student submits a program with alert.

(github doesn't let me review a line that far away from a diff'd line?)

PR #23, opened 2018-08-01
Add Chapter 4

remo5000:
### Features
- Added some functions from the docs
- Imported some type checking functions from a previous repository
- Used external `parse` function from sourceacademy

### TODO
- [x] Add a `parse` function
- [x] Add the source for the `parse` function

Changing the grammar of the metacircular parser will be raised as an issue (#24 )


ningyuansg:
@remo5000 You gotta fix the tests

remo5000:
Tried a fix, let's see if it works for coveralls

ningyuansg:
You've tested the the js files from this new setting in 10e1fa9 works on frontend?

scripts/build.sh
```diff
@@ -0,0 +1,7 @@
+# Compile all .ts files
+tsc
```
ningyuansg: Do a `&&` so `cp` does not run if `tsc` fails.

ningyuansg:
Bump

remo5000:
Missed the notif, done! @ningyuansg

martin-henz:
I think some more work is needed for this. The new specs for Source §4 are here. 
[source_4.pdf](https://github.com/source-academy/js-slang/files/2431261/source_4.pdf)

According to Yuchuan, apply_in_underlying_javascript is missing from the implementation, see page 7 of specs.

PR #27, opened 2018-08-06
Move list visualiser to chapter 2

remo5000:


remo5000:
@ningyuansg Should we rename the function to `visualise_list`? I feel that `draw` is too similar to some external libraries

ningyuansg:
@remo5000 try `draw_list`. Also, you need to rebase.

ningyuansg:
@remo5000 Are you renaming this draw to draw_list or no?

remo5000:
@ningyuansg Missed this out, done! I've updated the PR on the frontend as well

PR #34, opened 2018-08-19
Add Source Week 1 as an internal reference for the Avengers

martin-henz:
My intention is to focus the Avengers and tutors on this language in our interactions with the students, until we introduce more constructs.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,3 +1,9 @@
+### Naming 
+- The official specifications for Source § X is `source_X.pdf`.
+- Other `.tex` files are used as imports within the specs.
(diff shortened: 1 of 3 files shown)


remo5000:
Hi Prof @martin-henz, is there any update on the `make all` behaviour? I'd like to merge this PR once the above query is resolved

remo5000:
Not sure about `make all` and `make install` (which requires a particular sunfire account access), but I'm able to make the individual file. Updated the README to prevent confusion when seeing source_X and source_week_X.

martin-henz:
Sorry, I missed these comments. Good points. Yes, let's include the Week 1 document in 
make all
but exclude it from 
make install

Can you please change the Makefile accordingly?

Thanks,

-martin

PR #42, opened 2018-08-27
four character indentation after arrows

martin-henz:
Not sure how you'll like this. Indentation is sometimes hotly debated. Best to decide something and then stick with it.


remo5000:
I saw a poll going around, is this a consequence of it?

martin-henz:
It's obsolete.

PR #54, opened 2018-09-01
Implemented hoisting

openorclose:
Implemented hoisting for ```let```, ```const``` and function declarations according to es6 spec (Except that redeclaration of functions is disallowed here unlike in es6). Also moved block scope logic from ```IfStatement``` to ```BlockStatement``` since ```{const a = 1;}``` should be block scoped anyway.


martin-henz:
I'd rather not discuss hoisting in class. If students come up with one of the weird hoisting examples, I'd prefer to say that it's not Source. I don't view discussions of hoisting as particularly useful in the context of CS1101S.

So let's restrict the syntax of Source as follows:

Local functions must precede all other statements in a function body.

openorclose:
If function declarations are not hoisted but variable declarations are, the below code would allow accessing of shadowed variables since x is a function. 

```javascript
let x = 4;
if(true) {
	print(x); //will print 4 if below function is not hoisted
	function x() {
            return 1;
        }
}
```

Would it be better if function declarations were hoisted but like `let` and `const`, where only the declaration gets hoisted but not the assignment? (So the code would result in a ReferenceError instead.)

martin-henz:
> If function declarations are not hoisted but variable declarations are, the below code would allow accessing of shadowed variables since x is a function.
> 
> ```js
> let x = 4;
> if(true) {
> 	print(x); //will print 4 if below function is not hoisted
> 	function x() {
>             return 1;
>         }
> }
> ```
> 
> Would it be better if function declarations were hoisted but like `let` and `const`, where only the declaration gets hoisted but not the assignment? (So the code would result in a ReferenceError instead.)

Yes, correct. This was a misunderstanding on my part. This issue is called the "temporal dead zone" and we have an example in the textbook (1.3.2, footnote 3, see https://www.comp.nus.edu.sg/~cs1101s/sicp/chapters/20).

We should "hoist" declarations. Actually, I hate the term "hoisting" in this context. A better term is "proper block scope of variables".

yyc:
@openorclose yes, sorry, I should've been clearer. That was what I meant when I said we shouldn't hoist definitions, only the declarations.

ningyuansg:
If I have followed the discussion correctly so far, it seems source no longer needs to be a subset of javascript, following some change in scoping (correct me if I'm wrong). Honestly, I think this is a good change. There are many messy aspects of JS that would not benefit someone learning programming for the first time.

*If* this is the case, I don't see why we must prevent disallowed accessing shadowed variables. It is undoubtedly bad practice, and certainly illegal in some programming languages, but also perfectly logical to the beginner who simply expects all statements to run sequentially from the first line of their program to the last. A student would not benefit so much from this restriction as they would find it confusing. 

To summarise: the best argument for disallowing accessing of shadowed variables was to align with es specifications. But now that we have rid source of this requirement (or if not yet, I think it should happen), then there does not seem to be a good reason for this restriction.

On a similar note, I don't think the lengthly warning message about *"local functions within an outer function must precede all other statements in body of the outer function"*. This should be the default understanding to the student who only knows of a sequential, synchronous, 'line-after-line' interpretation of their program. All statements a student would have written so far would have been executed strictly one after another, so it should make sense by default that a function definition must come before it's call.

martin-henz:
> If I have followed the discussion correctly so far, it seems source no longer needs to be a subset of javascript, following some change in scoping (correct me if I'm wrong). Honestly, I think this is a good change. There are many messy aspects of JS that would not benefit someone learning programming for the first time.
> 
> _If_ this is the case, I don't see why we must prevent disallowed accessing shadowed variables. It is undoubtedly bad practice, and certainly illegal in some programming languages, but also perfectly logical to the beginner who simply expects all statements to run sequentially from the first line of their program to the last. A student would not benefit so much from this restriction as they would find it confusing.
> 
> To summarise: the best argument for disallowing accessing of shadowed variables was to align with es specifications. But now that we have rid source of this requirement (or if not yet, I think it should happen), then there does not seem to be a good reason for this restriction.
> 
> On a similar note, I don't think the lengthly warning message about _"local functions within an outer function must precede all other statements in body of the outer function"_. This should be the default understanding to the student who only knows of a sequential, synchronous, 'line-after-line' interpretation of their program. All statements a student would have written so far would have been executed strictly one after another, so it should make sense by default that a function definition must come before it's call.

Hi Ning Yuan: We are not going to allow accessing shadowed variables, following ES specs. Not so sure what you have in mind. Can you give an example of a program that should be dis-allowed following your suggestion but that would work in our proposal?

Regarding: local functions within an outer function must precede all other statements in body of the outer function

The issue here is to remain aligned with ES. If we impose this rule, then all legal programs work as expected and consistent with ES.

You mentioned that we move away from ES, anyway. We try not to. Do you have an example where we do things differently from ES? I feel it's important to remain a strict sub-language of JavaScript, in order to motivate groups of students beyond a CS programme  (where they really don't have a choice).

-martin

ningyuansg:
@martin-henz My intention is to allow accessing of shadowed variables, and to not follow ES specs. I think that it is good enough that source is *similar* to javascript, rather strictly a subset of javascript. The following program is allowed with my suggestion.

```javascript
let x = 4;
if (true) {
    display(x); // displays 4, no reference error thrown
    let x = 6;
}
display(x); // displays 4 still
```

Once again, bad practice, but completely understandable and *expected* to a student who knows only the environment module, and a synchronous line-to-line 'framework' of program execution.

> You mentioned that we move away from ES, anyway. We try not to. Do you have an example where we do things differently from ES? I feel it's important to remain a strict sub-language of JavaScript, in order to motivate groups of students beyond a CS programme (where they really don't have a choice).

I was referring to the discussion [here](https://github.com/source-academy/js-slang/issues/65#issuecomment-425624761). I see that the deviation from ES is currently a trivial one, but I suggest that the strict adherence to being ES compliant is detrimental to source. My understanding that the primary benefit of following ES specs is to teach programming methodology in a modern syntax so students can adjust to other languages (including JS) easier in other modules, or at their internships. If that is the case, it should be sufficient that source is *similar* to ES, but not necessarily following every aspect of the ES spec. The marginal benefit of strictly following ES here is not worth the confusion it may bring to the students.

martin-henz:
> @martin-henz My intention is to allow accessing of shadowed variables, and to not follow ES specs. I think that it is good enough that source is _similar_ to javascript, rather strictly a subset of javascript. The following program is allowed with my suggestion.
> 
> ```js
> let x = 4;
> if (true) {
>     display(x); // displays 4, no reference error thrown
>     let x = 6;
> }
> display(x); // displays 4 still
> ```
> 
> Once again, bad practice, but completely understandable and _expected_ to a student who knows only the environment model, and a synchronous line-to-line 'framework' of program execution.
> 
> > You mentioned that we move away from ES, anyway. We try not to. Do you have an example where we do things differently from ES? I feel it's important to remain a strict sub-language of JavaScript, in order to motivate groups of students beyond a CS programme (where they really don't have a choice).
> 
> I was referring to the discussion [here](https://github.com/source-academy/js-slang/issues/65#issuecomment-425624761). I see that the deviation from ES is currently a trivial one, but I suggest that the strict adherence to being ES compliant is detrimental to source. My understanding that the primary benefit of following ES specs is to teach programming methodology in a modern syntax so students can adjust to other languages (including JS) easier in other modules, or at their internships. If that is the case, it should be sufficient that source is _similar_ to ES, but not necessarily following every aspect of the ES spec. The marginal benefit of strictly following ES here is not worth the confusion it may bring to the students.

I strongly disagree. The hallmark of "lexical scoping" is that it is "static": You can look at a program and figure out what variable declaration a variable occurrence refers to. The order of execution, for example sequential composition or concurrency, should not matter. Your proposal would violate this principle. What would you do with the following program?

{ let p = y => x + y; let x = 2; p(3); }

It's perfectly fine, as long as you realise that the scope of a declaration is the surrounding block. 

So yes, I would like to defend JavaScript in this case. I think we can have the best of all worlds: 
* be a subset of the world's most popular language, following its semantics to the T, and
* have a clean language that is easy to teach, and
* have a powerful language that is fun to work with.

martin-henz:
@yyc can you take a look and check that this is consistent with our revised specs and your new loop implementation, both of which are merged now.

openorclose:
currently functions are hoisted to the top including the definition, which is not right. Will refactor to only hoist the declaration part to the top and not the definition.

The variable hoisting part should be fine.

martin-henz:
OK, so the merge of this branch will wait till you are done with the refactoring. In the meantime, we will deploy Source §3 in the Playground so that we can test the textbook examples of Chapter 3.

PR #57, opened 2018-09-16
Fix parse_int error using 'parseInt' name

ningyuansg:


yunpengn:
photo6120486155699988577

FYI: See the screenshot attached.

PR #62, opened 2018-09-22
Add block scoping for loops

yyc:
As in title. 
Also added separate scoping for loop variables and loop blocks in for loops (see test cases, `block_scoping.ts`


openorclose:
```let``` declarations are not correctly freshly bound at each iteration of the loop. See following example.

```javascript
function test(){
      let z = [];
      for (let x = 0; x < 10; x = x + 1) {
        z[x] = () => x;
      }
      return z[5]();
}
test();
```

```test```` should return 5 but instead returns 10.

yyc:
@openorclose Interesting, nice catch! Is this behaviour documented anywhere in the JS spec?

openorclose:
[Link to the spec](https://www.ecma-international.org/ecma-262/6.0/#sec-for-statement-runtime-semantics-labelledevaluation).
[Mozilla blog on it](https://hacks.mozilla.org/2015/07/es6-in-depth-let-and-const/).

Motive for this new binding in es2015 most likely comes from wanting to bind event handlers dynamically more easily, rather than the boilerplate creation of an IIFE to trap the needed scope.

yyc:
Fixed. Thanks @openorclose!

martin-henz:
function test(){
      let z = [];
      for (let x = 0; x < 10; x = x + 1) {
        z[x] = () => x;
      }
      return z[5]();
}
test();

Hmm, yes, this must return 5. How to explain any other result?

Looks like a legit program to me. No issues with scoping or anything else, is there?

martin-henz:
Do we know @openorclose ? Welcome in this group! Looks like you have a deep interest in the Source Academy.

PR #68, opened 2018-09-23
Refactor loop scoping (#65)

### LINKED ISSUE #65: Critical for loop issues
a) Now the following valid es2015 code results in an error:

```javascript
function test(){
      for (let x = 0; x < 10; x = x + 1) {
        let x = 1; //Error on this line, should be valid
      }
}
test();
```

```let``` statements should still be block scoped in the body of a for loop.

b) Another problem:

```javascript
function test(){
      let timesRun = 0;
      for (let x = 0; x < 10; x = x + 1) {
        timesRun = timesRun + 1;
        x = x + 1;
      }
      return timesRun;
}
test();
```
The loop should only run five times as x is updated in the loop, but due to shallow copying of the environment variables into the new frame, x is not updated as it should be.

yyc:
Resolves, if not completely fixes #65 (See comment by @martin-henz [here](https://github.com/source-academy/js-slang/issues/65#issuecomment-423804736))

Also thanks to @openorclose's suggestions, refactor the block scoping to be more DRY by putting it just in BlockStatement.

### DIFF
--- a/src/__tests__/__snapshots__/block_scoping.ts.snap
+++ b/src/__tests__/__snapshots__/block_scoping.ts.snap
@@ -7,13 +7,6 @@ Object {
 }
 `;
 
-exports[`for loop \`let\` variables are copied into the block scope 1`] = `
-Object {
-  "status": "finished",
-  "value": 1,
-}
-`;
-
 exports[`for loops use block scoping instead of function scoping 1`] = `
 Object {
   "status": "finished",
@@ -21,7 +14,7 @@ Object {
 }
 `;
 
-exports[`interior of for loops don't affect loop variables 1`] = `
+exports[`interior declarations for loops don't affect loop variables 1`] = `
 Object {
   "status": "finished",
   "value": 4,
--- a/src/__tests__/__snapshots__/block_scoping.ts.snap
+++ b/src/__tests__/__snapshots__/block_scoping.ts.snap
@@ -7,6 +7,13 @@ Object {
 }
 `;
 
+exports[`for loop \`let\` variables are copied into the block scope 1`] = `
+Object {
+  "status": "finished",
+  "value": 2,
+}
+`;
+
 exports[`for loops use block scoping instead of function scoping 1`] = `
 Object {
   "status": "finished",
--- a/src/__tests__/__snapshots__/block_scoping.ts.snap
+++ b/src/__tests__/__snapshots__/block_scoping.ts.snap
@@ -1,5 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Cannot overwrite loop variables within a block 1`] = `"Line 5: Cannot assign new value to constant x"`;
+
 exports[`const uses block scoping instead of function scoping 1`] = `
 Object {
   "status": "finished",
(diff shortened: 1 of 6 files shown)


yyc:
Also found and fixed a critical bug where `const` variables could also be assigned to in Source §3.

martin-henz:
Asking openorclose to comment on this pull request.

openorclose:
@martin-henz 

Referencing (4) in #65, where return is not allowed in body of loops. `test` still returns true when called. 

```javascript
  function test(){
      for (let x = 0; x < 2; x = x + 1) {
        return true;
      }
      return false;
  }
  test();
```

remo5000:
The `const` fix looks good. @openorclose any comments on the loop scoping?

openorclose:
This code works when it shouldn't as well. All 3 expressions in the for statement are required.

```javascript
  function test(){
      for (let x = 0; ; x = x + 1) {
        return true;
      }
      return false;
  }
  test();
```

`test()` still returns true.

openorclose:
Also, currently function declarations allow the function name to be assigned to:
```javascript
function a() {return 1;}
a =100;
```
Is this intended behaviour, because it is not stated in source 3 docs?

martin-henz:
> This code works when it shouldn't as well. All 3 expressions in the for statement are required.
> 
> ```js
>   function test(){
>       for (let x = 0; ; x = x + 1) {
>         return true;
>       }
>       return false;
>   }
>   test();
> ```
> 
> `test()` still returns true.

Yes, there should be error messages:
(1) for test not optional
(2) return not allowed in for loops. This error message is not a showstopper. If it's too complicated to implement, we can just say "undocumented feature; don't do it".

martin-henz:
> Also, currently function declarations allow the function name to be assigned to:
> 
> ```js
> function a() {return 1;}
> a =100;
> ```
> 
> Is this intended behaviour, because it is not stated in source 3 docs?

Yes, this is a gap in the documentation. My gut feeling is that we should keep it this way: A weirdness of Source. Explicitly preventing it and documenting this would create unnecessary confusion.

In any case: let's revisit this in 2019.

openorclose:
Other than the above, I believe there are no other issues regarding the for loops.

yyc:
@martin-henz we could disable overriding function declarations, so 
```function a() { return 1; }```
is semantically identical to 
```const a = () => 1;```

Or if we allow it, then  it will be translated to 
```let a = () =>;```

It's simple to change either way (and also simple for the students to get around if necessary), we just need to agree on one of the two.

Meanwhile, I'll make the changes that check that the three statements in a for loop are not empty.

martin-henz:
> @martin-henz we could disable overriding function declarations, so
> `function a() { return 1; }`
> is semantically identical to
> `const a = () => 1;`
> 
> Or if we allow it, then it will be translated to
> `let a = () =>;`
> 
> It's simple to change either way (and also simple for the students to get around if necessary), we just need to agree on one of the two.

Yes, make it identical to
`const a = () => 1;`.

This is consistent with what we said in the lectures. No one will miss this JavaScript "feature" to re-assign function variables. I won't bother documenting this, to avoid unnecessary confusion.

martin-henz:
@yyc : You have seen the discussion with @openorclose about the "temporal dead zone" elsewhere. This will take care of what we discussed earlier.

Issue #56, opened 2018-09-11
lists to string


martin-henz:
How to represent lists as strings?

The specs say for +:
(any, string) -> string
(string, any) -> string

That means we need some representation for lists (pairs). Currently we have:

display(list(1,2,3) + "");
"1,2,3,"

That's a bit weird.

Can we have:

"[1,[2,[3,[]]]]"

?

PR #70, opened 2018-09-29
Concat string with (functions and lists) enhancement

### LINKED ISSUE #69: "implementation hidden" feature not working when adding function to a string
You can print the implementation of any function just by adding it to an empty string `""`.
For instance:
```JavaScript
//input
make_binary_tree_node+"";
//output
"function make_binary_tree_node(left, value, right) {
    if (!is_binary_tree(left)) {
        throw new Error("Left subtree is not a valid binary tree");
    } else if (!is_binary_tree(right)) {
        throw new Error("Right subtree is not a valid binary tree");
    }
    return list(left, value, right);
}"
```
This can be avoided by checking the type of operand before performing addition operation to strings. (Although in the Source language description it says any type can be added to a string, but a function added to a string should be censored.)

openorclose:
Used Source ```toString``` method instead of native ```toString``` during ```BinaryExpression```.

Should resolve #69 and #56

### DIFF
--- a/doc/source_1.tex
+++ b/doc/source_1.tex
@@ -14,8 +14,7 @@
                                            \textbf{\texttt{=}}\  \textit{expression} \ \textbf{\texttt{;}}
                                                            && \textrm{constant declaration} \\
 &&                       && |   &\quad && \textbf{\texttt{function}}\  \textit{name} \ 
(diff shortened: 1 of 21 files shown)


martin-henz:
Not sure what's going on here. Travis build failed and coverage decreased. @openorclose : can you take a look?

openorclose:
Fixed it. Thanks

PR #75, opened 2018-10-01
fix standalone block statement scoping

yyc:
Fixed block statement scoping (so standalone block statements are now correctly scoped).

A downside now is that variables declared in loop blocks can now shadow loop variables, same as in ES6, but we should discourage this behaviour.
```
  function test(){
      for (let x = 0; x < 2; x = x + 1) {
        let x = 1;
        print(x);
      }
      return false;
  }
```


martin-henz:
@remo5000 took a look and @openorclose has commented extensively. I think this is ready to go.

PR #76, opened 2018-10-02
Disable UpdateExpression from Chapter 3

remo5000:
This is not a specified operator in the documentation, and should not be allowed. In addition, there is no support in the interpreter for handling this expression, so enabling this type of expression would require 
1. A change in the language specifications
2. Implementation of a handler for this expression


ningyuansg:
We should probably start writing tests for these disallowed syntax types

martin-henz:
Yes, agree.

PR #77, opened 2018-10-02
Updated hoisting

openorclose:
Now the identifiers after `let` and `const` and function names are hoisted to the top of the scope. The assignments are not hoisted.

### DIFF
--- a/src/__tests__/block_scoping.ts
+++ b/src/__tests__/block_scoping.ts
@@ -194,3 +194,49 @@ test("Cannot leave blank expressions in for loops", () => {
     });
   });
 });
+test(
+  'No hoisting of functions. Only the name is hoisted like let and const',
+  () => {
+    const code = `
+      const v = f();
+      function f() {
+        return 1;
+      }
+      v;
+    `
+    const context = mockContext()
+    const promise = runInContext(code, context, { scheduler: 'preemptive' })
+    return promise.then(obj => {
+      const errors = parseError(context.errors)
+      expect(obj).toMatchSnapshot()
+      expect(obj.status).toBe('error')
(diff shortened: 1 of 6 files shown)


src/__tests__/block_scoping.ts
```diff
@@ -194,3 +194,49 @@ test("Cannot leave blank expressions in for loops", () => {
...
+      v;
+    `
+    const context = mockContext()
+    const promise = runInContext(code, context, { scheduler: 'preemptive' })
+    return promise.then(obj => {
+      const errors = parseError(context.errors)
+      expect(obj).toMatchSnapshot()
+      expect(obj.status).toBe('error')
+      expect(errors).toEqual(
+        expect.stringMatching(/^Line 2: Name f not declared/)
```
thomastanck: Shouldn't the error be something more like "Cannot access f before it's assigned" or something?
maim
This is what firefox says.

openorclose: image
I used that error because it's what chrome gives. I guess firefox's message is more meaningful, as it specifically states that f is not defined _yet_, whereas chrome just says that it's undefined.

"Cannot access x before it's assigned" will be ambiguous to students when x is already assigned in an outer scope but not yet in the current scope like:
```javascript
  const x = 1;
  function inner() {
    const double_x = x * 2;
    const x = "x";
  }
  inner();
```

What about "`<name>` is not yet assigned in the current scope"?

thomastanck: The last one sounds good.

To nitpick, defining and declaring is the difference I was trying to note. (chrome says "not defined", not "not declared")

PR #82, opened 2018-10-04
Restrict parse_int radix to [2, 36]

### LINKED ISSUE #52: parse_int function does not match ECMAscript specification
```
parse_int("1",36); // returns 1 intended behavior
parse_int("1",46);
parse_int("uu1",36);
```

```"Error: parseInt expects two arguments a string s, and a positive integer i"``` is thrown while both arguments are as stated in the error. In addition, according to ECMAScript specification, a ```NaN``` is returned for these two cases.

ningyuansg:
Behaviour now follows ES6 except:

- If either args str or radix is not provided, source error is thrown
- If first arg is not str, source error is thrown
- If second arg is not integer in [2, 36], source error is thrown

Resolves #52.


martin-henz:
Yes, that's fine.

PR #86, opened 2018-10-10
Disable Compound Assignment Expressions from Chapter 3

Wingedevil:
The current handling of assignment expressions treats all assignment expression as a ```=```. Additionally, there is no error handling for compound assignment expressions.


openorclose:
Obsoleted by #116 .

Issue #90, opened 2018-10-12
Display of arrays broken


martin-henz:
function zero_matrix(rows, cols) {
    const M = [];
    for (let r = 0; r < rows; r = r + 1) {
        M[r] = [];
        for (let c = 0; c < cols; c = c + 1) {
            M[r][c] = 0; 
        }
    }
    return M;
}
zero_matrix(3,4);

==>
[0,0,0,0,0,0,0,0,0,0,0,0]
instead of
[[0,0,0,0],[0,0,0,0],[0,0,0,0]]

PR #91, opened 2018-10-12
Fix multidimensional array as string

openorclose:
#90


blackening:
This is a little weird. I implemented my own version since i needed it, then i checked and found that there's an existing PR.

Some comments:

Current implementation will attempt to stringify very long arrays. 
May want to consider truncating.
`[${value.map(v => toString(v, length + 1)).join(', ')}]`

This is what i implemented:

```
const arrayContentsToString = (array: Value[], length: number) => {
  let result = "";
  for(let i = 0; i < array.length; i++){
    if(length + i > MAX_LIST_DISPLAY_LENGTH){
      return i ? result + ", ...<truncated>" : "...<truncated>"
    }
    let contentString = toString(array[i], length + i + 1);
    result += i ? `, ${contentString}` : contentString
  }
  return result;
}
```

I know it doesn't truncate properly, i.e. if someone creates 98 arrays of 98 items and dumps it into an array, it will display all of them. This probably requires a proper rewrite to do truncation (recursive calls must return resultant length).

blackening:
Also can i check something? Does the second one work for you?
```
test("Functional bi-array", () => {
  const code = `
    let t = () => () => 1 + 1;
    const x = [t,t];
    x;
  `;
  const context = mockContext(3);
  const promise = runInContext(code, context, { scheduler: "preemptive" });
  return promise.then(obj => {
    expect(obj).toMatchSnapshot();
    expect(obj.status).toBe("finished");
    expect(toString((obj as Finished).value)).toBe("[() => () => 1 + 1, () => () => 1 + 1]");
  });
});
```

```
test("Functional pair", () => {
  const code = `
    let t = () => () => 1 + 1;
    const x = pair(t,t);
    x;
  `;
  const context = mockContext(3);
  const promise = runInContext(code, context, { scheduler: "preemptive" });
  return promise.then(obj => {
    expect(obj).toMatchSnapshot();
    expect(obj.status).toBe("finished");
    expect(toString((obj as Finished).value)).toBe("[() => () => 1 + 1, () => () => 1 + 1]");
  });
});
```

openorclose:
What's the rationale for truncating the string based on array length and
not string length? It seems rather arbitrary.

No, the second one doesn't work currently for me either. It's an issue
raised and fixed (partially) in #92. Prototypes don't work correctly for my
fix in #92 when the Closure/ArrowClosure is passed into a native Javascript
function, but OOP not in Source 2019.

Issue #88, opened 2018-10-12
Unintended behavior involving self-defined functions and pairs


Wingedevil:
```
function s() {
    return 1;
}
head(pair(s, s)) === s;
```
We expect this equality to be true, but unfortunately, it resolves to false;

openorclose:
This is due to a more general bug, where Source-defined functions are not passed into native Javascript functions properly, and while calling the functions give the same value, they do not refer to the same object.

thomastanck:
Resolved by #92 :tada:

PR #92, opened 2018-10-13
Fix functions equality

openorclose:
#88


thomastanck:
Might want to update this PR by resolving the merge conflicts.

src/types.ts
```diff
@@ -153,50 +153,65 @@ export interface Frame {
...
   /** Keep track how many lambdas are created */
   private static lambdaCtr = 0
 
   /** Unique ID defined for anonymous closure */
-  public name: string
+  public functionName: string
 
   /** Fake closure function */
-  // tslint:disable-next-line:ban-types
+    // tslint:disable-next-line:ban-types
```
thomastanck: Unnecessary spaces?

src/__tests__/index.ts
```diff
@@ -189,6 +203,21 @@ test(
   30000
 )
 
+test("Functions passed into non-source functions remain equal", () => {
+  const code = `
+  function t(x, y, z) {
+    return x + y + z;
+  }
+  head(pair(t, t)) === t && t(1, 2, 3) === 6;
```
thomastanck: I think to be more robust against future library changes (what if head/pair are later implemented in pure Source), you could use `defineSymbol` from `createContext.ts` to create your own built-in that's guaranteed to be non-source.

src/types.ts
```diff
@@ -153,50 +153,65 @@ export interface Frame {
...
+    const funJS = closureToJS(this, context, this.functionName)
+    this.fun = funJS
   }
 }
 
 /**
  * Modified from class Closure, for construction of arrow functions.
  */
-export class ArrowClosure {
+export class ArrowClosure extends Callable{
```
thomastanck: Missing space.

openorclose: Done

src/__tests__/index.ts
openorclose: Yes, it was probably wrong to rely on external "builtins".

Issue #89, opened 2018-10-12
empty statements should be documented in Source §x


martin-henz:
Currently
if (..) { ... } else {}
is an undocumented feature.

martin-henz:
This is fixed in the specs.

PR #95, opened 2018-10-14
Source Specification 2019

martin-henz:


doc/source_list_library.tex
```diff
@@ -0,0 +1,203 @@
...
+        || item1 === item2;
+}
+
+// returns the length of a given argument list
+// assumes that the argument is a list
+
+function length(xs) {
+    return is_empty_list(xs) 
+	? 0
+        : length(tail(xs));
```
openorclose: Should be `1 + length(...`

doc/source_typing_3.tex
```diff
@@ -0,0 +1,46 @@
...
+\texttt{\textbf{>}} & number   & number     & bool\\
+\texttt{\textbf{>}} & string   & string     & bool\\
+\texttt{\textbf{<}} & number   & number     & bool\\
+\texttt{\textbf{<}} & string   & string     & bool\\
+\texttt{\textbf{>=}} & number   & number     & bool\\
+\texttt{\textbf{>=}} & string   & string     & bool\\
+\texttt{\textbf{<=}}    & number   & number     & bool\\
+\texttt{\textbf{<=}} & string   & string     & bool\\
+\texttt{\textbf{\&\&}} & bool & bool & bool\\
+\texttt{\textbf{||}}   & bool & bool & bool\\
```
openorclose: For `&&` and `||` are the specs going to change to bool op any = any?

martin-henz: Yes. We don't make an explicit claim about this, but currently the wording suggests that (or at least leaves open the question whether) there are runtime tests for this. As it stands, the translation to conditional expressions would prevent such a runtime test.

thomastanck:
This PR resolves #89. Would be great to get this doc PR merged (after resolving the merge conflicts with #130)

doc/source_objects.tex
```diff
@@ -12,14 +12,14 @@ \subsection*{Object properties}
 If the string (without quotation marks) looks like a \textit{name}, the
 quotation marks around property keys can be omitted, for example:
 \begin{lstlisting}
-var my_motorcycle = { cc: 399,
+var the_motorcycle = { cc: 399,
```
openorclose: `var` -> `const`

doc/source_list_library.tex
```diff
@@ -1,40 +1,42 @@
...
 
-function equal(item1, item2){
-    return (is_pair(item1) && is_pair(item2))
-        || (is_empty_list(item1) && is_empty_list(item2))
-        || item1 === item2;
+function equal(x, y){
+    return (is_pair(x) && is_pair(y)) 
+        ? (equal(head(x), head(y)) &&
+           equal(tail(x), tail(y)))
+        : (is_null(x) && is_null(y))
```
openorclose: `is_null` is redundant, since `null === null`. Not sure if it was explicitly included for explanation's sake though.

Also, in Source 2018, `[] !== []` so there was `is_empty_list`. In Source 2019, is it better to use `===` to check for `null` instead?

doc/source_stream_library.tex
```diff
@@ -0,0 +1,206 @@
...
+            ? null
+            : pair(fun(i), 
+                   () => build(i + 1)); 
+    }
+    return build(0);
+}
+
+// stream_for_each applies first arg fun to the elements of the list
+// passed as second argument. fun is applied element-by-element:
+// for_each(fun,list_to_stream([1,[2,null]])) results in the calls fun(1)
```
openorclose: Refer to analogous list version.

doc/source_stream_library.tex
```diff
@@ -0,0 +1,206 @@
...
+function stream_length(xs) {
+    return is_null(xs)
+        ? 0
+        : 1 + stream_length(stream_tail(xs));
+}
+
+// stream_map applies first arg f to the elements of the second
+// argument, assumed to be a stream.
+// f is applied element-by-element:
+// stream_map(f,list_to_stream([1,[2,null]])) results in
```
openorclose: Refer to analogous list version.

doc/source_list_library.tex
```diff
@@ -1,40 +1,42 @@
...
 	? 0
-        : length(tail(xs));
+        : 1 + length(tail(xs));
 }
 
 // map applies first arg f, assumed to be a unary function,
 // to the elements of the second argument, assumed to be a list.
 // f is applied element-by-element: 
-// map(f, [1, [2, []]]) results in [f(1), [f(2), []]]
+// map(f, [1, [2, null]]) results in [f(1), [f(2), null]]
```
openorclose: Should replace `[1, [2, null]]` with `list(1, 2)` as per `append` definition for consistency? I recommend the `append` way as the current example can't be run in the playground in Source 1 and 2 as they don't support array syntax.  And maybe replace the result with `list(f(1), f(2)` also.

doc/source_list_library.tex
```diff
@@ -45,26 +47,26 @@ \section*{Appendix: List library}
...
     }
-    return build(n - 1, fun, []);
+    return build(n - 1, fun, null);
 }
 
 // for_each applies first arg fun, assumed to be a unary function,
 // to the elements of the second argument, assumed to be a list.
 // fun is applied element-by-element:
-// for_each(fun, [1, [2, []]]) results in the calls fun(1) and fun(2).
+// for_each(fun, [1, [2, null]]) results in the calls fun(1) and fun(2).
```
openorclose: Same as `map` above, might want to replace literal list syntax.

doc/source_list_library.tex
```diff
@@ -79,8 +81,8 @@ \section*{Appendix: List library}
 // When it encounters a non-list, it applies toString to it.
```
openorclose: Should be `to_string` as defined earlier.

doc/source_1.tex
```diff
@@ -25,7 +25,9 @@
 &&                       && |   &\quad &&  \textit{block} 
                                                            && \textrm{block statement}\\
 &&                       && |   &\quad &&  \textit{expression} \ \textbf{\texttt{;}}
-                                                           && \textrm{expression statement} \\[1mm] 
+                                                           && \textrm{expression statement} \\
+&&                       && |   &\quad &&  \epsilon
```
openorclose: Empty statements in Javascript are represented by a single semicolon, and is currently not valid in Source. If this is to allow the 

```javascript
if (true) {
  do_something();
} else { /*nothing here*/}
```

idiom that is being used, I think block should be changed instead to allow the empty block:

```block ::= {\epsilon | statement}```

and drop the empty statement definition.

In either case, the `while` definition in https://github.com/source-academy/js-slang/blob/355564134e836d56bb6b426252e629d1fe00643d/doc/source_loops.tex#L28 should drop the redundant `undefined;` statement.

openorclose: Currently a single semicolon as a statement is not allowed (results in " Empty statements are not allowed"), but a blank statement (nothing, without semicolon) is allowed, also referred to as an empty statement in the Source specifications.

This results in "empty statement" having two definitions (one referring to ` ` while one referring to `;`), which could be confusing.

Since empty statement in JavaScript already means `;`, maybe the name of ` ` as a statement could be changed?

Or we could just change the error message currently regarding ; statements to say "; is not a valid statement".

Either way, I believe that this dual usage of the term "empty statement" should be addressed.

martin-henz: I agree with your reasoning. I think we should move to JavaScript here, and allow the empty JavaScript statement
;
and disallow nothing to represent a statement.

doc/source_objects.tex
martin-henz: OK.

doc/source_stream_library.tex
martin-henz: I don't see the mistake here.

doc/source_list_library.tex
martin-henz: noted!

doc/source_list_library.tex
martin-henz: noted!

doc/source_list_library.tex
martin-henz: noted!

doc/source_1.tex
martin-henz: Yes, I'm proposing to have
;
as the empty statement, following JavaScript.

martin-henz:
I've pushed the requested changes.

Note that I'm proposing 
;
as empty statement.

Asking everyone to comment, in the group.

doc/source_stream_library.tex
openorclose: I'm referring to line 38, `stream_map(f,list_to_stream([1,[2,null]])) results in`, in particular `[1,[2,null]]`.

thomastanck:
After Daryl's thorough review it looks good to me.

martin-henz:
Quite close to deciding to drop objects altogether, also for Sem 2. The MCE wouldn't change that much. After all, the authors made a point at wrapping things quite well into an abstract syntax, with accessor functions. I'm keen to get rid of objects, to get closer to SICP. We don't need to cover objects, and the objects that we do cover don't add much, conceptually.

So how about just ripping out objects, already now?

thomastanck:
Agreed, as we discussed during the meeting we won't need objects during Sem 2. We can always change it back later if we change our minds, too.

thomastanck:
https://github.com/source-academy/js-slang/pull/95/files#diff-6a9133abb057030b6d9882f0cb7492d1R120

This needs to be removed from source_4.tex

martin-henz:
Just removed empty statements. All good?

openorclose:
Yep, looks good!

PR #107, opened 2018-11-01
C4 rush

martin-henz:


yyc:
@martin-henz it looks good to me but I don't have a lot of context on this so you might want @ningyuansg or kenneth to double-check.

PR #111, opened 2018-11-04
Fix multidimensional arrays as string.

openorclose:
Temporary fix, as no truncation is done for arrays with too many elements.


openorclose:
done

PR #114, opened 2018-11-07
(WIP) Fix Jison spec for metacircular-interpreter parser

thomastanck:
The spec is changed!

Identifiers are now *always* parsed as tagged "name" objects. In particular this means that `const`, `let`, assignment statements, etc. etc. all break now.

Summary of changes:

- Fix the `parser.jison.tpl` file. It had incorrect location objects as well as a missing closing brace.
- Generate `parser.js` files using `node generate.js --environment`. They are no longer divergent.
- Change behaviour of `identifier`: identifiers are transformed into tagged name objects immediately, instead of being kept around as a string until something turns it into a tagged name object.

Todo:

- Fix the MCE(s) to work with these changes.


thomastanck:
I'll just note that for smoothness sake, it'd be great not to merge this in until the module has concluded. Thanks.

ningyuansg:
Thanks! Though I was thinking that we should abandon this jison based parser in favour of something based off acorn, which the main parser uses anyways.

thomastanck:
Obsoleted by #117

PR #115, opened 2018-11-07
Write parser for MCI using acornjs

thomastanck:


thomastanck:
I need help finishing up this PR. I have no idea how to sanely throw errors from inside the library! I've scattered `throw new SyntaxError("blah")` around the place... Please advise on what I should do instead.

src/stdlib/__tests__/parser.ts
```diff
@@ -0,0 +1,368 @@
...
+  const context = mockContext(4)
+  const promise = runInContext(code, context, { scheduler: 'preemptive' })
+  return promise.then(obj => {
+    expect(JSON.stringify(context.errors, undefined, 2)).toBe('[]')
+    expect(obj).toMatchSnapshot()
+    expect(obj.status).toBe('finished')
+  })
+})
+
+test('Does not parse incomplete statements', () => {
```
openorclose: Should this be in the metacircular interpreter tests? It should already be checked in the `disallowed_syntax` tests. Same for the other syntax checks.

thomastanck: True, the negative tests are a bit redundant. I'll remove the negative tests (but keep the positive ones).

openorclose:
I think your commits diallowing assignment expressions and update expressions should be in another pr.

openorclose:
For the error throwing, I guess throwing SyntaxError is fine as they will still be caught? You can try providing the location of the error as well.

thomastanck:
Split into #116 and #117

Issue #80, opened 2018-10-04
Disallowed assignment operators having unintended behavior


Wingedevil:
```
var x = 1;
x += 2; // is allowed and is treated as x = 2;
```
All assignment operators are treated as the default ```=``` operator.

thomastanck:
Closing in favour of #129, a combination of a bunch of issues, to clean up the list of issues a little.

PR #116, opened 2018-11-07
Various parsing fixes

thomastanck:
- Disallow update assignment `a += 5;`
- Disallow assignment expressions `x = y = 3;`
- Disallow function expressions `(function(x){return x+1;})(1);`
- Add a lot of test cases


martin-henz:
We can't disallow
(function(x){return x+1;})(1);
before the semester ends. Mission grading depends on it.

thomastanck:
I've removed the last change, is this ready for merge?

thomastanck:
Note to close issue later as this would resolve #80.

thomastanck:
Squashed all the commits together so it's a bit more readable, along with a better commit message.

PR #117, opened 2018-11-07
2019: Rewrite MCI using acornjs (depends on #116)

### LINKED ISSUE #24: Metacircular interpreter's parser uses AY17/18 Source Lanugage specs
### Problem
- The parser that is generated using `jison` uses grammar that works for the previous implementation for Source
- The rules would have to be changed
  - [ ] `var` must be changed to `const` and `let`
  - [ ] Arrow functions must be added
  - [ ] Anonymous functions must be removed (?)

thomastanck:
New version of #115 

Resolves #24.


thomastanck:
Agreed, this change is definitely for next semester.

I think thats why @openorclose suggested to split off #116 as that one is not a breaking change.

martin-henz:
116 would disallow:

(function(x){return x+1;})(1);

We can't do that, because mission grading depends on it. Can't change this
during this semester. Need to wait for next.

-martin
http://www.comp.nus.edu.sg/~henz

thomastanck:
Should I disable that change for now, to be re-enabled next semester?

martin-henz:
Yes no changes please at this stage, unless it's serious bugs. Tx.

-martin

martin-henz:
This will break when objects are removed. Let's have a CP3108 project on an object-free MCI, and shelve this PR.

thomastanck:
We should probably still merge this in first so we can remove the old parser. The only thing left to change are the transformation rules for turning the estree spec compliant parse tree into a CS1101S compliant one.

martin-henz:
#116 is merged now. Feel free to merge this one. We can take a closer look later. I'll try to get a CP3108 team on this.

scripts/build.sh
```diff
@@ -3,4 +3,4 @@
 # This is required because it is not possible to both inlcude .js
```
openorclose: Time to remove these obsolete comments?

src/stdlib/parser.ts
```diff
@@ -0,0 +1,309 @@
...
+  })}],
+  ["ConditionalExpression", (node: es.Node) => {
+    node = <es.ConditionalExpression> node
+    return ({
+      tag: "conditional_expression",
+      predicate: transform(node.test),
+      consequent: transform(node.consequent),
+      alternative: transform(node.alternate)
+  })}],
+  ["FunctionExpression", (node: es.Node) => {
```
openorclose: Time to remove FunctionExpression

src/stdlib/parser.ts
```diff
@@ -0,0 +1,309 @@
...
+    if (node.key.type === 'Literal') {
+      return [node.key.value, transform(node.value)];
+    } else if (node.key.type === 'Identifier') {
+      return [node.key.name, transform(node.value)];
+    } else {
+      unreachable()
+      throw new ParseError('Invalid property key type')
+    }
+  }],
+  ["EmptyStatement", (node: es.Node) => {
```
openorclose: And empty statements

src/stdlib/parser.ts
```diff
@@ -0,0 +1,309 @@
...
+    node = <es.BreakStatement> node
+    return ({
+      tag: "break_statement"
+  })}],
+  ["ContinueStatement", (node: es.Node) => {
+    node = <es.ContinueStatement> node
+    return ({
+      tag: "continue_statement"
+  })}],
+  ["ThisExpression", (node: es.Node) => {
```
openorclose: Remove this as well.

scripts/build.sh
thomastanck: Removed the build script entirely as it's now useless

src/stdlib/parser.ts
thomastanck: Removed all the transformers you listed as well.

PR #118, opened 2018-11-08
Refactor for statement check into rules

openorclose:
The for statement requiring all 3 parts should be refactored out of the interpreter and into rules, avoiding duplication of such a check for the metacircular interpreter once it uses the same parser as the Source interpreter.

### DIFF
--- a/src/rules/forStatementMustHaveAllParts.ts
+++ b/src/rules/forStatementMustHaveAllParts.ts
@@ -1,0 +1,44 @@
+import { stripIndent } from 'common-tags'
+import * as es from 'estree'
+
+import { ErrorSeverity, ErrorType, Rule, SourceError } from '../types'
+
+export class ForStatmentMustHaveAllParts implements SourceError {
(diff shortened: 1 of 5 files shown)


src/rules/forStatementMustHaveAllParts.ts
```diff
@@ -0,0 +1,44 @@
...
+    `
+  }
+}
+
+const forStatementMustHaveAllParts: Rule<es.ForStatement> = {
+  name: 'for-statement-must-have-all-parts',
+
+  checkers: {
+    ForStatement(node: es.ForStatement) {
+      const missingParts = ["init", "test", "update"].filter(
```
thomastanck: Maybe change to ```if (["init", "test", "update"].some(part => node[part] === null) {...```?

thomastanck: nvm, you save the missingParts for use in the error message

PR #120, opened 2018-11-10
Fix property getting and setting on undefined and null

openorclose:
Reading and writing values to `null` and `undefined` (or any other cases since I just wrapped it in a try catch) fixed.


martin-henz:
Folks can you take a look at this? I don't have my hands on the interpreter so can't approve and merge. @yyc can you take a look?

PR #127, opened 2018-11-15
Improve error message interop with builtins (resolves #122)

### LINKED ISSUE #122: Error regarding interoperating with builtins.
Error messages not clear. Managed to reduce to the following:

`function f(a){
    return x;
}

map(f, list(1,2,3));

//Can make this worse by somehow bouncing between calls.

/*function g(a){
    return x;
}

function f(a){
    return map(g, list(1));
}

map(f, list(1));*/`

This creates the feeling that there's more than 1 error. In practice its a single error that gets wrapped around multiple times.

Traced it to here:

https://github.com/source-academy/js-slang/blob/8050f876b197c929b5c3bd5d47bda8390e29acf1/src/interpreter.ts#L579

thomastanck:


openorclose:
Looks good. 

I was playing around with providing a stack trace (like Chrome console) few weeks ago, so that it's clearer what function calls led to the error in the specified line. Unfortunately, 

- I was unable to figure out how to catch stackoverflows
- Tail call optimisation results in function calls at a tail position not being traceable. (Can probably be solved by storing the call of a tail position function somewhere and referencing it later.)

Just a note, since providing a more complete trace of the error is a good to have.

PR #137, opened 2018-12-05
Block get and set access to inherited properties of objects (Resolves #81)

### LINKED ISSUE #81: (Critical) eval exposed in Source 3
```javascript
pair.constructor("return window;")();
```
Basically any code can be run.

jiachen247:
I don't know if there is any legitimate reason for preserving the entire prototype chain for objects in source.

This fix checks and see if the property being accessed exist on the prototype chain and throws and error if it is.

This fix will block all calls to inherited functions such as `.constructor` (as referenced in #81) and other functions like slice that are undocumented features in source.

This however does not block access to properties local to the object like `.length` for arrays.

I can't think of a reason why students would need to need access inherited properties in this 1101s? and i believe this fix will clean and block a lot of undocumented features out.

Are there any unintended side effects of such a fix?

### DIFF
--- a/src/__tests__/interpreter-errors.ts
+++ b/src/__tests__/interpreter-errors.ts
@@ -87,17 +87,17 @@ test('Error when deeply accessing property on undefined', () => {
   })
 })
 
-test('In case a function ever returns null, should throw an error as well', () => {
-  const code = `
-    const myNull = pair.constructor("return null;")();
-    myNull.prop;
-   `;
-  const context = mockContext(4)
-  const promise = runInContext(code, context, { scheduler: 'preemptive' })
-  return promise.then(obj => {
-    expect(obj).toMatchSnapshot()
-    expect(obj.status).toBe('error')
-    expect(context.errors).toMatchSnapshot()
-    expect(parseError(context.errors)).toBe('Line 3: Cannot read property prop of null')
-  })
-})
+// test('In case a function ever returns null, should throw an error as well', () => {
+//   const code = `
+//     const myNull = pair.constructor("return null;")();
+//     myNull.prop;
+//    `;
+//   const context = mockContext(4)
+//   const promise = runInContext(code, context, { scheduler: 'preemptive' })
+//   return promise.then(obj => {
+//     expect(obj).toMatchSnapshot()
+//     expect(obj.status).toBe('error')
+//     expect(context.errors).toMatchSnapshot()
+//     expect(parseError(context.errors)).toBe('Line 3: Cannot read property prop of null')
+//   })
+// })
--- a/src/__tests__/__snapshots__/interpreter-errors.ts.snap
+++ b/src/__tests__/__snapshots__/interpreter-errors.ts.snap
@@ -135,33 +135,6 @@ Array [
 ]
 `;
 
-exports[`In case a function ever returns null, should throw an error as well 1`] = `
-Object {
-  "status": "error",
-}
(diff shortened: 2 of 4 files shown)


openorclose:
`in` is not allowed on primitives, so `1..constructor;` will fail silently.

martin-henz:
**Are there any unintended side effects of such a fix?**

I think it's a good move, because it plugs a lot of holes.

jiachen247:
Good catch @openorclose! used `typeof obj[prop] !== 'undefined'` instead.

src/__tests__/interpreter-errors.ts
```diff
@@ -87,17 +87,17 @@ test('Error when deeply accessing property on undefined', () => {
...
-  const promise = runInContext(code, context, { scheduler: 'preemptive' })
-  return promise.then(obj => {
-    expect(obj).toMatchSnapshot()
-    expect(obj.status).toBe('error')
-    expect(context.errors).toMatchSnapshot()
-    expect(parseError(context.errors)).toBe('Line 3: Cannot read property prop of null')
-  })
-})
\ No newline at end of file
+// test('In case a function ever returns null, should throw an error as well', () => {
```
thomastanck: Don't remove this test case. Just rewrite it to
```
null.prop;
```

`null` will be allowed from next semester onwards.

jiachen247: @thomastanck i commented out the test case cause theres no way to create null in source (at least not yet). we can bring back the test once we introduce null?

thomastanck:
Blocked until we add objects back in as a disabled but maintained feature (chapter 100)0

thomastanck:
> `in` is not allowed on primitives, so `1..constructor;` will fail silently.
> 
> As for unintended side effects, I can't see any.

I just noticed that `1..constructor` is probably a syntax error. I get this in Firefox:

maim

openorclose:
Yeah window..constructor is. But I meant it as (1). constructor, which is the same as 1.. constructor

thomastanck:
> Yeah window..constructor is. But I meant it as (1). constructor, which is the same as 1.. constructor

I see, then testcases need to be added for these. @jiachen247 could you add some test cases for accessing properties of primitives like number or boolean etc? Also, `null.prop` now works on chapter 100, could you re-enable that test case?

We also linted all the files using prettier, it'd be great if you could resolve conflicts and format your files as well.

.gitignore
```diff
@@ -2,3 +2,8 @@ node_modules
 *.js
 *.map
 dist/
+.idea/
+*.lock
```
thomastanck: Don't add `yarn.lock` to the .gitignore! Also remove these empty lines

src/__tests__/interpreter-errors.ts
```diff
@@ -519,3 +519,60 @@ test('Runtime error when redeclaring variable', () => {
...
+  const promise = runInContext(code, context, { scheduler: 'preemptive' })
+  return promise.then(obj => {
+    expect(obj).toMatchSnapshot()
+    expect(obj.status).toBe('error')
+    expect(context.errors).toMatchSnapshot()
+    expect(parseError(context.errors)).toBe('Line 2: Cannot read property prop of undefined')
+  })
+})
+
+test('Error when accessing inherited property', () => {
```
thomastanck: Add error for trying to access property of a builtin as well.

PR #143, opened 2018-12-08
Changed empty list to null.

openorclose:
Hopefully all the things depending on empty lists have been updated too.

### DIFF
--- a/src/__tests__/lists.ts
+++ b/src/__tests__/lists.ts
@@ -1,0 +1,257 @@
+import { mockContext } from '../mocks/context'
+import { parseError, runInContext } from '../index'
+import { Finished } from '../types'
+
+test('list creates list', () => {
+  const code = `
+    function f() { return 1; }
+    list(1, 'a string ""', () => a, f, true, 3.14);
+  `
+  const context = mockContext(2)
+  const promise = runInContext(code, context, { scheduler: 'preemptive' })
+  return promise.then(obj => {
+    expect(obj).toMatchSnapshot()
+    expect(parseError(context.errors)).toBe('')
+    expect(obj.status).toBe('finished')
+  })
+})
+
+test('pair creates pair', () => {
+  const code = `
+    pair(1, 'a string ""');
+  `
+  const context = mockContext(2)
+  const promise = runInContext(code, context, { scheduler: 'preemptive' })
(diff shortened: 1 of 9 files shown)


src/__tests__/lists.ts
```diff
@@ -0,0 +1,257 @@
...
+  const context = mockContext(3)
+  const promise = runInContext(code, context, { scheduler: 'preemptive' })
+  return promise.then(obj => {
+    expect(obj).toMatchSnapshot()
+    expect(parseError(context.errors)).toBe('')
+    expect(obj.status).toBe('finished')
+    expect((obj as Finished).value).toBe(10)
+  })
+})
+
```
thomastanck: You should test `equal` as well if not `equal` returning `true` for all input would actually pass all the test cases :o

src/stdlib/list.ts
```diff
@@ -14,7 +14,7 @@ declare global {
 // Author: Martin Henz
 // Translated to TypeScript by Evan Sebastian
 
-export type List = Value[]
+export type List = Value[] | null
```
thomastanck: This shouldn't be changed. Would imply that stuff like `head(null)` should work, but it actually throws an error. Tbh half the type annotations don't really make sense anyway since we are working off SICP, loads of stuff go straight to `any` type haha.

If you really want to, make `is_pair` a [type guard] so that you don't need to spam `xs!` everywhere.

[type guard]: https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types

src/createContext.ts
```diff
@@ -156,14 +158,7 @@ export const importBuiltins = (context: Context, externalBuiltIns: CustomBuiltIn
...
     ) {
-      const res = []
-      var i = 0
-      while (!(args.length === 0)) {
-        res[i] = args[0]
-        i = i + 1
-        args = args[1]
-      }
-      return fun.apply(fun, res)
+      return fun.apply(fun, list_to_vector(args))
```
thomastanck: Good catch.

src/stdlib/list.ts
openorclose: Yup. I've changed it so that some functions like map return `List | null` instead. But it still looks quite ugly.

PR #144, opened 2018-12-08
Make || and && shorthand for ?: and improve TCO

thomastanck:
|| and && are now transformed into ?: before being evaluated.
TCO now handles tail calls hidden inside ?:, ||, && expressions properly.


martin-henz:
I like this a lot. No need to talk about && and || any longer.

src/interpreter.ts
```diff
@@ -480,16 +487,37 @@ export const evaluators: { [nodeType: string]: Evaluator<es.Node> } = {
...
+      return new ReturnValue(undefined)
+    }
+
+    // If we have a conditional expression, reduce it until we get something else
+    while (return_expression.type === 'LogicalExpression' ||
+      return_expression.type === 'ConditionalExpression') {
+      if (return_expression.type === 'LogicalExpression') {
+        return_expression = transformLogicalExpression(return_expression)
+      }
+      const test = yield* evaluate(return_expression.test, context)
```
openorclose: this part looks like duplication of code from the IfStatement. Maybe can refactor out into a function just returning the resulting node, and use the refactored function here in ReturnStatement, and evaluate the result of the refactored function inside IfStatement? Something like 

```javascript

function getResultantNodeFromIf(args) {
  return node.consequent or node.alternate;
}

ReturnStatement() {
while() {
 ...
 return_expression = getResultantNodeFromIf(args);
 ...
}
}

IfStatement() {
  yield* evaluate(getResultantNodeFromIf(args));
}

And return_expression should be returnExpression instead.

thomastanck: Ah, you are right. Thanks, shall refactor now.

thomastanck:
Blocked until #140 is merged as that one handles proper error handling of infinite recursion.

src/interpreter.ts
thomastanck: Can't deduplicate this code as I run into a typescript bug. Can't find a way to resolve...

PR #145, opened 2018-12-08
Updated display() function to return values

### LINKED ISSUE #97: Function display should return its argument
This will make the function much more useful for debugging. You can then just wrap any expression with "display(...)" without changing the meaning of the program.

urieltan:
Display function now returns values so that 'undefined' won't be displayed.
Resolves #97


martin-henz:
Great!

PR #147, opened 2018-12-09
Make dot abbreviations work again in chapter 100

thomastanck:
FYI chapter 100 is where we put features we want to disable for now but may want to bring back in future versions.


src/rules/noDotAbbreviation.ts
```diff
@@ -24,6 +24,8 @@ export class noDotAbbreviationError implements SourceError {
 const noDotAbbreviation: Rule<es.MemberExpression> = {
   name: 'no-dot-abbreviation',
 
+  disableOn: 100,
```
thomastanck: This is the only real change. The rest are all updating test cases.

PR #153, opened 2018-12-11
Remove dead code from node.ts

openorclose:


thomastanck:
Consider removing the `__id` tagging from the walker parser.ts as well

PR #154, opened 2018-12-11
More coverage

thomastanck:


src/interpreter.ts
```diff
@@ -403,17 +403,13 @@ export const evaluators: { [nodeType: string]: Evaluator<es.Node> } = {
       if (value instanceof ReturnValue || value instanceof TailCallReturnValue) {
         break
       }
-      if (node.update) {
-        yield* evaluate(node.update, context)
-      }
-      test = node.test ? yield* evaluate(node.test, context) : true
+
+      yield* evaluate(updateNode, context)
+      test = yield* evaluate(node.test!, context)
```
openorclose: this is redundant no? you already evaluate it in the while loop

PR #161, opened 2018-12-13
Implement a better stringify, etc.

thomastanck:
- Better stringify
- No more toString
- toString for functions now overridden instead
- Got rid of `__SOURCE__`, use `defineBuiltin` to hide implementation


src/interop.ts
```diff
@@ -26,48 +25,125 @@ export const closureToJS = (value: Value, context: Context, klass: string) => {
...
-      return arrayToString(value, length)
-    }
-  } else if (typeof value === 'string') {
-    return `"${value}"`
-  } else if (typeof value === 'undefined') {
-    return 'undefined'
-  } else if (typeof value === 'function') {
-    if (value.__SOURCE__) {
-      return `function ${value.__SOURCE__} {\n\t[implementation hidden]\n}`
+function indentify(indent: string, s: string): string {
```
openorclose: We should stick to camelCase like the rest of the interpreter.

src/interop.ts
```diff
@@ -26,48 +25,125 @@ export const closureToJS = (value: Value, context: Context, klass: string) => {
...
+  return s
+    .split('\n')
+    .map(v => indent + v)
+    .join('\n')
+}
+
+function repeat(s: string, times: number): string {
+  return new Array(times + 1).join(s)
```
openorclose: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat

src/interop.ts
```diff
@@ -26,48 +25,125 @@ export const closureToJS = (value: Value, context: Context, klass: string) => {
...
+  splitline_threshold = 80
+): string => {
+  let ancestors = new Set()
+  const indent = makeIndent(indent_)
+  let arr_prefix = '[' + indent.substring(1)
+  let obj_prefix = '{' + indent.substring(1)
+  let arr_suffix = indent.substring(0, indent.length - 1) + ']'
+  let obj_suffix = indent.substring(0, indent.length - 1) + '}'
+
+  const helper = (value: Value, indent_level: number = 0): string => {
```
openorclose: helper sounds generic. not sure if there's a more descriptive name for it.

src/interop.ts
```diff
@@ -26,48 +25,125 @@ export const closureToJS = (value: Value, context: Context, klass: string) => {
...
+    } else if (Array.isArray(value)) {
+      if (ancestors.size > MAX_LIST_DISPLAY_LENGTH) {
+        return '...<truncated>'
+      }
+      ancestors.add(value)
+      let value_strs = value.map(v => helper(v, 0))
+      let multiline =
+        indent !== '' &&
+        (value_strs.join(', ').length > splitline_threshold ||
+          value_strs.some(s => s.indexOf('\n') !== -1))
```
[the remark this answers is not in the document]
openorclose: Also, this is used again later. An `isMultiline` function could be more useful.

src/interop.ts
```diff
@@ -26,48 +25,125 @@ export const closureToJS = (value: Value, context: Context, klass: string) => {
...
+        let key_str = helper(entry[0], 0)
+        let val_str = helper(entry[1], 0)
+        if (val_str.indexOf('\n') === -1) {
+          // Single line
+          return key_str + ': ' + val_str
+        } else {
+          return key_str + ':\n' + indentify(indent, val_str)
+        }
+      })
+      let multiline =
```
openorclose: Place where multiline is used again.

src/interpreter-errors.ts
```diff
@@ -68,7 +68,7 @@ export class MaximumStackLimitExceeded extends RuntimeSourceError {
     CallExpression(node: any, state: any) {
       state.write(node.callee.name)
       state.write('(')
-      const argsRepr = node.arguments.map((args: any) => toString(args.value))
+      const argsRepr = node.arguments.map((args: any) => stringify(args.value))
```
openorclose: looks like we should rename `args` to `arg`.

src/createContext.ts
```diff
@@ -51,44 +53,50 @@ export const defineSymbol = (context: Context, name: string, value: Value) => {
   })
 }
 
+export const defineBuiltin = (context: Context, name: string, f: Function) => {
+  let wrapped = (...args: any) => f(...args)
+  wrapped.toString = () => {
+    const params = getParameterNames(f).join(', ')
```
openorclose: Recomputing the string representation each time toString is called gives the impression that the function might change signature when reading the code. It should probably be computed outside of the function.

src/createContext.ts
```diff
@@ -51,44 +53,50 @@ export const defineSymbol = (context: Context, name: string, value: Value) => {
...
+    return `function ${f.name}(${params}) {\n\t[implementation hidden]\n}`
```
openorclose: We should use `name` instead of `f.name`. That's what the students will use. For example, `stringify(math_sin)` will now display that the name of the function is `sin` instead of `math_sin`, which is confusing.

src/createContext.ts
```diff
@@ -51,44 +53,50 @@ export const defineSymbol = (context: Context, name: string, value: Value) => {
   })
 }
 
+export const defineBuiltin = (context: Context, name: string, f: Function) => {
```
openorclose: Should we just have one function called `defineBuiltin`? then if it is a function we add the toString, if not we leave it as it is. This would simplify the logic of externalSymbols and importing math libraries.

src/createContext.ts
```diff
@@ -51,44 +53,50 @@ export const defineSymbol = (context: Context, name: string, value: Value) => {
...
+  }
+  defineSymbol(context, name, wrapped)
+}
+
 export const importExternalSymbols = (context: Context, externalSymbols: string[]) => {
   ensureGlobalEnvironmentExist(context)
 
   externalSymbols.forEach(symbol => {
-    defineSymbol(context, symbol, GLOBAL[symbol])
+    if (typeof GLOBAL[symbol] === 'function') {
```
openorclose: only have one way to definieBuiltin.

src/createContext.ts
```diff
@@ -98,9 +106,9 @@ export const importBuiltins = (context: Context, externalBuiltIns: CustomBuiltIn
       if (objs.hasOwnProperty(i)) {
         const val = objs[i]
         if (typeof Math[val] === 'function') {
-          defineSymbol(context, 'math_' + val, Math[val].bind())
+          defineBuiltin(context, 'math_' + val, Math[val].bind())
```
openorclose: I believe the `.bind` is useless. Also can simplify if there is only one way to define a builtin.

PR #171, opened 2018-12-24
Add strict types for + and member expressions

thomastanck:


src/utils/rttc.ts
```diff
@@ -37,6 +37,9 @@ const typeOf = (v: Value) => {
 const isNumber = (v: Value) => typeOf(v) === 'number'
```
openorclose: The above,

https://github.com/source-academy/js-slang/blob/b5d72aed774cc2d58ed7d474c3ad371b73f066e0/src/utils/rttc.ts#L30-L36

can be removed to simply use the typeof operator, as Closures are now a subclass of Function.

src/utils/rttc.ts
```diff
@@ -37,6 +37,9 @@ const typeOf = (v: Value) => {
...
 const isString = (v: Value) => typeOf(v) === 'string'
 const isBool = (v: Value) => typeOf(v) === 'boolean'
+const isObject = (v: Value) =>
+  typeOf(v) === 'object' && !(v instanceof Closure) && !(v instanceof Function) && !Array.isArray(v)
```
openorclose: Similarly here, the check for Closure is redundant. I'm not sure if you meant to have null be an object here, but note that null will return true for isObject.

src/utils/rttc.ts
```diff
@@ -37,6 +37,9 @@ const typeOf = (v: Value) => {
...
+const isArray = (v: Value) => typeOf(v) === 'object' && Array.isArray(v)
```
openorclose: I think the typeOf = object is redundant.

src/utils/rttc.ts
```diff
@@ -105,3 +98,14 @@ export const checkIfStatement = (context: Context, test: Value) => {
   const node = context.runtime.nodes[0]
   return isBool(test) ? undefined : new TypeError(node, ' as condition', 'boolean', typeOf(test))
 }
+
+export const checkMemberAccess = (context: Context, obj: Value, prop: Value) => {
```
openorclose: Note that `null.prop` is not getting caught here. Not sure if we should pretend that null is an object for Source... If we are saying that null is not an object in Source, we could make isObject return false for null so that an error "expected an object or array, got null instead" here?

Not too sure about this.

src/utils/testing.ts
```diff
@@ -255,17 +255,29 @@ export function expectParsedErrorNoSnapshot(
   ).resolves
 }
 
+function evalWithBuiltins(code: string, testBuiltins?: TestBuiltins) {
+  // Ugly, but if you know how to `eval` code with some builtins attached, please change this.
+  if (testBuiltins === undefined) {
```
openorclose: I think we should use default parameters instead.

src/utils/__tests__/rttc.ts
```diff
@@ -216,3 +226,39 @@ test('Invalid ternary/if test expressions return TypeError', () => {
...
+test('Valid member expressions are OK', () => {
+  const operatorValues = [[obj, str], [arr, num]]
+  const context = mockRuntimeContext()
+  const errors = operatorValues.map(opVals => rttc.checkMemberAccess(context, opVals[0], opVals[1]))
+  errors.map(error => expect(error).toBeUndefined())
+})
+
+test('Invalid member expressions return TypeError', () => {
+  const operatorValues = [
+    [obj, num],
```
openorclose: Missing out something * undefined and undefined * something check.

src/utils/__tests__/rttc.ts
```diff
@@ -6,6 +6,8 @@ const num = 0
 const bool = true
 const str = ' '
 const func = mockClosure()
```
openorclose: we should test both `mockClosure` and builtin functions, as well as `undefined`.

src/utils/testing.ts
```diff
@@ -255,17 +255,29 @@ export function expectParsedErrorNoSnapshot(
...
+    testBuiltins = {}
+  }
+  let evalstring = ''
+  for (const key in testBuiltins) {
```
openorclose: you can do 

```javascript
function evalWithContext(code) {
return eval(code);
}
return evalWithContext.call(testBuiltins)
```

src/utils/__tests__/rttc.ts
thomastanck: Done, and added null as well.

src/utils/rttc.ts
thomastanck: Made null.prop a type error instead of a property access error.

src/utils/rttc.ts
thomastanck: Rewrote typeOf to show 'null' and 'array' for `null` and arrays respectively, removed special case for function.

PR #174, opened 2019-02-10
Native MVP

openorclose:
# Features:

- [x] Native!
- [x] Builtins work! 
- [ ] Enforcing of strict types (`'' * ''` still works)
- [ ] Error handling (just throws a bunch of unhelpful JS errors)
- [ ] stringifying functions (shows the transpiled code currently) 

# How to use: 

Change 

```typescript
  scheduler: 'async',
  steps: 1000,
}
```
in `src/index.ts` 


src/__tests__/native-tests.ts
```diff
@@ -0,0 +1,149 @@
+import { runInContext } from '../index'
+import { mockContext } from '../mocks/context'
+import { Finished } from '../types'
+import { expectResult, stripIndent } from '../utils/testing'
+
+function nativeTest(code: string, expectedResult: any) {
```
thomastanck: We can probably move this to testing utils.

src/index.ts
```diff
@@ -33,14 +39,26 @@ export function runInContext(
   context.errors = []
   const program = parse(code, context)
   if (program) {
-    const it = evaluate(program, context)
-    let scheduler: Scheduler
-    if (theOptions.scheduler === 'async') {
-      scheduler = new AsyncScheduler()
+    if (theOptions.isNativeRunnable) {
```
thomastanck: Is it better to use a condition here or should we have a separate `runInNativeContext` or `nativeRunInContext` function instead? Not sure what the plans are for interoperability.

src/__tests__/native-tests.ts
```diff
@@ -0,0 +1,149 @@
+import { runInContext } from '../index'
```
thomastanck: I'm getting a weird bug where parse errors on native sometimes get `Line -1`. Can you confirm?

maim

src/__tests__/native-tests.ts
```diff
@@ -0,0 +1,149 @@
...
+        return y;
+      } else {
+        return f(x-1, y+1);
+      }
+    }
+    f(5000, 5000);
+  `,
+    10000
+  )
+)
```
thomastanck:
```suggestion
)

test(
  'Tail calls in mixed tail-call/non-tail-call recursion work',
  nativeTest(
    stripIndent`
    function f(x, y, z) {
      if (x <= 0) {
        return y;
      } else {
        return f(x-1, y+f(0, z, 0), z);
      }
    }
    f(5000, 5000, 2);
  `,
    15000
  )
)
```

If I add this test case, it fails on native but works on interpreter. Change `test` to `test.skip` if you don't want to fix this test case yet in this PR.

src/__tests__/native-tests.ts
openorclose: This is actually not caused by the Source parser.

The source parser only checks if `let b = 0;` is valid syntax, which it is.

After that, the code to 'put back' b in the scope gets prepended like 

```javascript
// previous globals;
let b = previous value of b;
// current line
let b =0;
```

So it gets evaled and throws the mentioned error. 

Any error caused by eval currently has an unknown location (since providing the actual location wouldn't make sense as it would be the location of the error of the transpiled code).

thomastanck: I see. Can we print Line unknown instead? Doesnt need to be in this PR.

src/__tests__/native-tests.ts
openorclose: Yeah, I'll have to think a bit about this fix..

src/index.ts
openorclose: I'd wait to see how debug and environment model works first.

src/constants.ts
```diff
@@ -1,5 +1,7 @@
 import * as es from 'estree'
 
+export const GLOBAL = typeof window === 'undefined' ? global : window
+export const NATIVE_STORAGE_GLOBAL = '$$NATIVE_STORAGE'
```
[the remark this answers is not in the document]
openorclose: Nope, I'm doing window['$$NATIVE_STORAGE'] to access.

There should be a better name for it though.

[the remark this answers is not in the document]
thomastanck: Doesn't `window[someSymbol]` also work?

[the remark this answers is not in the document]
openorclose: Yeah it does when there's access to the Symbol.

But during eval I'm also doing eval('$$NATIVE_STORAGE') to access it, and the Symbol can't be accessed from there.

src/__tests__/native-tests.ts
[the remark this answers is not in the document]
openorclose: Yup. All the tests should give the same results for native and interpreted ideally.

src/constants.ts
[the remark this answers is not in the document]
thomastanck: I see. You're right.

src/index.ts
thomastanck: Roger. For now let's shelve the issue until we get more info.

src/__tests__/native-tests.ts
openorclose: Fixed! (I hope)

src/transformer.ts
```diff
@@ -0,0 +1,209 @@
...
+      test: transform(node.test),
+      consequent: transform(node.consequent),
+      alternate: transform(node.alternate)
+    }
+  },
+  ArrowFunctionExpression(node: es.ArrowFunctionExpression) {
+    if (node.expression) {
+      return transform(
+        create.blockArrowFunction(
+          node.params.map(transform),
```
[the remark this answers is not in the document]
openorclose: Oh right, they are.

src/transformer.ts
```diff
@@ -0,0 +1,209 @@
+import * as es from 'estree'
+import * as create from './utils/astCreator'
+
+export function transform(node: es.Node) {
```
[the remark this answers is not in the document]
openorclose: This is just (mostly) cloning the nodes. When the error checking comes in the original nodes will need to be accessed so transform is just groundwork for that.

thomastanck:
We should make sure these lines are definitely covered.

https://coveralls.io/builds/21608528/source?filename=src%2Futils%2FproperTailCalls.ts

martin-henz:
@thomastanck can you take a (hopefully final) look at this?

PR #175, opened 2019-02-19
Refactoring

SwampertX:


src/closure.ts
```diff
@@ -20,8 +20,8 @@ const closureToJS = (value: Closure, context: Context, klass: string) => {
   })
   Object.setPrototypeOf(DummyClass, () => undefined)
   Object.defineProperty(DummyClass, 'Inherits', {
-    value: (Parent: Value) => {
```
openorclose: This is for OOP. Doesn't make sense for it to be renamed to tail.

src/createContext.ts
```diff
@@ -32,21 +32,21 @@ export const createEmptyContext = <T>(
   runtime: createEmptyRuntime()
 })
 
-export const ensureGlobalEnvironmentExist = (context: Context) => {
+export const ensureGlobalFrameExist = (context: Context) => {
```
openorclose: Why is this being renamed to frame?

src/closure.ts
martin-henz: I agree. @SwampertX please change this back to Parent.

src/createContext.ts
martin-henz: Yes, this should be ensureGlobalEnvironmentExist.

Another question: Why does a Context need multiple Environments?

PR #176, opened 2019-02-21
added toggle for verbose error messages

LeowWB:
actual verbose error messages not included as part of this pull request


src/index.ts
```diff
@@ -29,10 +43,28 @@ export function runInContext(
   context: Context,
   options: Partial<IOptions> = {}
 ): Promise<Result> {
+  function getFirstLine(theProgram: Program) {
+    const firstLineOfProgram = theProgram.body[0] as ExpressionStatement
+    if (!!firstLineOfProgram) {
+      const firstLineExpression = firstLineOfProgram.expression as Literal
```
openorclose: Similar to above, I think you should explicitly check expression.type === 'Literal'

src/index.ts
```diff
@@ -29,10 +43,28 @@ export function runInContext(
   context: Context,
   options: Partial<IOptions> = {}
 ): Promise<Result> {
+  function getFirstLine(theProgram: Program) {
+    const firstLineOfProgram = theProgram.body[0] as ExpressionStatement
+    if (!!firstLineOfProgram) {
```
openorclose: I think it's better if you explicitly check if the first line exists, then check if the first line type === 'ExpressionStatement'.

src/interpreter-errors.ts
```diff
@@ -107,7 +112,8 @@ export class UndefinedVariable extends RuntimeSourceError {
   }
 
   public elaborate() {
-    return 'TODO'
+    return `Before you can call ${this.name}, you need to declare it as a variable or a constant.
```
openorclose: Call usually refers to a function call. Its usage here could be confusing.

src/interpreter-errors.ts
```diff
@@ -121,7 +127,7 @@ export class UnassignedVariable extends RuntimeSourceError {
   }
 
   public elaborate() {
-    return 'TODO'
+    return `Please assign a value to ${this.name} first. Example:\n\t${this.name} = 42;`
```
openorclose: This is wrong. Example of when this error occurs.

```javascript

const a = 1;
const f = () => {
  const i_think_i_can_access_outer_a_here = a; // UnassignedVariableError occurs here
  // other stuff;
  const a = 'another use of a';
}
f();
```

This is due to the proper scoping of variables, so access to any variables that are ever declared in the current scope is blocked until the statement where it is being declared is evaluated.

src/interpreter-errors.ts
```diff
@@ -1,4 +1,9 @@
-/* tslint:disable:max-classes-per-file */
+/*
```
openorclose: Not true. InvalidNumberOfArguments is not a JS error but a Source one for example. Should this mean that we should try to separate these away from the JS ones?

src/rules/noDeclareMutable.ts
```diff
@@ -1,3 +1,4 @@
+import astr = require('astring')
```
openorclose: Why is this require instead of import?

thomastanck:
You'll have to sync master after #174 is merged and update tests. There is a refactor to the test API.

src/repl/repl.ts
```diff
@@ -1,24 +1,30 @@
...
+      }
     }
-  })
+  )
 }
 
 function main() {
-  const chapter = process.argv.length > 2 ? parseInt(process.argv[2], 10) : 1
+  // TODO const chapter alternative case should be 1
+  const chapter = process.argv.length > 2 ? parseInt(process.argv[2], 10) : 4
```
thomastanck: Did you forget to undo this change or was this intended?

src/repl/repl.ts
```diff
@@ -1,24 +1,30 @@
-import repl = require('repl')
```
thomastanck: Try not to make refactors in the same PR as a feature PR. It adds additional noise to the PR which reviewers would have to filter out.

src/types.ts
```diff
@@ -1,4 +1,11 @@
+/*
+	This file contains definitions of some interfaces and classes that are used in Source (such as
+	error-related classes).
+*/
+
+/* tslint:disable:interface-functionName max-classes-per-file */
```
thomastanck: It's slightly strange that you had to disable this lint for this file even though it appears that this file didn't change. Was this line added because you made a change in this file and later reverted it?

src/index.ts
```diff
@@ -1,3 +1,4 @@
+import { ExpressionStatement, Literal, Program } from 'estree'
```
thomastanck: Can we have some test cases to check that the toggle works?

I have a hunch that it doesn't. Line 47 may want to check for `'Directive'`. https://github.com/estree/estree/blob/master/es5.md#directive

src/index.ts
```diff
@@ -29,10 +43,32 @@ export function runInContext(
...
 ): Promise<Result> {
+  function getFirstLine(theProgram: Program) {
+    if (theProgram.body[0] && theProgram.body[0].type === 'ExpressionStatement') {
+      const firstLineOfProgram = theProgram.body[0] as ExpressionStatement
+      if (firstLineOfProgram.expression.type === 'Literal') {
+        const firstLineExpression = firstLineOfProgram.expression as Literal
+        if (!!firstLineExpression) {
+          return firstLineExpression.value
+        } else {
+          return undefined
```
thomastanck: You can omit all the `return undefined` here as this function is small and it's obvious that the function returns quickly if the conditions are not met.

src/types.ts
LeowWB: You're about right. I added the comment block at the top to help myself understand what was going on when I was still figuring things out on my own. During a merge I decided to keep that block there, while also accepting incoming changes (the comment that disabled the linter). At the time I figured there'd be no harm in leaving a one-line comment in the code. That said, what would you suggest now? Should I remove line 6?

LeowWB: You're about right. I added the comment block at the top to help myself understand what was going on when I was still figuring things out on my own. During a merge I decided to keep that block there, while also accepting incoming changes (the comment that disabled the linter). At the time I figured there'd be no harm in leaving a one-line comment in the code. That said, what would you suggest now? Should I remove line 6?

thomastanck: Yes. You can view the blame for this file to see where these lines come from. (either view it on github or use `git blame`)

src/repl/repl.ts
LeowWB: Nope, it was an oversight on my part. Fixed it.

src/types.ts
LeowWB: It seems to have come from #175 ? Either way, I'll remove it.

src/index.ts
LeowWB: The toggle works, though I agree it doesn't always work as expected, and there's a nuanced difference in how you would toggle it for interpreter errors and errors that result from breaking Source specs. I have yet to fix that. Meanwhile I'll start pulling over commits from my other branch which does have the test cases. I've also implemented the Directive check, though I had to implement my own interface since the ESTree index that we have doesn't seem to export any such interface.

src/repl/repl.ts
LeowWB: Noted, I'll keep that in mind for future PRs.

src/index.ts
```diff
@@ -35,10 +49,26 @@ export function runInContext(
   context: Context,
   options: Partial<IOptions> = {}
 ): Promise<Result> {
+  function getFirstLine(theProgram: Program) {
+    if (theProgram.body[0] && theProgram.body[0].type === 'ExpressionStatement') {
+      const firstLineOfProgram = theProgram.body[0] as ExpressionStatement
+      const theDirective = (firstLineOfProgram as Directive).directive
+      if (theDirective) {
```
thomastanck: I would check for undefined explicitly. 
maim

src/index.ts
```diff
@@ -35,10 +49,26 @@ export function runInContext(
...
+    }
+
+    return undefined
+  }
   const theOptions: IOptions = { ...DEFAULT_OPTIONS, ...options }
   context.errors = []
   const program = parse(code, context)
   if (program) {
+    if (getFirstLine(program) === 'enable verbose') {
+      verboseErrors = true
```
thomastanck: This kind of `if true then set to true else set to false` pattern is not as good as simply writing `verboseErrors = getFirstLine(program) === 'enable verbose'`

On the other hand, as commented above, might be good to extract this function out for testing

src/index.ts
```diff
@@ -21,11 +22,24 @@ const DEFAULT_OPTIONS: IOptions = {
...
     const line = error.location ? error.location.start.line : '<unknown>'
+    const column = error.location ? error.location.start.column : '<unknown>'
     const explanation = error.explain()
-    return `Line ${line}: ${explanation}`
+    const elaboration = error.elaborate()
+
+    if (verbose) {
+      // TODO currently elaboration is just tagged on to a new line after the error message itself. find a better
+      // way to display it.
+      return `Line ${line}, Column ${column}: ${explanation}. ${elaboration}\n`
```
thomastanck: I think your \n needs to be placed behind the elaboration not after it.

src/interpreter-errors.ts
```diff
@@ -107,7 +107,10 @@ export class UndefinedVariable extends RuntimeSourceError {
   }
 
   public elaborate() {
-    return 'TODO'
+    return `Before you can read the value of ${
+      this.name
+    }, you need to declare it as a variable or a constant.
+		You can do this using the let or const keywords.`
```
thomastanck: Is the second sentence supposed to be on a new line? If so you should get rid of the leading spaces. Spaces are preserved in backtick strings. (the ones surrounded by `)

src/interpreter-errors.ts
```diff
@@ -117,11 +120,17 @@ export class UnassignedVariable extends RuntimeSourceError {
...
 
   public elaborate() {
-    return 'TODO'
+    return `If you're trying to access the value of ${
+      this.name
+    } from an outer scope, please rename the inner ${
+      this.name
+    }. An easy way to avoid this issue in future would be to avoid declaring any variables or constants with the name ${
+      this.name
+    } in the same scope. `
```
thomastanck: Extra space at the end of the string.

src/__tests__/interpreter-errors.ts
```diff
@@ -13,6 +13,13 @@ test('Undefined variable error is thrown', () => {
   `).toMatchInlineSnapshot(`"Line 1: Name im_undefined not declared"`)
 })
 
+test('Undefined variable error is thrown - verbose', () => {
+  return expectParsedError(stripIndent`
+	"enable verbose";
+	im_undefined;
+	`).toMatchSnapshot()
```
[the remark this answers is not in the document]
LeowWB: My reason for using snapshots rather than inline snapshots was for readability - I didn't want the JSON content to make the test suite files messy. That said, you raise a good point, so I'm considering changing the inline snapshots to regular snapshots instead. Thoughts?

Also I feel like your last suggestion would be adding unnecessary functionality, since the interpreter has no actual use for such a function other than in testing.

src/index.ts
LeowWB: I'd be against removing the '\n', since the intention was to separate the stack trace from the error message itself, to enhance readability. I'll add another '\n' between the explanation and elaboration.

src/index.ts
LeowWB: Sorry, could you rephrase? You'd check for undefined on which name?

thomastanck:
Check on `theDirective`

openorclose:
Is there a reason verbose mode is not another option to be passed into runInContext? Currently native already works this way.

LeowWB:
> Is there a reason verbose mode is not another option to be passed into runInContext? Currently native already works this way.

The main reason is that the verbosity of error messages has little to do with how the program is being run, but rather the way the output is displayed to the user. A verbose error mode doesn't result in any change to the number of errors, or any other aspect of what the interpreter does in the process of running the program.

martin-henz:
I think we should keep such annotations out of the student programs. The reason: Project NGED is working on a collaborative editor. One user may want to have verbose error messages, while another one may want to have terse messages. I can foresee those two going back and forth.

Second reason: We want to keep the language js-slang clean. It should be a strict subset of JavaScript.

It's best to add an option to runInContext, as @openorclose suggests.

In order to test this in the frontend, you may want to look at this frontend branch, implementing an auto-run feature: https://github.com/source-academy/cadet-frontend/tree/auto-run-feature

We would for now add a similar button to the frontend. We can later think about user preference settings for such things.

thomastanck:
The decision was made to mimic the `"use strict";` directive quite some time ago, but we need to revisit that decision now that NATIVE is merged in. Perhaps we should merge this in first since I believe the other PR has changes that depends on this one. After both of these are merged, it shouldn't be hard to make a final PR switching to the option method.

src/__tests__/interpreter-errors.ts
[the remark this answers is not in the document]
thomastanck: Snapshot content should be treated as code, and in this case it's better to inline the snapshots for better viewing. Other snapshots are not inlined because of two reasons:

- they contain much much more information, so inlining would really hurt readability, and
- they act more as a regression test.

In this case we probably want to inline because we also want to test the behaviour.

I don't know if all tests currently follow this convention, but if you find any feel free to PR.

I agree with you @ the other suggestion. No need to overcomplicate the code.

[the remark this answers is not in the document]
thomastanck: On another note, please tell me when you're ready for review again.

martin-henz:
Sure. No problem to merge the current solution

"enable verbose";

The directive
"use strict"; 
itself is a bit of a trick. Expression statements like this are consistent with JavaScript syntax (and don't have any effect).

As Thomas suggests, let's re-visit this later.

martin-henz:
Looks like we need to resolve a conflict in the test cases; both branches added test cases in the same location.

PR #177, opened 2019-02-21
added verbose error messages and test cases/snapshots for verbose error messages

LeowWB:


src/interpreter-errors.ts
```diff
@@ -89,11 +94,20 @@ export class CallingNonFunctionValue extends RuntimeSourceError {
...
   }
 
   public elaborate() {
-    return 'TODO'
+    const calleeVal = this.callee
+    const calleeStr = stringify(calleeVal)
+    const elabStr = `Because ${calleeStr} is not a function, you cannot run ${calleeStr}().`
+    const multStr = `If you were planning to perform multiplication by ${calleeStr}, you need to use the * operator.`
+
+    if (typeof calleeVal !== 'number') {
```
openorclose: I would suggest swapping your if statement (easier to not have to parse negatives)

openorclose: Might want to check for Infinity/NaN too.

src/interpreter-errors.ts
```diff
@@ -117,39 +133,67 @@ export class UnassignedVariable extends RuntimeSourceError {
...
   public explain() {
-    return `Redeclaring name ${this.name}`
+    return `Redeclaring name ${this.name}.`
   }
 
   public elaborate() {
-    return 'TODO'
+    const elabStr = `Since ${
+      this.name
+    } has already been declared, you can assign a value to it without re-declaring.`
```
openorclose: To take into account trying to redeclare constants in Source? It might be confusing to suggest reassigning a value to a constant.

src/rules/noDeclareMutable.ts
```diff
@@ -21,7 +22,14 @@ export class NoDeclareMutableError implements SourceError {
   }
 
   public elaborate() {
-    return this.explain()
+    const name = (this.node.declarations[0].id as es.Identifier).name
+    let value = astr.generate(this.node.declarations[0].init)
+
+    if (value.length >= 15) {
```
openorclose: Why does only this particular rule have an arbitrary max length of 15 when reproducing code? I think it's better to be consistent and apply this rule to all other errors or to not have it at all.

src/rules/noDotAbbreviation.ts
```diff
@@ -17,7 +17,7 @@ export class NoDotAbbreviationError implements SourceError {
   }
 
   public elaborate() {
-    return ''
+    return "Source doesn't use object-oriented programming, so you don't need any dots in your code."
```
openorclose: Good to rephrase this. Dots are needed for numbers.

src/rules/noNull.ts
```diff
@@ -13,11 +13,11 @@ export class NoNullError implements SourceError {
   }
 
   public explain() {
-    return `null literals are not allowed`
+    return `null literals are not allowed.`
   }
 
   public elaborate() {
-    return this.explain()
+    return "They're not part of the Source specs."
```
openorclose: Should clarify that it's not part of Source 1 specs only?

src/rules/noUpdateAssignment.ts
```diff
@@ -13,11 +14,17 @@ export class NoUpdateAssignment implements SourceError {
...
+    return 'The assignment operator ' + this.node.operator + ' is not allowed. Use = instead.'
   }
 
   public elaborate() {
-    return ''
+    const leftStr = astr.generate(this.node.left)
+    const rightStr = astr.generate(this.node.right)
+    const newOpStr = this.node.operator.slice(0, -1)
+
+    const elabStr = `\n\t${leftStr} = ${leftStr} ${newOpStr} ${rightStr};`
```
openorclose:
```javascript
let a = 0;
a <<= 1;
```

would suggest `a = a << 1` when << is not allowed.

src/interpreter-errors.ts
```diff
@@ -89,11 +94,20 @@ export class CallingNonFunctionValue extends RuntimeSourceError {
...
   public explain() {
-    return `Calling non-function value ${stringify(this.callee)}`
+    return `Calling non-function value ${stringify(this.callee)}.`
   }
 
   public elaborate() {
-    return 'TODO'
+    const calleeVal = this.callee
+    const calleeStr = stringify(calleeVal)
+    const elabStr = `Because ${calleeStr} is not a function, you cannot run ${calleeStr}().`
```
openorclose: Is "you cannot run notfunction()" confusing, when maybe their code was `notfunction(1, 2, 3)`?

src/rules/noDeclareMutable.ts
LeowWB: I was thinking mainly because it has the potential to give you an entire function definition expression, which might lead to a messy error log. There's not as much of a need to have this in most other error messages, since many of them only have to deal with printing names. That said, I suppose the limit could be increased by quite a bit. I personally don't think it's good to not have a limit, though. Thoughts?

martin-henz: Is the "code" nicely formatted? If yes, and if "verbose" is turned on, and if we can scroll, I don't see much harm. If in a particular situation, there is "too much information", the user can turn off "verbose". 

Having no limits: In this case we are talking about the programs that the user wrote. In the big scheme of things, programs are naturally finite, and in practice rather small, so there might be not much actual harm in having no limit (if the program is nicely presented).

LeowWB: Fair point regarding disabling verbose; I'll remove the limit.

martin-henz:
Revert "Revert "Revert "Revert "Merge branch 'pr/8'""""

Ohmy

thomastanck:
Ready for review @openorclose

src/interpreter-errors.ts
```diff
@@ -134,30 +154,56 @@ export class UnassignedVariable extends RuntimeSourceError {
...
   }
 
   public explain() {
-    return `Redeclaring name ${this.name}`
+    return `Redeclaring name ${this.name}.`
   }
 
   public elaborate() {
-    return 'TODO'
+    if (this.writable) {
```
openorclose: Should explicitly check `=== true`

src/interpreter-errors.ts
```diff
@@ -134,30 +154,56 @@ export class UnassignedVariable extends RuntimeSourceError {
...
-    return `Expected ${this.expected} arguments, but got ${this.got}`
+    return `Expected ${this.expected} arguments, but got ${this.got}.`
   }
 
   public elaborate() {
-    return 'TODO'
+    const calleeStr = this.calleeStr
+    const pluralS = this.expected === 1 ? '' : 's'
+
+    return `Try calling function ${calleeStr} again in the same way, but with ${
```
openorclose: I think 'in the same way' should be removed, since changing the number of arguments is not really 'the same way'.

src/rules/noAssignmentExpression.ts
```diff
@@ -17,7 +18,21 @@ export class NoAssignmentExpression implements SourceError {
...
+      } else {
+        return theNode
+      }
+    }
+
+    const lastNode = lastAssignmentNode(this.node)
+    const leftStr = generate(this.node.left)
+    const rightStr = generate(lastNode.right)
+
+    const elabStr = `Try moving this to another line:\n\n\t${leftStr} = ${rightStr};`
```
openorclose: Just thought about something. 

Would it be feasible to check if this assignment is in an IfStatement, WhileStatement, ForStatement, LogicalExpression, or ConditionalExpression test condition?

Then we could say something like the multiplication thing, 

If you were planning to check if left is equal to right, use === instead.

src/interpreter-errors.ts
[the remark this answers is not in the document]
LeowWB: The explicit check for false is to weed out cases where writable isn't passed as an argument. But fair enough, it doesn't hurt to have an explicit check for true.

src/rules/noAssignmentExpression.ts
LeowWB: I don't think that's relevant? Correct me if I'm wrong, but this rule is only broken by code that looks like 
`let a = 1; 
let b = a = 2;`
Whereas your equivalence checking is more likely to occur from statements like 
`c = 5`

openorclose: I mean it's also broken in places like
```js
if (a = b) { //meaning to check if a === b
  dostuff();
} else {}
```

openorclose: In any case, this can be done in another PR.

PR #180, opened 2019-02-24
clarifying display, error, stringify, pairs-as-arrays, renaming to 1920

martin-henz:


doc/source_1.tex
```diff
@@ -2,7 +2,7 @@
 
 \begin{document}
 	%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-	\docheader{2019}{Source}{\S 1}
+	\docheader{1920}{Source}{\S 1}
```
thomastanck: Should we use AY1920 instead? It's a more common usage for showing the year.

martin-henz: Haha, I thought 1920 is kinda cute. The next ones will be 2021, then 2122, and finally people see the pattern...

PR #182, opened 2019-02-26
[NATIVE] Lay groundwork for basic errors

openorclose:
# Features:

- [x] Native!
- [x] Builtins work! 
- [ ] Enforcing of strict types (`'' * ''` still works)
- [x] Error handling (most errors are now 1-1 with interpreter)
- [ ] stringifying functions (shows the transpiled code currently) 

# How to use: 

Change 

```typescript
  scheduler: 'async',
  steps: 1000,
}
```
in `src/index.ts` 


src/index.ts
```diff
@@ -40,14 +44,55 @@ export function runInContext(
...
         return Promise.resolve({
           status: 'finished',
-          value: sandboxedEval(generate(transpile(program, context.contextId)))
+          value: sandboxedEval(transpiled)
         } as Result)
       } catch (error) {
-        context.errors.push(new ExceptionError(error, UNKNOWN_LOCATION))
-        return Promise.resolve({ status: 'error' } as Result)
+        if (error instanceof RuntimeSourceError) {
+          context.errors.push(error)
```
thomastanck: Just checking, do these errors come from stuff like type checks being injected into the code? I'm not sure what triggers this line.

src/index.ts
```diff
@@ -40,14 +44,55 @@ export function runInContext(
...
+        if (
+          match === null ||
+          sourceMapJson === undefined ||
+          lastStatementSourceMapJson === undefined
+        ) {
+          context.errors.push(new ExceptionError(error, UNKNOWN_LOCATION))
+          return resolvedErrorPromise
+        }
+        const line = Number(match![1])
+        const column = Number(match![2])
```
thomastanck: Seems like you're parsing the stacktrace, did you consider using a library? A quick google search yields StackTrace.JS, but I'm not sure if overkill.

src/transpiler.ts
```diff
@@ -249,51 +327,67 @@ function getStatementsToAppend(program: es.Program): es.Statement[] {
 
 function splitLastStatementIntoStorageOfResultAndAccessorPair(
   lastStatement: es.Statement
-): es.Statement[] {
+): [es.Statement, es.Statement, {}?] {
```
thomastanck: Same as above, I didn't notice this before but consider using an object instead. Object literal shorthands and destructuring make it quite easy to return and assign these. Objects are more "type safe" compared to arrays as there are names going along with each field.

src/utils/rttc.ts
```diff
@@ -34,18 +34,13 @@ const typeOf = (v: Value) => {
   }
 }
 
-const isNumber = (v: Value) => typeOf(v) === 'number'
-const isString = (v: Value) => typeOf(v) === 'string'
-const isBool = (v: Value) => typeOf(v) === 'boolean'
-const isObject = (v: Value) => typeOf(v) === 'object'
-const isArray = (v: Value) => typeOf(v) === 'array'
+export const isNumber = (v: Value) => typeOf(v) === 'number'
```
thomastanck: Reason for exporting all of these?

src/index.ts
openorclose: Yes, look at operators.ts, there it throws an InvalidNumberOfArguments RuntimeSourceError.

src/index.ts
```diff
@@ -40,14 +44,55 @@ export function runInContext(
...
     if (theOptions.isNativeRunnable) {
+      let transpiled
+      let sourceMapJson
+      let lastStatementSourceMapJson
       try {
+        const temp = transpile(program, context.contextId)
+        // some issues with formatting and semicolons and tslint so
+        transpiled = temp[0]
+        sourceMapJson = temp[1]
+        lastStatementSourceMapJson = temp[2]
```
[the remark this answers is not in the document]
openorclose: I'll have to work on the destructuring part, ts lint and prettier don't seem to play nice with

```;({a, b, c}) = obj```

But I do agree that returning them as objects is definitely nicer.

src/index.ts
openorclose: It's just parsing a single line, nothing too advanced yet.

src/transpiler.ts
openorclose: Agree.

src/utils/rttc.ts
openorclose: Thought i was going to use them, but turns out I didn't

PR #185, opened 2019-02-26
[NATIVE #4 and hopefully final] Strict types

openorclose:
# MOST STUFF WORKING

# Features:

- [x] Native!
- [x] Builtins work! 
- [x] Enforcing of strict types ("" * "" gives error)
- [x] Error handling (most are covered)
- [x] stringifying functions (should be same as interpreter now)
- [ ] strict types for array access and object access (array[can only be number here])
- [ ] BONUS: detecting and stopping of infinite loops
# How to use: 

Change 

```typescript
  scheduler: 'async',
  steps: 1000,
}
```
in `src/index.ts` 


martin-henz:
I guess you need to merge the conflicts, now.

PR #186, opened 2019-02-28
[NATIVE #5] Infinite loop protection

openorclose:
# Infinite loop protection:

Native eval doesn't have the luxury of allowing us to pause it like the interpreter can, so I've hacked it a bit.

I added timeouts for while loops, for loops, and tail recursive functions. Non-tail recursive functions will cause a stackoverflow so it's ok. They use a global timeout of 1s, at which point this error happens:

```
Potential infinite loop detected.
If you are certain your code is correct, press run again without editing your code.
The time limit will be increased from ${from} to ${to} seconds.
This page may be unresponsive for up to ${to} seconds if you do so.
```

The 1 second limit is there for the autorun feature, where students may accidentally create infinite loops during the creation of their code, such as

```js

let i = 0;
while (i < 10) {

//haven't incremented i yet, will result in browser hang with autorun on.
}
```

We need to test this a bit with the autorun feature to see if 100ms is a better starting timeout, because the browser hanging for 1s each time when editing a loop is not very user-friendly.

100ms should be enough for most non-intensive runes and sounds mission. 

Each time the same code is run again, the time limit will increase by a factor of 10, going from 1s to 10s to 100s. The 10s limit should be enough for most sound/rune missions, so the extra clicking of the run button will hopefully be offset by the safety of the browser tab not hanging for too long.

# Features:

- [x] Native!
- [x] Builtins work! 
- [x] Enforcing of strict types ("" * "" gives error)
- [x] Error handling (most are covered)
- [x] stringifying functions (should be same as interpreter now)
- [ ] strict types for array access and object access (array[can only be number here])
- [x] BONUS: detecting and stopping of infinite loops
# How to use: 

Change 

```typescript
  scheduler: 'async',
  steps: 1000,
}
```
in `src/index.ts` 


martin-henz:
I like this pragmatic approach for infinite loop protection.

A suggestion: After the first (or maybe the second?) warning, we could tell the programmer that the browser may hang, and that she should save her changes, or open the same browser again, to retrieve the latest version of the program. (maybe only in "verbose" mode?)

openorclose:
I think that in this case it's better to have this lengthy error message even in non-verbose mode, since hangs do affect user experience. 

(I tried the autorun feature and typed in an infinite loop, `while (true) { }` with native enabled. My browser hung and even refreshing didn't work, probably because the autorun setting was enabled and my code was saved so it kept on trying to run it. I had to delete my cookies to even reopen the playground.)

I don't think saving should be an issue, the frontend should ensure that the code is saved properly first before trying to execute it.

The updated error message is as follows:

```
Line x: Potential infinite loop detected.
If you are certain your code is correct, press run again without editing your code.
The time limit will be increased from ${from} to ${to} seconds.
This page may be unresponsive for up to ${to} seconds if you do so.
```

for loops and

```
Line x: Potential infinite recursion detected: f(1000001) ... f(1000002) ... f(1000003).
If you are certain your code is correct, press run again without editing your code.
The time limit will be increased from ${from} to ${to} seconds.
This page may be unresponsive for up to ${to} seconds if you do so.
```
for functions.

Suggestions on how to make it more concise or making it more clear are welcome.

openorclose:
Potential problem: prompt requires user interaction and will most likely exceed the time limit.

martin-henz:
> Potential problem: prompt requires user interaction and will most likely exceed the time limit.

Nice catch. Any solution?

openorclose:
https://github.com/source-academy/js-slang/blob/0ee282166083c9e4711d72e737dc8cf89ee8802d/src/createContext.ts#L101-L102

Currently alert and prompt are defined as such.

we can just do 
```js
function alert(t) {
  const startTime = runtime();
  oldAlert(t);
  const duration = runtime() - startTime();
  PROPERTIES.maxTimeLimit += duration
}
```
That's the main idea for now.

thomastanck:
For a (longer term) better solution we can probably do a transformation to a form similar to how the interpreter is currently written (wrap most things in function*)

Then we would do simply run it with PreemptiveScheduler as usual.

martin-henz:
You probably haven't noticed, Daryl: The word "code" is a four-letter swear word in CS1101S. In polite company, we use "program" consistently. 

You may want to rewrite this:
If you are certain your code is correct, press run again without editing your code.
to
If you are certain your program is correct, press run again without editing the program.

PR #190, opened 2019-03-02
Nice transpiled output + tests

openorclose:
Makes it easier to parse the transpiled output.


src/__tests__/transpiled-code.ts
```diff
@@ -0,0 +1,152 @@
...
+    // replace native[<number>] as they may be inconsistent
+    const replacedNative = transpiled.replace(/native\[\d+]/g, 'native')
+    // replace the line hiding globals as they may differ between environments
+    const replacedGlobalsLine = replacedNative.replace(/\n\(\(.*\)/, '\n(( <globals redacted> )')
+    // replace declaration of builtins since they're repetitive
+    const replacedBuiltins = replacedGlobalsLine.replace(
+      /\n *const \w+ = native\.builtins\.get\("\w+"\);/g,
+      ''
+    )
+    expect({ code, transpiled: replacedBuiltins }).toMatchSnapshot()
```
thomastanck: Maybe better to return the promise and let the client code do the toMatchSnapshot(). Same amount of typing in both cases

src/__tests__/transpiled-code.ts
```diff
@@ -0,0 +1,168 @@
...
+  const context = mockContext(4)
+  const transpiled = transpile(parse(code, context)!, context.contextId).transpiled
+  // replace native[<number>] as they may be inconsistent
+  const replacedNative = transpiled.replace(/native\[\d+]/g, 'native')
+  // replace the line hiding globals as they may differ between environments
+  const replacedGlobalsLine = replacedNative.replace(/\n\(\(.*\)/, '\n(( <globals redacted> )')
+  expect({ code, transpiled: replacedGlobalsLine }).toMatchSnapshot()
+})
+
+test('Empty code returns empty string', expectTranspiledToMatchSnapshot(''))
```
thomastanck: Then here you can have expectBlah(...).toEqual('')

(I'll pause here for now, will read the rest later)

src/__tests__/transpiled-code.ts
openorclose: What do you mean by this?

`return expect(..)` and then do `expectTranspiled('code').toMatchSnapshot()`?

src/__tests__/transpiled-code.ts
openorclose: won't this be then
```js
test('empty return empty', () => {
  expectBlah('').toBe('');
)
```
?

thomastanck: Yeah, it's been a while since I touched jest :P

src/__tests__/transpiled-code.ts
thomastanck: Yep, exactly

thomastanck: Looking back, I seem to have standardised `test*` which returns a promise that returns objects (which may or may not secretly check things with expect, e.g. status is successful), `snapshot*` which returns a similar promise but snapshots things first, and `expect*` which returns an expect object instantiated with one of the fields (and also may or may not secretly check other things with expect)

src/__tests__/__snapshots__/transpiled-code.ts.snap
```diff
@@ -0,0 +1,514 @@
...
+Object {
+  "code": "function f() {
+  return null;
+}",
+  "transpiled": "const native = $$NATIVE_STORAGE;
+const boolOrErr = native.operators.get(\\"itselfIfBooleanElseError\\");
+const callIfFuncAndRightArgs = native.operators.get(\\"callIfFunctionAndRightArgumentsElseError\\");
+const wrap = native.properTailCalls.wrap;
+(( <globals redacted> ) => {
+  return (() => {
```
thomastanck: Now that I'm actually looking at the transpiled output, what's the purpose of this inner wrap?

src/__tests__/__snapshots__/transpiled-code.ts.snap
```diff
@@ -0,0 +1,514 @@
...
+  "transpiled": "",
+}
+`;
+
+exports[`Ensure return values are of the form {isTail...} after function declarations are transformed 1`] = `
+Object {
+  "code": "function f() {
+  return null;
+}",
+  "transpiled": "const native = $$NATIVE_STORAGE;
```
thomastanck: Actually I'm starting to think that these transpiled snippets are small enough that I can live with adding these to all the native-enabled tests... What do you think?

openorclose: Yes this does sound reasonable.

src/__tests__/__snapshots__/transpiled-code.ts.snap
openorclose:
```js
const native = $$NATIVE_STORAGE;
const boolOrErr = native.operators.get(\\"itselfIfBooleanElseError\\");
const callIfFuncAndRightArgs = native.operators.get(\\"callIfFunctionAndRightArgumentsElseError\\");
const wrap = native.properTailCalls.wrap;
(( <globals redacted> ) => { // (A) list globals as parameters here to set to undefined at A2
  // if there's no inner wrap, 
  // const Number or const window would be illegal,
  // but it doesn't make sense to not allow them as variable names.
  return (() => {
    for (1; 1; 1) {}
    const lastStatementResult = eval(\\"1;\\");
    return lastStatementResult;
  })();
})(); // A2
```

thomastanck: I see, thanks.

PR #192, opened 2019-03-04
Implement partial parsing of program to obtain first line

LeowWB:
Rationale: Syntax error or broken rule occurring anywhere within the program previously meant the entire program could not be parsed. This would mean that we cannot check for the verbose toggle in the first line, so we cannot provide elaborated versions of the syntax error messages, even if the student requested it. This pull request makes it such that the first line of the program will always be parsed regardless of the rest of the program.


src/index.ts
```diff
@@ -51,22 +51,43 @@ export function runInContext(
...
-      }
+  function getFirstLine(theCode: string) {
+    function doNothing() {
+      return undefined
+    }
+
+    let theProgram
+    try {
+      theProgram = parseAt(theCode, 0)
+    } catch (error) {
```
openorclose: Why not get rid of this whole try catch block and not let parseAt throw an error in the first place?

src/parser.ts
```diff
@@ -92,6 +97,21 @@ export class TrailingCommaError implements SourceError {
   }
 }
 
+export function parseAt(source: string, num: number) {
+  let program: es.Program | undefined
+  try {
+    program = (acornParseAt(source, num) as unknown) as es.Program
+    ancestor(program as es.Node, walkers, undefined)
+  } catch (error) {
+    throw error
```
openorclose: Can just return undefined here.

src/index.ts
```diff
@@ -51,22 +51,43 @@ export function runInContext(
...
-      const theDirective = (firstLineOfProgram as Directive).directive
-      if (theDirective !== undefined) {
-        return theDirective
-      }
+  function getFirstLine(theCode: string) {
+    function doNothing() {
+      return undefined
+    }
+
+    let theProgram
```
openorclose: This can be more aptly named `firstExpression` I feel.

src/parser.ts
```diff
@@ -92,6 +97,21 @@ export class TrailingCommaError implements SourceError {
   }
 }
 
+export function parseAt(source: string, num: number) {
+  let program: es.Program | undefined
+  try {
+    program = (acornParseAt(source, num) as unknown) as es.Program
```
openorclose: This comment is directed at me so don't take it personally.

https://github.com/acornjs/acorn/blob/f164d84661f4f79486c84a4c0d1ca2233614c088/acorn/dist/acorn.d.ts#L7

parseExpressionAt and parse returns acorn.Node, not es.Program which is why there's this weird cast to unknown. I understand you just copied it from what I wrote in parse below, so I'll fix that later.

Please check the return types of imported functions next time.

src/parser.ts
```diff
@@ -92,6 +97,21 @@ export class TrailingCommaError implements SourceError {
...
+    ancestor(program as es.Node, walkers, undefined)
```
openorclose: Why does this need to use the ancestor function?

I thought it just needed to get the first string literal.

src/index.ts
```diff
@@ -51,22 +51,43 @@ export function runInContext(
...
+				But the linter is not okay with empty blocks.
+				We don't want to change the linter settings for this one-off situation.
+
+				THerefore this.
+		*/
+      doNothing()
+    }
+
+    if (theProgram) {
+      return ((theProgram as unknown) as Literal).value
```
openorclose: As always, it's much better to do a literal check, because when you do

`if (theProgram !== undefined && theProgram.type === 'Literal') ...`

I think it should be quite clear that theProgram is actually never the type 'Program' in the first place...

src/parser.ts
```diff
@@ -92,6 +97,20 @@ export class TrailingCommaError implements SourceError {
   }
 }
 
+export function parseAt(source: string, num: number) {
+  let theNode: acorn.Node | undefined
+  try {
+    theNode = acornParseAt(source, num)
```
openorclose: Return the node directly here, and the ifstatement wouldn't be needed anymore.

PR #195, opened 2019-03-16
Substitution model

openorclose:
Only evaluates "math" expressions for now, specifically the binary operators. Example:
```js
(1 + 2) * (3 + 4);
3 * 5;

3 * (3 + 4);
3 * 5;

3 * 7;
3 * 5;

21; // plain literals just disappear
3 * 5;

3 * 5;

15;
```


src/substituter.ts
```diff
@@ -0,0 +1,106 @@
...
+        newContext
+      ]
+    }
+  }
+}
+
+function reduce(node: es.Node, context: Context): [es.Node, Context] {
+  const reducer = reducers[node.type]
+  if (reducer === undefined) {
+    return [node, context] // if reducer is not found we just get stuck
```
thomastanck: I think this should throw an error or otherwise signal the end of the reduction.

src/substituter.ts
```diff
@@ -0,0 +1,106 @@
...
+      const [reduced, newContext] = reduce(firstStatement, context)
+      return [
+        ast.blockStatement([reduced as es.Statement, ...otherStatements], node.loc!),
+        newContext
+      ]
+    }
+  }
+}
+
+function reduce(node: es.Node, context: Context): [es.Node, Context] {
```
thomastanck: I think this should be exported so the user has more control over how they want to use the substitution model (eg one step reduction)

src/substituter.ts
```diff
@@ -0,0 +1,106 @@
...
+export function getEvaluationSteps(code: string, context: Context): Array<[es.Node, Context]> {
+  const steps: Array<[es.Node, Context]> = []
+  try {
+    const program = parse(code, context)
+    if (program === undefined) {
+      return [[parse('', context)!, context]]
+    }
+    steps.push([program, context])
+    let [reduced, newContext] = reduce(program, context)
+    while ((reduced as es.Program).body.length > 0) {
```
thomastanck: I think the logic for whether the reduction has ended should be something else... I would have Program/Block reduce to a literal and then for literals to stop the reduction.

Tell me what you think is a better idea.

src/substituter.ts
openorclose: Yeah. I'll throw an irreducible exception

src/substituter.ts
openorclose: Ok.

src/substituter.ts
openorclose: I think I'll make a thunk class that has the property isReducible.

martin-henz: We kind of agreed that the result of an evaluation should be an array of programs (syntax trees). Each element of the array is a step in the small-step reduction. I agree with a "reducible" property, but not with raising an exception in case a program becomes irreducible. The frontend can then decide how to handle. There may be a way to highlight the issue that leads to irreducibility (type error, undeclared name, etc).

openorclose: Errors such as these will be caught (currently string * number already throws an error), so there shouldn't be a case of us not knowing what to do.

openorclose: Also, the current throwing of an Exception of irreducible statements is a temporary measure since not everything is implemented.

martin-henz: Maybe it's good to throw an exception, and handle it, and as usual show an error in the REPL. The SUBST Inspector window can still show the sequence of small-steps.

SwampertX:
Attempted to hook up the substituter to the repl for easier inspection.
Hope to get more feedback on how to improve my poor software engineering decisions.

martin-henz:
This seems an old branch? Are you guys still working on this one? (Last commit 14 days ago)

martin-henz:
The team has answered my question above: They will re-base to master and commit any new changes to this PR.

SwampertX:
Most of the new functions are added in 7882b37.

SwampertX:
One new decision being made here: a program's first evaluation step will always be substituting in predefined functions. As we have not established the distinction in the environment, now the aforementioned step makes no changes to the code evaluated.

SwampertX:
Closing this since the branch has diverged too much from master.

PR #197, opened 2019-03-22
Member check

openorclose:
Check array/ object property getting/setting.


src/transpiler.ts
```diff
@@ -403,6 +405,44 @@ function transformUnaryAndBinaryOperationsToFunctionCalls(program: es.Program) {
...
 }
 
+function transformPropertyAssignment(program: es.Program) {
+  simple(program, {
+    AssignmentExpression(node) {
+      const assignmentExpression = node as es.AssignmentExpression
+      if (assignmentExpression.left.type === 'MemberExpression') {
+        const { object, property, computed, loc } = assignmentExpression.left
+        const { line, column } = loc!.start
+        node = node as es.CallExpression
```
thomastanck: Is it possible to use create.mutateToCallExpression here?

src/transpiler.ts
```diff
@@ -403,6 +405,44 @@ function transformUnaryAndBinaryOperationsToFunctionCalls(program: es.Program) {
...
+  })
+}
+
+function transformPropertyAccess(program: es.Program) {
+  simple(program, {
+    MemberExpression(node) {
+      const memberExpression = node as es.MemberExpression
+      const { object, property, computed, loc } = memberExpression
+      const { line, column } = loc!.start
+      node = node as es.CallExpression
```
thomastanck: Same.

openorclose: Same.

src/transpiler.ts
openorclose: Yes. Seems like I made it because it was horrible to change something to callExpression, but forgot all about it.

PR #219, opened 2019-04-05
INSPECT

podocarp:
Enables the use of `debugger` and provides an entry point for the frontend to use line based breakpoints. The best way to test this out is with the corresponding branch in the frontend [here](https://github.com/source-academy/cadet-frontend/pull/506). Clone them into the same directory and it should be fine.


src/stdlib/inspector.ts
```diff
@@ -0,0 +1,44 @@
...
+let moved: boolean = true
+let prevStoppedLine: number = -1
+
+export const checkEditorBreakpoints = (context: Context, node: Node): void => {
+  if (node.loc) {
+    const currentLine = node.loc.start.line - 1
+    if (!moved && currentLine !== prevStoppedLine) {
+      moved = true
+    }
+    if (context.runtime.debuggerOn && breakpoints[currentLine] === 'ace_breakpoint' && moved) {
```
openorclose: Haven't played with it too much yet to review, but we're trying to have js-slang be independent of cadet-frontend, so I'm not sure about this 'ace_breakpoint'.

podocarp: This is just something carried over from the editor, we could change it to be as long as it is not undefined.

PR #251, opened 2019-06-17
Parser support for no objects mce

martin-henz:


openorclose:
When does a sequence happen?

https://github.com/source-academy/js-slang/blob/5e01265829dbe7f720600f481765e57780508256/src/stdlib/parser.ts#L83-L85

The statements in a function declaration get treated as a "sequence",


https://github.com/source-academy/js-slang/blob/5e01265829dbe7f720600f481765e57780508256/src/stdlib/parser.ts#L292

but the statements in a while/for statement get treated as a "block".

openorclose:
When does a sequence happen?

https://github.com/source-academy/js-slang/blob/5e01265829dbe7f720600f481765e57780508256/src/stdlib/parser.ts#L83-L85

The statements in a function declaration get treated as a "sequence",


https://github.com/source-academy/js-slang/blob/5e01265829dbe7f720600f481765e57780508256/src/stdlib/parser.ts#L292

but the statements in a while/for statement get treated as a "block".

src/stdlib/parser.ts
```diff
@@ -173,58 +176,59 @@ transformers = new Map([
...
+        transform(node.consequent),
+        transform(node.alternate)
+      ])
     }
   ],
 
   [
     'ArrowFunctionExpression',
     (node: es.Node) => {
       node = node as es.ArrowFunctionExpression
```
openorclose: Are block arrow functions allowed in the MCE?

PR #275, opened 2019-07-04
Simple change for inspector to be restricted to chapter 3 and 4 only

podocarp:
https://github.com/source-academy/cadet-frontend/issues/616


martin-henz:
This would have been an elegant solution, but it does not work:

Screenshot 2019-07-05 at 6 17 03 AM

Do your gutter clicks bypass the parser?

(Then: do we actually need to support "debugger;" statements, or is it enough to click the gutter?)

PR #281, opened 2019-07-07
Documentation: deviations from JavaScript, fixes issue 226

martin-henz:
see https://github.com/source-academy/js-slang/pull/new/issue-226

### DIFF
--- a/doc/source_js_differences.tex
+++ b/doc/source_js_differences.tex
@@ -28,21 +28,5 @@ sake of internal consistency and esthetics.
   \end{lstlisting}
   The result of evaluating this program can be \texttt{undefined}
   for implementations of Source. Note that this issue only arises
-  at the toplevel---outside of functions.
-\item[Literal object notation at toplevel:] JavaScript implementations support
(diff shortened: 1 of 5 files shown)


doc/source_js_differences.tex
```diff
@@ -0,0 +1,48 @@
...
+}
+  \end{lstlisting}
+  The result of evaluating this program can be \texttt{undefined}
+  for implementations of Source. Note that this issue only arises
+  at the toplevel---outside of functions.
+\item[Literal object notation at toplevel:] JavaScript implementations support
+  the creation of literal objects using \verb#{...}#. Occasionally,
+  ambiguities with block syntax arises as a result, compounded by
+  optional semicolons in the JavaScript syntax. Apparently,
+  literal objects are preferred in JavaScript in some such cases.
```
openorclose: `{}` resulting in an empty object is just a quirk of chrome's javascript engine, other browsers like firefox and microsoft edge (pre chromium) correctly give `undefined`. So this should not be listed as a difference.

martin-henz:
OK, I've removed the Chrome-specific issue. You may want to take another look.

PR #282, opened 2019-07-08
(RFC) utils/rttc.ts: Add integer check for array indices

### LINKED ISSUE #279: LANG: Non-integer number keys in arrays
are currently supported

Screenshot 2019-07-07 at 6 31 49 PM

but the §3 specs say that keys must be integers.

My suggestion: keep the language specs simple, and allow *any number* to be used as key. 

So I propose that we adapt the specs to the current implementation.

What do you all think?

thomastanck:
One way to resolve #279.


martin-henz:
I find the current error message not so clear:

Screenshot 2019-07-09 at 12 34 38 PM

Current message:

Line 1: Expected integer as prop, got number.

The problem: the number is actually an integer, by any definition of the word.

My suggestion:

Line 1: Expected array index as key, got other number

thomastanck:
Thanks for the review, I've just made the error message clearer:

Examples:

    Expected array index as prop, got other number.
    Expected array index as prop, got string.
    ...

martin-henz:
Travis chokes on the recent change.

Issue #305, opened 2019-07-16
RFC: Improve `parse` readability


thomastanck:
When parsing this code,

```
function test(x, y) { return x; }const x = 1;let y = "test";const z = x + y * (x / y) - f(x, y);
```

![maim9956](https://user-images.githubusercontent.com/829070/61290977-be236300-a7ff-11e9-80bf-d9ada1eb1254.png)

While the `loc` parameter was ok when using objects to represent the AST because they can be easily ignored, when there are only lists it becomes very hard to see what is going on.

Also, there are many areas where there are unnecessary lists in lists. For example, sequences are lists of 2 elements, where the 2nd element is a list. Why not simply have a list where the first element is `sequence` and the remaining elements are the statements in the sequence? In general, if there is an n-ary AST node, we can represent this simply as a list of length n+1.

Here is what it looks like without `loc` information, and with "redundant" list nestings removed. Notice how the adaptive indentation performs better, grouping statements and subexpressions into their own line, and indentation naturally expressing the current level of nesting (not increasing by more than 1 at once). Thoughts?

```
> display(parse('function test(x, y) { return x; }const x = 1;let y = "test";const z = x + y * (x / y) - f(x, y);'));
[ "sequence",
[ [ "constant_declaration",
  [ ["name", ["test", null]],
  [ [ "function_definition",
    [ [["name", ["x", null]], [["name", ["y", null]], null]],
    [["return_statement", [["name", ["x", null]], null]], null] ] ],
  null ] ] ],
[ ["constant_declaration", [["name", ["x", null]], [1, null]]],
[ ["variable_declaration", [["name", ["y", null]], ["test", null]]],
[ [ "constant_declaration",
  [ ["name", ["z", null]],
  [ [ "application",
    [ ["name", ["-", null]],
    [ [ "application",
      [ ["name", ["+", null]],
      [ ["name", ["x", null]],
      [ [ "application",
        [ ["name", ["*", null]],
        [ ["name", ["y", null]],
        [ [ "application",
          [["name", ["/", null]], [["name", ["x", null]], [["name", ["y", null]], null]]] ],
        null ] ] ] ],
      null ] ] ] ],
    [ [ "application",
      [["name", ["f", null]], [["name", ["x", null]], ["name", ["y", null]]]] ],
    null ] ] ] ],
  null ] ] ],
null ] ] ] ] ]
```

martin-henz:
[...]

PR #307, opened 2019-07-16
RFC: Improve readability of `parse`

thomastanck:
Implementation of #305


martin-henz:
I think I followed the original Scheme version quite closely. We need to see how textbook section 4.1.2 reads, with these changes:

https://sicp.comp.nus.edu.sg/chapters/75

thomastanck:
Better to follow SICP closely for historical and clarity reasons, so closing this proposal.

PR #308, opened 2019-07-17
Implement prelude

openorclose:
Some built in functions, like list functions and stream functions can be implemented in Source.

Let's have a prelude that allows us to evaluate these in Source rather than native Javascript, so as to maintain the single source of truth for list and stream libraries.

Current problems:

- [ ] no tests for stream libraries yet (will be added soon)
- [ ] image Since `length` is now in Source, it reports that an error has occurred on line 24. This is correct as indeed, `length` calls `tail` on line 24 of the prelude. However, this is a useless error message to anyone encountering it.

The above problem already exists to a smaller extent as seen below: 

image

Source reports the error being on Line 2, but it is not the Line 2 of the current program, but the Line 2 of the program that `f` had been declared in.

I'm not sure how to solve this issue.


martin-henz:
> Some built in functions, like list functions and stream functions can be implemented in Source.
> 
> Let's have a prelude that allows us to evaluate these in Source rather than native Javascript, so as to maintain the single source of truth for list and stream libraries.
> 
> Current problems:
> 
> * [ ]  no tests for stream libraries yet (will be added soon)
> * [ ]  image Since `length` is now in Source, it reports that an error has occurred on line 24. This is correct as indeed, `length` calls `tail` on line 24 of the prelude. However, this is a useless error message to anyone encountering it.
> 
> The above problem already exists to a smaller extent as seen below:
> 
> image
> 
> Source reports the error being on Line 2, but it is not the Line 2 of the current program, but the Line 2 of the program that `f` had been declared in.
> 
> I'm not sure how to solve this issue.

Some thoughts on this; not sure what to do:

Even for the current way to load predeclared functions, we have bad error messages:
Line -1: Error: head(xs) expects a pair as argument xs, but encountered 1

In the current design (no prelude), we could change the error message a bit, maybe:
Error in a predeclared function: head(xs) expects a pair as argument xs, but encountered 1

If we go for a prelude, error messages could look like this:

Line 42, [Prelude](https://sicp.comp.nus.edu.sg/source/prelude_source_1): Expected number on right hand side...

Line 42: Expected number on right hand side.... (this error message would be the same as what we currently have: The editor is the default place to look for line numbers)

I think to simplify the discussion, we ignore the minor issue: line numbers in REPL might be confusing. This issue is affected by the decision at hand. Maybe file an Issue on this, for now.

PR #310, opened 2019-07-25
Disallow var statements

openorclose:
Let's disallow var statements and allow only const and let.

### DIFF
--- a/src/rules/noVar.ts
+++ b/src/rules/noVar.ts
@@ -1,0 +1,42 @@
+import { generate } from 'astring'
+import * as es from 'estree'
+
+import { ErrorSeverity, ErrorType, Rule, SourceError } from '../types'
+
+export class NoVarError implements SourceError {
+  public type = ErrorType.SYNTAX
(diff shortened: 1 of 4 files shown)


src/rules/noVar.ts
```diff
@@ -0,0 +1,42 @@
...
+    return `Use keyword "let" instead, to declare a variable:\n\n\tlet ${name} = ${value};`
+  }
+}
+
+const noVar: Rule<es.VariableDeclaration> = {
+  name: 'no-var',
+
+  checkers: {
+    VariableDeclaration(node: es.VariableDeclaration, ancestors: [es.Node]) {
+      if (node.kind === 'var') {
```
thomastanck: I think it might be better to have a rule to allow const and let only rather than to disallow var, and we can merge with noDeclareMutable

thomastanck: Nevermind, I take this back. Not currently sane to have a whitelist using rules.

openorclose:
Yeah. I wanted to combine the two also but it's not possible for now.

PR #320, opened 2019-08-02
SUBST

SwampertX:


martin-henz:
Any instructions how I can use this branch? How can I play with it? What's the language currently supported?

SwampertX:
> Any instructions how I can use this branch? How can I play with it? What's the language currently supported?

Still **not** a working prototype. The quickest way is to 
```bash
yarn build
node dist/repl/repl.js 1 subst
```
to run the REPL.

SwampertX:
Last milestone left - make list prelude work.

openorclose:
After much bug fixing, many edge cases have been taken care of. There's probably a few left, such as:
```function id(f) { return f; } id(id) === id(id);``` should be true, but returns false.

Nonetheless, it is a very useful tool, and these bugs can be sorted out along the way.

PR #321, opened 2019-08-04
Fix let (requires prelude to be merged first) (BREAKING CHANGE)

openorclose:
image

## Problem
Program:
```js
let a = 1;
function f() {
  return a;
}
```
roughly becomes:
```js
let a = 1;
function f() {
  return a;
}
variables.a = {value: a, kind: 'let'};
variables.f = {value: f, kind: 'const'};
```
------------
In the same context, if you do:
```js
a = 2;
f();
```
it gets roughly transformed into 
```js
{
// put back the old values
let a = variables.a.value;
const f = variables.f.value;
{// new scope
  a = 2;
  f();
}
variables.a.value = a;
}
```

**Problem** the `a` referenced in f is still unchanged!

## Fix
Program:
```js
let a = 1;
function f() {
  return a;
}
```
roughly becomes:
```js
let a = 1;
function f() {
  return a;
}
variables.a = {value: a, kind: 'let', assignNewValue: function (newValue) {
  return a = this.value = newValue;
}
};
variables.f = {value: f, kind: 'const'};
```
------
In the same context, if you do:
```js
a = 2;
f();
```
it gets roughly transformed into 
```js
{
// put back the old values
let a = variables.a.value;
const f = variables.f.value;
{// new scope
  a = variables.a.assignNewValue(2);
  f();
}
}
```
so that the actual `a` referenced by `f` does indeed get updated correctly.
:tada:


openorclose:
Note that this includes breaking changes:
```js
const f = () => x;
```
is now now longer a valid program, since `x` is undefined. Previously this was valid, and only threw an error when `f` was called.

martin-henz:
> Note that this includes breaking changes:
> 
> ```js
> const f = () => x;
> ```
> 
> is now now longer a valid program, since `x` is undefined. Previously this was valid, and only threw an error when `f` was called.

Will take note of this for the staff briefing next week.

PR #332, opened 2019-08-14
Change variable storage to a function that returns that variable

### LINKED ISSUE #331: Bug with let in NATIVE
image

Let variables that are mutated via a function in the same scope do not change the variable itself.

openorclose:
Fixes #331 
```js
// first eval:
let counter = 0;
function f() {
  counter = counter + 1;
}
becomes
let counter = 0;
function f() {
  counter = counter + 1;
}
variables.set("counter", getValue: ()=>counter, assignNewValue: newValue => counter = newValue);
```
```js
// second eval (wrong):
f(); counter;
becomes
let counter = variables.get("counter").getValue(); // set back counter;
{
  f(); // this has a side effect of incrementing counter
  return counter; // that does not get propagated here
}
```

Let's change **ALL** identifiers that refer to previous scopes also do a retrieval again to make sure the value is the most updated one.
```js
// second eval (fixed):
f(); counter;
//becomes
let counter = variables.get("counter").getValue(); // set back counter;
{
  f(); // this has a side effect of incrementing counter
  return (counter = variables.get("counter").getValue()); // we have no choice but to
   // always assume that counter is updated through a side effect and always retrieve its
   // real value
}
```


openorclose:
> Frankly, my only comment would be to rename `previousVariablesToTimesGoneUp` since what it stored has changed.

Yup, I'll change this! Thanks for the review

PR #346, opened 2019-09-01
Add argument check for builtins

openorclose:
We check arguments for Source functions, but do not for functions implemented in Javascript.

If needed, they have to check the arguments themselves, such as what `parse_int` does.

Changes
-------

Let's make (almost) all functions check for correct number of arguments too.

This would improve error messages received by students when they pass too few arguments to rune libraries, for example.

### Notable exceptions
- list
- stream
- display
- error
- stringify
- raw_display
- math_min
- math_max

These all allow a variable number of arguments.

I did a skim through all of the frontend libraries to make sure none of the functions there use varargs, but feel free to help to check again.

Future
------

Passing in wrong things to functions still have weird error messages (all the `isPrimary` error stuff on piazza)

The way to solve it now would be to manually do type checking for all arguments passed in to each builtin function.

Maybe in the future we can add some simple type checking to Source.

### DIFF
--- a/src/createContext.ts
+++ b/src/createContext.ts
@@ -117,12 +117,12 @@ const defineSymbol = (context: Context, name: string, value: Value) => {
 // If the builtin is a function, wrap it such that its toString hides the implementation
 export const defineBuiltin = (context: Context, name: string, value: Value) => {
   if (typeof value === 'function') {
-    const wrapped = (...args: any) => value(...args)
     const funName = name.split('(')[0].trim()
     const repr = `function ${name} {\n\t[implementation hidden]\n}`
-    wrapped.toString = () => repr
+    value.toString = () => repr
+    value.hasVarArgs = name.includes('...') || name.includes('=')
 
-    defineSymbol(context, funName, wrapped)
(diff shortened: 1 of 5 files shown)


thomastanck:
@openorclose Can you check if you do argument checks for the prelude? Production currently evaluates `pair(1);` to `[1, undefined]`.

martin-henz:
Regarding "Notable Exceptions": Please include the function error here as well. @openorclose is it handled?

martin-henz:
There are conflicts. @openorclose can you take a look?

martin-henz:
@openorclose there is a merge conflict. Can you take a look so that we don't miss anything?

martin-henz:
Looks like this PR is stale. Should we close it?

thomastanck:
@martin-henz Yes, error is part of the exceptions as it allows varargs: `error(1);` and `error(1, 'prepended string')` are both valid. I've edited the PR description to mention this.

The PR should be up to date with the current master as well, so if there are no other problems it should be ok to merge.

PR #355, opened 2019-09-30
Name anonymous functions with more details

podocarp:
Just a small change so the inspector can identify them.

### DIFF
--- a/src/closure.ts
+++ b/src/closure.ts
@@ -86,9 +86,10 @@ export default class Closure extends Callable {
       this.functionName = this.node.id.name
     } else {
       this.functionName =
-        `(` +
+        (this.node.params.length === 1 ? '' : '(') +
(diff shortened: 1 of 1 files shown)


src/closure.ts
```diff
@@ -88,7 +88,10 @@ export default class Closure extends Callable {
     if (this.node.type === 'FunctionDeclaration' && this.node.id !== null) {
       this.functionName = this.node.id.name
     } else {
-      this.functionName = `Anonymous${++Closure.lambdaCtr}`
+      this.functionName =
+        `Anonymous${++Closure.lambdaCtr} (` +
+        this.node.params.map(o => Object(o).name).join(', ') +
```
openorclose:
```suggestion
        this.node.params.map((o: es.Identifier) => o.name).join(', ') +
```

Let's try to use types as much as possible

PR #362, opened 2019-10-18
Improve error messages, fix naming bug and assignment bug in native

openorclose:
Fixes https://github.com/source-academy/cadet-frontend/issues/963

When js functions throw their own errors (e.g. head(xs) will throw if xs is not a pair), it gets wrapped in an ExceptionError. If these functions are nested down, they each get wrapped in their own ExceptionError, causing the [object Object] error message.

Now, we check whether the error has already been wrapped before wrapping the error.

This also improves many of the Line -1 messages that occur.

This also fixes the bug here: https://sourceacademy.nus.edu.sg/playground#chap=3&exec=1000&ext=NONE&prgrm=GYVwdgxgLglg9mABAdwE4EMAOAKAlIgbwChFTEiBfIA

and the bug here:

https://sourceacademy.nus.edu.sg/playground#chap=3&exec=1000&ext=NONE&prgrm=DYUwLgBAhhC8EAcoEsBOBuAUDeS3qA

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-slang",
-  "version": "0.3.3",
+  "version": "0.3.4",
   "description": "Javascript-based interpreter for slang, written in Typescript",
   "author": {
     "name": "Source Academy",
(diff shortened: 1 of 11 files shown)


chrisgzf:
can this be merged soon? many students are facing the `[object Object]` errors on piazza

PR #396, opened 2020-02-06
Allow source file to be passed to repl

openorclose:
Lets you code long stuff in an actual file (like an MCE).

Emulates the Source Academy playground,

where you press run on the left (the file you pass in to the repl)

and eval on the right (the node repl itself).


thomastanck:
LGTM

anubh-v:
Thanks for this @openorclose, I would really appreciate this feature

PR #398, opened 2020-02-20
Add Source Compiler CI check

J0:
**Source -> Wasm Compiler**

This pull request adds a Github action to test if our Source->Wasm compiler can be integrated into Source-jslang.


martin-henz:
Do I understand the logic right: When js-slang changes, this action would check if the Source -> Wasm Compiler breaks. If yes, we would not allow a push to the master branch of js-slang.

The problem that I see with this is that if Source -> Wasm Compiler is not maintained well, this might become a vulnerability for js-slang. Any views: @openorclose , @thomastanck

martin-henz:
> Do I understand the logic right: When js-slang changes, this action would check if the Source -> Wasm Compiler breaks. If yes, we would not allow a push to the master branch of js-slang.
> 
> The problem that I see with this is that if Source -> Wasm Compiler is not maintained well, this might become a vulnerability for js-slang. Any views: @openorclose , @thomastanck

bump

Any views?

J0:
> Do I understand the logic right: When js-slang changes, this action would check if the Source -> Wasm Compiler breaks. If yes, we would not allow a push to the master branch of js-slang.

When js-slang changes it will check if the Source->Wasm compiler in it's current state can be integrated with js-slang. If the integration of Source->wasm to the current repository fails then the push will be blocked.

We can leave this as a non-blocking check(Users can still merge if the build does not pass)

J0:
Closed as per our telegram discussion

PR #427, opened 2020-03-03
tidy up code for source modes

BobbyZhouZijian:
Extracted out repeated codes and modularise them into a separate file named source-rules.

PR #428, opened 2020-03-03
tidy up codes for source mode

BobbyZhouZijian:
Extract repeated codes from source1 to source4 to source-rules.ts


BobbyZhouZijian:
This is a follow up to #427

PR #450, opened 2020-03-18
Source §1 type inference specs

martin-henz:


muruges95:
Made edits to the document based on our group's experience with trying to implement the spec. Most of the changes made were to the rules governing how we add a new constraint to the list of constraints as everything else is pretty straightforward. The idea is that we want to maintain the list of constraints as a DAG, where we get more specific as we move from left to right. Do take a look and let me know how it looks! 

In addition I also added a section at the top that mentions how else if statements are folded into nested if else statements, though I am unsure if this is unnecessary. Besides that, I just added a couple of formatting changes.

PR #452, opened 2020-03-20
Concurrent VM Specification

zhengqunkoo:
The specification is a `tex` file. A `LaTeX` compiler like `pdflatex` is needed to generate the `pdf` document.


zhengqunkoo:
Opened pull request from wrong branch.

PR #456, opened 2020-03-22
Add Source 3 Concurrent

jonchan51:
@zhengqunkoo 

Introduces Source 3 Concurrent into js-slang (internally represented as source chapter 3.4)

Source 3 Concurrent is implemented using two main parts, the SVML compiler, which compiles Source programs into SVML Json Assembly format, and the SVML machine, which runs SVML Json Assembly formatted programs. Refer to https://github.com/source-academy/js-slang/wiki/SVML-Specification for more information.

Source 3 Concurrent introduces three constructs for concurrency. `concurrent_execute`, `test_and_set` and `clear`. Source 3 Concurrent programs are all concurrent programs, and always return undefined. `display` must be used instead to get feedback from the program. 

Outline of changes:
- Changed all `parseInt(sourceChapter, 10)` to `parseFloat(sourceChapter)` (unsure if this is a good idea, should we just represent 3.4 as 34 so that it remains as an integer? Functionality wise, aside from `parseFloat` being able to parse floats as well, they appear to be the same in base 10)
- Changed jsdoc script to support folder names such as 'source_3_concurrent'
- CLI frontend for compiler (Hao Wei)
- Introuced Source 3 Concurrent
  - Generated library documentation
  - Added support for Source 3 Concurrent primitives to AceEditor
  - Added support for Source 3 Concurrent to `runInContext` via `sourceChapter = 3.4` (and hence accessible from cadet-frontend)
  - Tests

TODO:
- Remove debug functions (initially introduced to display current state of machine, with its registers / heap) - Don't think we should remove this so that it is easier for future maintainers to debug
- Documentation on Source 3.4 specs (#452 )
- Dev documentation on compiler and machine (on wiki?)
- Make Source 3.4 program execution stoppable like the interpreter (stretch)


zhengqunkoo:
Let's squash some commits?

PR #457, opened 2020-03-23
Add utility functions in test code

anubh-v:
Added some utility functions to reduce the amount of code needed to write tests


anubh-v:
S

PR #460, opened 2020-03-24
[CS4215] T2 Scope Tools for SA

kush1509:
## Context
The PR adds scope-directed navigation functionality. Using `acorn-walk`, the AST generated by `acorn.parse()` is traversed to find:
1. The node at a particular location (occurence)
2. The declaration, if any, of the above node.

## Implementation
The `findDeclaration()` function is exposed, which takes in a location (this is where the user clicked/requested for the declaration) and returns the declaration of the node at that location. This is done in two steps:
- First, we look for the node at the given location. This node should be an identifier for the process to continue.
- Then, we get all the ancestors of this node. Once we have the ancestors, we recursively search them for a declaration matching the identifier, starting from the innermost to outermost in terms of scope.

## Other Information

Here is the PR that adds this functionality to the frontend: https://github.com/source-academy/cadet-frontend/pull/1028


src/finder.ts
```diff
@@ -0,0 +1,128 @@
...
+      if (identifier.name === node.name && identifier.loc === node.loc) {
+        foundAncestors = Object.assign([], ancestors).reverse()
+        foundAncestors.shift() // Remove the identifier node
+      }
+    },
+    /* We need a separate visitor for VariablePattern because
+    acorn walk ignores Identifers on the left side of expressions.
+    Here is a github issue in acorn-walk related to this:
+    https://github.com/acornjs/acorn/issues/686
+    */
```
martin-henz: nice catch!

PR #461, opened 2020-03-25
Added Support for Lazy Evaluation to Interpreter and Transpiler

jansky:
This pull request adds support for an optional lazy evaluation mode for the interpreter and transpiler. An additional `evaluationMethod` property is added to the execution context to allow for selecting the `lazy` or `strict` evaluation method. The command line argument structure for the REPL has also been overhauled to make it easier to select the chapter, execution method (i.e., native or interpreted), and evaluation method.

## Implementation

In the interpreter, we create a new `Thunk` class that allows us to delay the evaluation of arguments at function application time. We also create the functions `delayIt` to construct thunks and `forceIt` to force the evaluation of thunks (using memoization) as needed. User-defined functions are non-strict, meaning arguments are thunked at application time. Functions in the standard library and binary and unary operators are strict, which means they force their arguments when applied.

In the transpiler, we modify the `transformCallExpressionsToCheckIfFunction` transformation to also thunk the arguments supplied to the wrapped function call. Thunks are constructed as JavaScript objects with the computation suspended in an arrow function taking no arguments, and also allow for memoization. At runtime, a new `forceIt` operator allows us to force the evaluation of thunks where necessary.

Added tests confirm that programs making use of lazy evaluation produce the same result when run with the interpreter and transpiler.


openorclose:
Looks okay to me, except the display of these functions need some fixing.

`const f = x => x; f;` looks like this:
image

Also only function arguments can be lazy? Variable declarations still look strict to me, are they supposed to be lazy?

jansky:
> Looks okay to me, except the display of these functions need some fixing.

It looks like this bug predates my changes, but was missed because it's only exposed when running the REPL with the interpreter (in `master` and `cs4215_lazy_evaluation` the REPL uses native mode by default, whereas now the interpreter is used by default). I've gone ahead and rectified this with a small change to the REPL.

> Also only function arguments can be lazy? Variable declarations still look strict to me, are they supposed to be lazy?

I am following the semantics for lazy evaluation given in SICPJS §4.2. In these semantics only the application of user-defined functions is made non-strict - variables, binary operators, and built-in functions remain strict. Now, in Haskell all of these things are in fact non-strict, so we could take the same approach for Lazy Source. I don't have a strong *personal* preference for one approach over the other, but I think the overall goal of the project is to create a lazy language in alignment with the description in SICPJS. Perhaps @martin-henz can comment on this?

martin-henz:
> > Looks okay to me, except the display of these functions need some fixing.
> 
> It looks like this bug predates my changes, but was missed because it's only exposed when running the REPL with the interpreter (in `master` and `cs4215_lazy_evaluation` the REPL uses native mode by default, whereas now the interpreter is used by default). I've gone ahead and rectified this with a small change to the REPL.
> 
> > Also only function arguments can be lazy? Variable declarations still look strict to me, are they supposed to be lazy?
> 
> I am following the semantics for lazy evaluation given in SICPJS §4.2. In these semantics only the application of user-defined functions is made non-strict - variables, binary operators, and built-in functions remain strict. Now, in Haskell all of these things are in fact non-strict, so we could take the same approach for Lazy Source. I don't have a strong _personal_ preference for one approach over the other, but I think the overall goal of the project is to create a lazy language in alignment with the description in SICPJS. Perhaps @martin-henz can comment on this?

Yes, let's stick to the SICP approach on this. Then we can use the evaluator (or transpiler) in the textbook.

martin-henz:
Can you target the master branch with this PR?

PR #464, opened 2020-03-26
[CS4215] Type inference

openorclose:
View the tests and the corresponding inference at: https://github.com/source-academy/js-slang/blob/5f0dc338d255549306f3c92ec461d3be739a89fe/src/inferencer/__tests__/inferencer.ts

`list` function doesn't work yet ( will probably add a special case for it soon, as of now achieve same functionality with `pair(a, pair(b, null))`.

`pair`, `head`, and `tail` all work with normal pairs and lists!

This currently lives as a separate tool under `src/inferencer`, because the list prelude has some functions that can't be inferred correctly (like `is_list` and `equal`).

Example of type errors:

Screenshot from 2020-03-25 11-53-57

# how lists and pairs work

Here's my proposal for lists and pairs:

let's pretend lists don't exist for the moment.

then pairs would have the polymorphic type `Pair<H, T>` with

```
pair: forall((T1, T2) -> Pair<T1, T2>)
head: forall( Pair<T1, T2> -> T1)
tail: forall( Pair<T1, T2> -> T2)
```

where T1 and T2 represent the head type and tail type.

so functions like below will have corresponding types
```
fourth: (Pair<T106, Pair<T109, Pair<T112, Pair<T115, T116>>>> -> T115);
function fourth(tuple) {
   return head(tail(tail(tail(tuple))));
}
```
Now we introduce the list type `List<E>`, where E is the element type.

So

```null: List<E>```

But we want tail and head and pair to work on these! So I propose that we temporarily represent lists with `Pair<E, List<E>>`, so that 

```
head: Pair<E, List<E>> -> E
tail: Pair<E, List<E>> -> List<E>
pair: (E, Pair<E, List<E>>) -> Pair<E, Pair<E, List<E>>>
```

Note that these still have the same types as what we've defined above, so no 'overloading' is needed.

Then, after each type rule, we have a _normalising_ step, where we transform the type inferred at that particular step. So for all typing rules

```
conditions...
___________
<program fragment>: t
```

becomes
```
conditions...
___________
<program fragment>: t'
where
t' = Pair<e, List<e>> if t = List<e>
t' = Pair<e, List<e>> if t = Pair<e, Pair<e, List<e>>>
t' = t otherwise
```
This is to ensure that all list types remain consistent.

Only one further modification is needed, which is the unification of `T1` with `Pair<T2, T1>`.

Normally, this results in recursive unification which throws an error in all other cases, but in this case we allow it, and say that `T1` can be unified with `Pair<T2, T1>` with `T1` = `List<T2>`.

And just like how we replace all free `Number` types with `Integer` at the end of the inference process, we will go in and replace all `Pair<E, List<E>>` types with `List<E>`.

### CODE REVIEW

File: src/inferencer/inferencer.ts [hunk 1]
```diff
@@ -0,0 +1,394 @@
...
+      } else if (b instanceof t.Var && b.canContain(a)) {
+        b.instance = a
+      } else {
+        throw new Error('Unable to unify ' + a + ' with ' + b)
+      }
+    }
+  } else if (!(a instanceof t.Var) && b instanceof t.Var) {
+    unify(b, a)
+  } else if (!(a instanceof t.Var || b instanceof t.Var)) {
+    if (a instanceof t.Polymorphic && b instanceof t.Polymorphic) {
```
muruges95: From my understanding, you should not see any polymorphic types in unify I think. You only get polymorphic types in the type environment. When fetching a polymorphic type, you would replace the bound variables with fresh type variables. Within the rules itself you only make use of monomorphic types and only create a polymorphic type when storing back into the type environment.

src/inferencer/types.ts
```diff
@@ -0,0 +1,300 @@
...
+    super('List', [v])
+  }
+}
+
+export class Pair extends Polymorphic {
+  constructor(head: Type, tail: Type) {
+    super('Pair', [head, tail])
+  }
+}
+
```
muruges95: Wanted to ask why should the function, list and pair types be polymorphic by default? I believe that they should be monomorphic by default and then polymorphism is like some sort of wrapper that wraps around any monomorphic type.

muruges95:
Also the error handling is really good! Something that you did that our group didn't think to do was having try catch blocks around every call to unify in each of the case statements, which allows for more specific messages. Will probably look to adapting some of it for my own group's implementation for helpful error messages!

src/inferencer/util.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  env.set('===', comparableXcomparableRBool())
+  env.set('!==', comparableXcomparableRBool())
+  env.set('<', comparableXcomparableRBool())
+  env.set('<=', comparableXcomparableRBool())
+  env.set('>', comparableXcomparableRBool())
+  env.set('>=', comparableXcomparableRBool())
+
+  let headType = new t.Var()
+  let tailType = new t.Var()
```
raynoldng, line 64: Since you treat all functions are polymorphic, and fresh variables are generated, is there a need for new vars for each pair/list functions?

src/inferencer/util.ts
```diff
@@ -0,0 +1,153 @@
...
+  env.set('pair', new t.Function([headType, tailType], new t.Pair(headType, tailType)))
+  headType = new t.Var()
+  tailType = new t.Var()
+  env.set('tail', new t.Function([new t.Pair(headType, tailType)], tailType))
+  headType = new t.Var()
+  tailType = new t.Var()
+  env.set('head', new t.Function([new t.Pair(headType, tailType)], headType))
+  type = new t.Var()
+  env.set('is_null', new t.Function([new t.Pair(type, new t.List(type))], new t.Bool()))
```
raynoldng, line 73: is_null should `any -> bool` right?

raynoldng: Oh I am guessing this decision is so that you can parse the `listPrelude`?

openorclose: Yup!

src/inferencer/types.ts
openorclose: I guess it should be named Parameterised instead to reduce confusion?

src/inferencer/util.ts
openorclose: Yeah, thanks for catching!

src/inferencer/inferencer.ts
openorclose: Yeah, it should be called Parameterised instead, in that Functions, Lists, and Pair are all parameterised over multiple types.

src/inferencer/__tests__/inferencer.ts
```diff
@@ -0,0 +1,378 @@
...
+filter: (T1 -> boolean, List<T1>) -> List<T1>
+enum_list: (number, number) -> List<number>
+list_ref: (List<T1>, number) -> T1
+accumulate: ((T1, T2) -> T2, T2, List<T1>) -> T2"
+`)
+})
+
+test('Type inference of permutation', async () => {
+  const code = `function accumulate(op, init, xs) {
+    return is_null(xs) ? init : op(head(xs), accumulate(op, init, xs));
```
raynoldng, line 74: Should be this right?
`accumulate(op, init, tail(xs))`

src/inferencer/inferencer.ts
```diff
@@ -0,0 +1,394 @@
...
+        }
+      } else if (a.canContain(b)) {
+        a.instance = b
+      } else if (b instanceof t.Var && b.canContain(a)) {
+        b.instance = a
+      } else {
+        throw new Error('Unable to unify ' + a + ' with ' + b)
+      }
+    }
+  } else if (!(a instanceof t.Var) && b instanceof t.Var) {
```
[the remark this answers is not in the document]
openorclose: Yup, simplified the whole if else part. Thanks!

src/inferencer/__tests__/inferencer.ts
openorclose: Yeah, thanks for noticing!

PR #474, opened 2020-03-29
[CS4215] T2 Lazy Evaluation in Source Academy

alcen:
In this Pull Request, we have implemented a lazy mode of evaluation for js-slang's transpiler and interpreter. This evaluation mode supports functions up to Source 2 (Pairs and Lists). As a result, certain programs are now possible in Source that were not possible before.

For example:

```
const ones = pair(1, ones);
const twos = map(x => x * 2, ones);
head(twos); // returns 2
```

Here is a summary of the changes made, in terms of implementation of the lazy evaluation.

* Lazy evaluation is activated on two new chapter numbers: 2001 for Source 1 (Lazy), and 2002 for Source 2 (Lazy)

* Based on execution method, program execution will be handled either by the transpiler or interpreter

* Implementation details for transpiler and interpreter are slightly different

  * In the transpiler, every possible denotable value is wrapped into a lambda function, which is stored in an object that holds more information such as its type, and whether it is evaluated or not

  * In the interpreter, parts of the syntax tree identified to be function arguments or variable declarations are not evaluated immediately, but instead stored in a "Thunk" object for later use

Team members:
@alcen 
@jomcruz93


martin-henz:
This PR decreases test coverage by a whooping 12%. There are lots of new test cases, so don't know what's going on. Any idea, anyone?

src/interpreter/interpreter.ts
```diff
@@ -326,22 +380,45 @@ export const evaluators: { [nodeType: string]: Evaluator<es.Node> } = {
   UnaryExpression: function*(node: es.UnaryExpression, context: Context) {
```
Ahmedjjj, line 380: Just curious. Which Unary operators are lazily evaluated ?

src/stdlib/interpreterLazyList.ts
```diff
@@ -0,0 +1,177 @@
...
+
+// head returns the first component of the given pair,
+// throws an exception if the argument is not a pair
+// LOW-LEVEL FUNCTION, NOT SOURCE
+export function head(xs: any) {
+  if (isInterpreterThunk(xs)) {
+    const result = force(xs)
+    if (result !== null && is_pair(result)) {
+      // If head is a thunk, force its value.
+      if (result[0] !== null && isInterpreterThunk(result[0])) {
```
Ahmedjjj, line 57: It's also possible to delay the evaluation of the head as much as possible, until it's needed.

src/stdlib/interpreterLazyTypeCheck.ts
```diff
@@ -0,0 +1,192 @@
...
+export function is_thunk(v: Value) {
+  if (isInterpreterThunk(v)) {
+    // Unwrap once to expose the argument.
+    const result = force_once(v)
+    return result !== null ? isInterpreterThunk(result) : false
+  } else {
+    return false
+  }
+}
+
```
Ahmedjjj, line 18: I'm not sure, but can the initial Type Checking methods can be used if you just always force the arguments ?

src/stdlib/interpreterLazyList.ts
martin-henz: I agree.

src/stdlib/interpreterLazyList.ts
```diff
@@ -0,0 +1,177 @@
...
+    } else {
+      throw new Error('head(xs) expects a pair as argument xs, but encountered ' + stringify(xs))
+    }
+  }
+}
+
+// tail returns the second component of the given pair
+// throws an exception if the argument is not a pair
+// LOW-LEVEL FUNCTION, NOT SOURCE
+export function tail(xs: any) {
```
martin-henz, line 79: Same here: tail could be lazy, also.

src/stdlib/interpreterLazyTypeCheck.ts
martin-henz: T2 Lazy Evaluation team: Are you proposing a user-level function `is_thunk`?

src/stdlib/interpreterLazyList.ts
jomcruz93: You are right that it can be delayed. The reason it was made to behave this way is because for this lazy interpreter, function call statements will return the final value. For example, if we have a function `add(x, y)` that returns `x + y`, then the statement `add(1, 2);` will return `3` and not a thunk.

Having said that, there is a way to delay its evaluation by assigning it to a constant.
`const a = pair(1, 2);`
`const b = head(a);` - This will not evaluate the call to head until the value of b is needed.

However,
`head(a);` will force it to evaluate and will return `1`

src/stdlib/interpreterLazyTypeCheck.ts
jomcruz93: To @Ahmedjjj:
In our lazy interpreter, a thunk or value may possibly be 'thunked' multiple times because function arguments are delayed. To better explain, here is an example:
`is_thunk(a);`
Assuming that `a` contains a thunk, then by delaying the arguments of the function, `a` will be 'thunked' once again.
In the body of `is_thunk(v: Value) { .. }`, it will first check if `v` is a thunk (returns `true` since arguments are always delayed except for some built-in functions that remains strict).
Then it will force the value of `v` **once** to expose the actual argument which is the identifier named `a`. Since we know that `a` contains a thunk (whether its already evaluated or not is not considered), then it will return `true`.

To @martin-henz:
It was originally created to help us with minor debugging in the interpreter. I'm not sure how useful it is for the users, but I'm more inclined to the idea that it should be taken down as the concept of thunks need not be transparent to the users.

src/stdlib/interpreterLazyList.ts
martin-henz: ok

src/stdlib/interpreterLazyTypeCheck.ts
martin-henz: ok

src/stdlib/interpreterLazyList.ts
martin-henz: Background info on laziness of pair, head and tail: 
https://sicp.comp.nus.edu.sg/chapters/84#p4 :
Note that these lazy lists are even lazier than the streams of chapter 3: The head of the list, as well as the tail, is delayed.[2] In fact, even accessing the head or tail of a lazy pair need not force the value of a list element. The value will be forced only when it is really needed—e.g., for use as the argument of a primitive, or to be printed as an answer.

PR #478, opened 2020-03-30
Add non-deterministic Source 3

anubh-v:
This PR introduces non-deterministic Source 3 to `js-slang`.

Non deterministic Source 3 is a version of Source 3 with a built-in search mechanism.
Programmers can specify sets of *values*, and *requirements* that the values must satisfy.
The program evaluator then automatically identifies the values that meet the requirements.

We have created a REPL for this language in `src/repl/repl-non-det.ts`.
Here is the workflow for using the REPL:

Note:  **Any Source 3 program that does not use while loops, for loops or arrays is a valid non deterministic Source 3 program.**

1. run `yarn build` (only needed the first time)
2. run `node dist/repl/repl-non-det.ts`
3. type `.editor` to enter "editor" mode in the REPL. This allows input to be entered over multiple lines.
4. enter a program into the REPL. 
5. press CTRL + D
6. The REPL will return the first value that satisfied the requirements in your program.
7. If you want to see the next value that satisfied the requirements given in your program, enter `try again`. Note that `try again` is just a REPL command and not part of the language syntax.
8. If you want to evaluate a new program, repeat from Step 3.

#### How to specify possible values and requirements for those values?
The `amb` operator can be used to specify a set of values.
The expression `amb(e1, e2, e3, ...)` returns one of the values `e1`, `e2`, `e3,` , ...
If the program evaluator fails the requirements given after this expression, it will backtrack and
choose another value from the values `e1`, `e2`, `e3`, ...

Use the `require` operator to specify a requirement.
The operator accepts a single boolean expression. 
Example: `require(f > 3);` represents the requirement that the variable f should be greater than 3.

You may read about non-deterministic programming in SICP [Chapter 4.3](https://sicp.comp.nus.edu.sg/chapters/85)

## Examples
The following example programs can be tried in the REPL:
<br /> In the example below, we list all values of f larger than 3 <br />
```js
> const f = amb(1, 2, 3, 4, 5, 6); require(f > 3); f; 
4
> try again
5
> try again
6
> try again
undefined
```
A "hello world" program:
```js
> amb('hello', 'bye') + ' world!';
'hello world!'
> try again
'bye world!'
> try again
undefined
```
An expression that generates an infinite sequence of integers
```js
> const integer_from = n => amb(n, integer_from(n + 1));
n => amb(n, integer_from(n + 1))
> integer_from(1);
1
> try again
2
> try again
3
> try again
4
> try again
5
> try again
6
```
 
You may find more examples from SICP JS in `src/__tests__/non-det-sicp-examples.ts`

## Implementation details
The interpreter is used for evaluation, in `interpreter-non-det.ts`. A general idea behind the modifications to the original interpreter is to avoid yielding any intermediate values of the program. Consequently, we are able to yield only final values of the program by using a custom scheduler, `NonDetScheduler`.

The following are the specific ideas harnessed in the evaluation of different expression types: <br />
- __variable declaration__: the right hand expression is evaluated. This returns a generator of possible values and we perform the variable declaration for each value.
- __binary / unary expressions__: we first evaluate the argument expressions of the application. This gives us generators for the arguments, and we apply the operator to each possible set of arguments
- __sequences__: the evaluation of each sequence is run as loop which nests the evaluation of each subsequent statement. This allows each possible value of statement 1 to be yielded with each possible value of statement 2 and so on, thereby enabling the combinatorial behaviour of `amb`
- `amb`: each of the provided arguments are evaluated in the original context (via deep cloning and resetting to the clone). This serves as a mechanism to backtrack to the original program state and undo effects such as assignments in a particular _world_
- `require`:  a (arbitrary) value is yielded iff the predicate evaluates to `true`. Therefore, no value being yielded causes subsequent statements to not be evaluated and the parent statement to continue to its next possible value
- `cut`: the `cut` operator functions similarly to that in the Prolog programming language. If a statement in a sequence is a `cut` statement or one of the subsequent statement is, then the statement is not added back to the sequence. Therefore, only the statements after `cut` remain in the sequence and are evaluated again for further values (eg. on invoking _try-again_)


martin-henz:
Overall looks pretty good. 

Test coverage has decreased quite a bit. Can you check:
https://coveralls.io/builds/29700903/source?filename=src%2Finterpreter%2Finterpreter-non-det.ts#L55
Can you add more cases to get the coverage up?

I appreciate that you have a non-det specific REPL. That's the way to go.

anubh-v:
> Can you add more cases to get the coverage up?

Yes, we'll be adding more tests to cover the cases where a program results in a runtime error.

src/createContext.ts
```diff
@@ -205,6 +206,11 @@ function importPrelude(context: Context) {
   if (context.chapter >= 3) {
     prelude += streamPrelude
   }
+
```
martin-henz, line 209: See discussion in group "SA js-slang-related", April 2 2020:

Martin Henz, [2 Apr 2020 at 3:17:09 PM]:
Folks, we need to find a solution for this. I just saw this in the non-det PR:

if (context.chapter === 4.3) {
    prelude += nonDetPrelude
  }
I don't think we should use floating point numbers for this clearly discrete issue.

From the user point of view, I'm suggesting that we keep integer chapter numbers: 1-4, but we add modifiers, perhaps we can call them "variants".

So we would have in the drop-down:

Source §1
Source §2
Source §2 Lazy
Source §3 Non-det
Source §3 Concurrent
Source §4
and eventually

Source §x Arduino
Source §x Wasm

Under the hood, we currently have:
Source §1: native (default)
Source §1: small-step evaluator (when stepper tool is chosen)
Source §2: native (default)
Source §2: small-step evaluator (when stepper tool is chosen)
Source §3: native (default)
Source §3: interpreter (when debugger/env viz are chosen)
Source §4: native (default)
Source §4: interpreter (when debugger/env viz are chosen)

We are proposing to add:
Source §2 Lazy: native
Source §3 Non-det: interpeter
Source §3 Concurrent`: virtual machine

src/types.ts
```diff
@@ -104,9 +105,18 @@ export interface Context<T = any> {
...
+   */
   executionMethod: ExecutionMethod
 
   evaluationMethod: EvaluationMethod
+
+  /**
+   * Describes the strategy / paradigm to be used for evaluation
+   * Examples: lazy, concurrent or non-deterministic
+   */
+  variant?: Variant // This will replace evaluationMethod
```
anubh-v, line 119: I have added a new field `Variant` that captures the evaluation strategy / paradigm.

Because our non-deterministic interpreter needs the value of this field when we create our prelude, I had to modify the `createContext` and `createEmptyContext` methods as well. This meant I also had to modify test code in other js-slang components which were using `createContext` and `createEmptyContext`.

arsalanc-v2, line 119: Would it be better if `variant` were a required property? 
Since:
- there is a default value for it ('default')
- the default value is set when creating the context if no value for `variant` is specified. Therefore, there isn't a case where the context is created without having the property

anubh-v: Yes, I think it can be a required property. I've changed it.

src/stdlib/non-det.prelude.ts
```diff
@@ -0,0 +1,28 @@
+export const nonDetPrelude = `
+    function require(predicate) {
+        return predicate ? "Satisfied require" : amb();
+    }
+
+    /* DISTINCT */
+    /* The distinct function checks whether the items in a list are unique. */
+    /* Taken from SICP JS section 4.3.2 */
+
+    function distinct(items) {
```
martin-henz, line 10: The function distinct is not needed in the prelude of non-det. It can be easily added in each example that needs it.

src/interpreter/interpreter-non-det.ts
```diff
@@ -0,0 +1,502 @@
...
+
+const handleRuntimeError = (context: Context, error: RuntimeSourceError): never => {
+  context.errors.push(error)
+  context.runtime.environments = context.runtime.environments.slice(
+    -context.numberOfOuterEnvironments
+  )
+  throw error
+}
+
+const HOISTED_BUT_NOT_YET_ASSIGNED = Symbol('Used to implement hoisting')
```
martin-henz, line 62: This should be called DECLARED_BUT_NOT_YET_ASSIGNED.
The notion of "hoisting" is not useful when we understand that the scope of any declaration is the block in which it appears.

src/interpreter/interpreter-non-det.ts
```diff
@@ -0,0 +1,502 @@
...
+
+function hoistIdentifier(context: Context, name: string, node: es.Node) {
```
martin-henz, line 64: Should be called declareIdentifier.
Same for
hoistVariableDeclarations
hoistFunctionsAndVariableDeclarationsIdentifiers

src/interpreter/interpreter-non-det.ts
arsalanc-v2: I see, I have updated it.

PR #479, opened 2020-03-30
WIP Concurrent VM

zhengqunkoo:
The specification is a `tex` file. A `LaTeX` compiler like `pdflatex` is needed to generate the `pdf` document.


zhengqunkoo:
> My suggestion: Just refer to the specs in the README.org document
I'm not sure we should leave `README.org` - I believe the barrier of entry to using `emacs` is higher than editing the raw `*.tex` file. There are considerable set-up costs of installing and learning `org-mode`. At the very least, keeping `README.org` around means `org-mode` is yet another layer of abstraction that future editors need to think about.

So I suggest we leave the `*.tex` file for future edits.

martin-henz:
> > My suggestion: Just refer to the specs in the README.org document
> 
> I don't think we should keep `README.org` - I believe the barrier of entry to using `emacs` is higher than editing the raw `*.tex` file. There are considerable set-up costs of installing and learning `org-mode`. At the very least, keeping `README.org` around means `org-mode` is yet another layer of abstraction that future editors need to think about.
> 
> So I suggest we leave the `*.tex` file for future edits.

Good thinking.

zhengqunkoo:
There is now only one canonical specification, as discussed.

zhengqunkoo:
Since I'm using `README.org` to export the same content to multiple formats, future content changes in `*.tex` should also be reflected in `README.org`. My repo's branch structure has changed to accomodate this.

This PR is closed in favor of #482.

PR #481, opened 2020-03-31
Fix navigation issue

kush1509:
In the following code, on pressing Cmd+B at the x in the line `const z = x + 2;`, the cursor was moved to the next line, instead of the x in the first line.

```
{
    const x = 1;
    {
        const z = x + 2;
        const x = 2;
        function f(y) {
            return x + y;
        }
    }
}
```

Solution: Added a check to make sure the found declaration of the occurrence does not occur after the occurrence in the code, since Source does not allow this anyway.


kush1509:
Not needed, the behaviour is correct.

PR #487, opened 2020-04-02
Fix undefined error that causes editor to crash on refactoring

hcwong:
Fixes https://github.com/source-academy/cadet-frontend/issues/1039

Loose parsing gives us a tree which may not exactly fit acorn specifications. 
Therefore, we add an additional null/undefined check in `scopeVariables`

I've tested with a few other non valid programs and it does not seem to crash anymore with this check.


hcwong:
@BobbyZhouZijian for your review thanks

PR #495, opened 2020-04-03
[WIP] Type Inferencing for up to Source 2

raynoldng:
Implementation of the specs provided in the attached file.
[source_1_type_inference (5).pdf](https://github.com/source-academy/js-slang/files/4428510/source_1_type_inference.5.pdf)

## Extending support for Pairs and Lists
largely based on what was proposed in [PR464](https://github.com/source-academy/js-slang/pull/464), especially the normalization step. 

## Bugs/Odd behaviour

An odd side effect of `const`s is that first order functions will not work if used in a monomorphic manner. For example:
```
function accumulate(op, init, xs) {
  return is_null(xs) ? init : op(head(xs), accumulate(op, init, tail(xs)));
}
const xs = pair(1, pair(2, null));
const ys = pair(true, pair(true, null));
accumulate((x,y)=>x+y,0,xs);
accumulate((x,y)=>x||y,0,ys);
```

would be typed as:
```
accumulate: ((T67, T21) -> T21, T21, [T67, List<T67>]) -> T21
xs: [number, List<number>]
ys: [boolean, List<boolean>]
```

On the other hand:
```
function accumulate(op, init, xs) {
  return is_null(xs) ? init : op(head(xs), accumulate(op, init, tail(xs)));
}
const xs = pair(1, pair(2, null));
const ys = pair(true, pair(true, null));
const a = accumulate((x,y)=>x+y,0,xs);
const b = accumulate((x,y)=>x||y,0,ys);
```

Would error as `accumulate` would have the type `accumulate: ((number, number) -> number, number, [number, List<number>]) -> number` and can no longer take in `bool` as input

# Remaining loose ends
- better error messages, given that this implementation operates on constraint lists as opposed to Hindley Milner's node-centric unify, it is hard to reconcile type errors when solving the constraint lists with the source node. Also making use of the notion of untypability as we haven't taken advantage of it yet.


src/typeChecker.ts
```diff
@@ -0,0 +1,925 @@
...
+  program: TypeAnnotatedNode<es.Program>
+): [TypeAnnotatedNode<es.Program>, TypeError[]] {
+  typeIdCounter = 0
+  typeErrors = []
+  const env: Env = initialEnv
+  const constraints: Constraint[] = []
+  try {
+    // dont run type check for predefined functions as they include constructs we can't handle
+    // like lists etc.
+    if (program.body.length < 100) {
```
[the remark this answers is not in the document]
muruges95: ah that was some old code we forgot to remove, I will remove it.

src/typeChecker.ts
```diff
@@ -0,0 +1,925 @@
...
+ */
+export function typeCheck(
+  program: TypeAnnotatedNode<es.Program>
+): [TypeAnnotatedNode<es.Program>, TypeError[]] {
+  typeIdCounter = 0
+  typeErrors = []
+  const env: Env = initialEnv
+  const constraints: Constraint[] = []
+  try {
+    // dont run type check for predefined functions as they include constructs we can't handle
```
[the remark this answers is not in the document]
muruges95: yup same here this was before we handled source 2

src/typeChecker.ts
```diff
@@ -0,0 +1,925 @@
...
+function blockStatementHasReturn(node: es.BlockStatement): boolean {
+  const body = node.body
+  for (const stmt of body) {
+    if (stmt.type === 'ReturnStatement') {
+      return true
+    } else if (stmt.type === 'IfStatement') {
+      if (ifStatementHasReturn(stmt)) {
+        return true
+      }
+    }
```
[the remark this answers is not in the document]
muruges95: ah didn't even know this was legal. Thanks!

src/__tests__/typeChecker.test.ts
```diff
@@ -0,0 +1,311 @@
...
+    // expect(program.body[2].declarations[0].init.inferredType).toEqual<Type>({
+    //   kind: 'primitive',
+    //   name: 'number'
+    // })
+    const [program, errors] = typeCheck(parse(code1, 2))
+    expect(topLevelTypesToString(program)).toMatchInlineSnapshot(`
+"accumulate: ((number, number) -> number, number, [number, List<number>]) -> number
+xs: [number, List<number>]
+y: number
+map: Couldn't infer type
```
openorclose, line 82: is there something wrong here? It says happy path but some types don't get inferred.

src/__tests__/typeChecker.test.ts
```diff
@@ -0,0 +1,311 @@
...
+  it('unhappy paths for recursive functions', () => {
+    const code = `
+      function foo(f) {
+        return foo;
+      }
+    `
+    const [program, errors] = typeCheck(parse(code, 2))
+    expect(topLevelTypesToString(program)).toMatchInlineSnapshot(`"foo: Couldn't infer type"`)
+    expect(parseError(errors)).toMatchInlineSnapshot(`
+"Line 1: Contains cyclic reference to itself, where the type being bound to is a function type
```
openorclose, line 160: Is it possible to only show one error?

src/__tests__/typeChecker.test.ts
```diff
@@ -0,0 +1,311 @@
...
+    `
+    const [program, errors] = typeCheck(parse(code, 2))
+    expect(topLevelTypesToString(program!)).toMatchInlineSnapshot(`"foo: (T3, T4) -> [T3, T4]"`)
+    expect(parseError(errors)).toMatchInlineSnapshot(`""`)
+  })
+
+  it('happy paths for pair functions', () => {
+    const code = `
+const x = pair(3, 4);
+head(x) + 56;
```
openorclose, line 184: should probably move 184 down so that the type of foo can be inferred.

On another note, why does y have a correct type when foo doesn't?

src/__tests__/typeChecker.test.ts
```diff
@@ -0,0 +1,311 @@
...
+      }
+    `
+    const [program, errors] = typeCheck(parse(code1, 2))
+    expect(topLevelTypesToString(program)).toMatchInlineSnapshot(
+      `"append: ([T29, List<T29>], [T29, List<T29>]) -> [T29, List<T29>]"`
+    )
+    expect(parseError(errors)).toMatchInlineSnapshot(`""`)
+  })
+
+  it('unhappy paths for recursive functions', () => {
```
hilltse, line 151: What does it mean to be unhappy?

muruges95:
According to my correspondence with @openorclose , in some of the cases which have been highlighted are the expected behavior as the validity checker is tagging the nodes as untypable. Hence, why some of the top level types are appearing as 'cannot infer type'. If in the future, we relax the validity checker to only make declarations using unassigned names be tagged as 'untypable' these issues will be resolved.

src/__tests__/typeChecker.test.ts
muruges95: It simply refers to recursive functions being defined/used in a way that it will throw a type error when using our checker.

src/__tests__/typeChecker.test.ts
muruges95: I think it has got to do with the validity checker being a bit too restrictive. If it was a type error bug in our checker it would have raised a error, but we didnt raise any such bugs. Reordering the statements seems to make the issue go away so I will just do that.

src/__tests__/typeChecker.test.ts
muruges95: Same as the other issue below. Resolved by reordering the statements.

src/__tests__/typeChecker.test.ts
muruges95: Fixed it by reordering the statements.

src/__tests__/typeChecker.test.ts
martin-henz: unhappy sounds intuitive enough.

package.json
```diff
@@ -99,7 +99,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn tslint --fix && yarn format && yarn tslint",
+      "pre-commit": "yarn format && yarn tslint",
```
openorclose, line 102: Can we revert this change?

muruges95: fixed and removed duplicate entries in the file

muruges95:
> Overall implementation follows the formal specs, seems good to go.
> 
> Just one small change:
> Can we have the type checker in its own folder, just like the other parts of js-slang?

Done. We will be adding more contextual error messages by this weekend, similar to how your PR handles errors in each statement/expression so that it's more useful and add more tests to show them.

PR #526, opened 2020-04-11
add test cases for ace mode

BobbyZhouZijian:
add test cases for:
* function tokens
* operator syntax
* forbidden keywords
* forbidden js reserved words


BobbyZhouZijian:
Seems like I accidentally used npm to install the package instead of yarn. Just fixed it. It should work now.

PR #527, opened 2020-04-11
Improve error handling for Source 3 Concurrent

jonchan51:
- Fixed return values for `display` and `draw_data`
- Added runtime type checks for primitive operations
- Added runtime number of arguments check for function calls
- Fixed broken if else if compilation
- Added more verbose error messages


src/stdlib/vm.prelude.ts
```diff
@@ -377,12 +381,20 @@ function _runtime(x) {}
 
 // 74 unlike Source version, does not fail gracefully
```
martin-henz, line 382: remove this comment also?

src/stdlib/vm.prelude.ts
```diff
@@ -377,12 +381,20 @@ function _runtime(x) {}
...
 function _set_head(xs,x) {
-  xs[0] = x;
+  if (!is_pair(xs)) {
+    error('set_head(xs) expects a pair as argument xs, but encountered ' + stringify(xs));
+  } else {
+    xs[0] = x;
+  }
 }
 
 // 75 unlike Source version, does not fail gracefully
```
martin-henz, line 391: same

src/stdlib/vm.prelude.ts
```diff
@@ -500,7 +519,11 @@ function _stream_to_list(xs) {
 
 // 89 unlike Source version, does not fail gracefully
```
martin-henz, line 520: same

PR #532, opened 2020-04-12
Better Type Inferencing Errors

raynoldng:
3 Main changes
1. better error messages following the format in https://github.com/source-academy/js-slang/pull/495
2. A slight deviation from the specs for function application. Was supposed to add constraints before inferring the argument node types, we swapped the order so that we can infer the node types so that if the parameter types mismatch we can report the given argument types. This was not supposed previously as the added constraints would block the inferring of argument types in the event of a mismatch and we cannot easily recover the given type. (see below on change in inference rule)
3. Added built-in functions that accept variable number of arguments, `error`, `display`, `list`. 
4. For final type output, normalize `[T1, List<T1>] to List<T1>`

## Change in inferring function application

Previously;
Screen Shot 2020-04-14 at 1 06 49 AM
Updated to:
Screen Shot 2020-04-14 at 1 06 45 AM

So the key change here is that we infer the argument types first and add their constraints in before we add function call specific constraints. That was why there is the case of mismatch in argument types can recover what the passed in types where. Whereas previously $\sigma_1$ would prevent us from inferring argument types.


martin-henz:
> 3 Main changes
> 
> 1. better error messages following the format in #495
> 2. A slight deviation from the specs for function application. Was supposed to add constraints before inferring the argument node types, we swapped the order so that we can infer the node types so that if the parameter types mismatch we can report the given argument types. This was not supposed previously as the added constraints would block the inferring of argument types in the event of a mismatch and we cannot easily recover the given type.
> 3. Added built-in functions that accept variable number of arguments, `error`, `display`, `list`.
> 4. For final type output, normalize `[T1, List<T1>] to List<T1>`

Regarding (2) can you change the specs accordingly, in the same PR?

hilltse:
I am writing an improvement to the display of type variables

martin-henz:
Coverage decreased a bit. Is it because of the changes in this PR? Can we add test cases?

hilltse:
I see the coverage is roughly unchanged.  Can you write some tests for `InternalTypeErros.ts`?

The lines that have not been tested is listed in
#L7[https://coveralls.io/builds/30089392/source?filename=src%2FtypeChecker%2FinternalTypeErrors.ts#L7](url)

muruges95:
Ok fixed the issue. There was one error not being thrown at all. Removed that error as there was no way for it to be thrown. Should be fine now.

muruges95:
Added test programs from chapter 1 and 2 of SICP, basically the one thing that stands out is tree programs, because the way we are doing list inferencing will cause there to be an infinitely recursive type when trying to infer the the types of trees. This can be seen from the expected result from typing the tree programs in the test file. Everything else works as expected.

PR #536, opened 2020-04-13
Module Syntax Support

Hou-Rui:
Add import syntax to js-slang to use modules.
This branch is based on a very old version of js-slang and thus need manual mergings.


Hou-Rui:
> Ive advised the non-det team to avoid the confusing term "hoist" in their implementation in src/interpreter/interpreter-non-det.ts
> So for example hoistVariableDeclarations will just become declareVariables etc.
> Can you please change interpreter.ts accordingly? Thanks.

Okay, changes committed.

PR #544, opened 2020-04-14
revamp source mode to include all library functions automatically

BobbyZhouZijian:
*missed out two function names when adding them from source documentation. Now add them back.


BobbyZhouZijian:
I can try to do that

BobbyZhouZijian:
The code passed the test cases I wrote. Should be fine.

BobbyZhouZijian:
However, here is the problem. Each json file contains all function names up to and until that chapter. If I simply get the function names by accessing SourceDocumentation with a key, it would violate the 'single source of truth'. So I have to spend quite a fair amount of code selecting non-overlapping names.

PR #549, opened 2020-04-16
REPL: add variant option

### LINKED ISSUE #582: Chapter and variant combinations are undefined
The supported chapter and variant combinations are currently defined in _cadet-frontend_, in the file `src/reducers/states.ts` at  line 227:
https://github.com/source-academy/cadet-frontend/blob/1c7d40933bae38b2b922d98b844ac7d91b3db783/src/reducers/states.ts#L227

Considering the purpose of _js-slang_, it might be more appropriate for the supported combinations to be defined in _js-slang_ instead and made use of by _cadet-frontend_.

This would also help #549

arsalanc-v2:
This PR adds an option to the REPL for specifying a variant.

It also defines the supported chapter and variant combinations and uses it to check if the chapter and variant values passed to the REPL are valid.

Resolves #582


martin-henz:
Someone needs to clean this up. We should have:
```Usage: js-slang [PROGRAM_STRING] [OPTION]
  -c, --chapter=CHAPTER set the Source chapter number (i.e., 1-4) (default: 1)
  -v, --variant=..., one of default, lazy, non-det, interpreter, stepper
  -h, --help            display this help
  -e, --eval            don't show REPL, only display output of evaluation```
So lazy, interpreter and subst can all be wrapped in variant. We shouldn't entertain all the possible combinations. It's already complicated enough.

arsalanc-v2:
Resolves #582

PR #551, opened 2020-04-17
Type Inferencer for S3

raynoldng:
Add typing support for Source 3, which introduces `for`/`while` loops, reassignment (`let` statements) and arrays. 

Added `set_head` and `set_tail` to the list of predefined functions. 

Updated specs, pdf available here: 
[source_3_type_inference.pdf](https://github.com/source-academy/js-slang/files/4558166/source_3_type_inference.pdf)

# Limitations

As we require lists to be monomorphic, the type inferencer breaks on trees and streams programs. See test cases: 2.2.2 (trees) and 3.5.1 (streams)


martin-henz:
Great. FYI: Daryl is working on the specs for S2, so there would be a hand-over.

martin-henz:
Also note the merge conflicts. Please resolve.

src/errors/typeErrors.ts
```diff
@@ -1,11 +1,111 @@
...
+  ) {}
+
+  get location() {
+    return this.node.loc!
+  }
+
+  public explain() {
+    const [varName, assignmentStr] = formatAssignment(this.node)
+    return stripIndent`
+    Expected reassignment of ${varName}:
```
martin-henz, line 89: How about:
Expected assignment of ${varName}:
...
to get a value of type:
...
but got a value of type:
...

raynoldng: Yes, that is easier to read. Thanks for the suggestion!

martin-henz: I suggested the use of the term "assignment" and not "reassignment". Any assignment is a "reassignment", so the "re" doesn't add anything.

src/typeChecker/__tests__/typeChecker.test.ts
```diff
@@ -579,10 +600,485 @@ describe('type checking functions used in polymorphic fashion', () => {
...
+    `
+    const [program, errors] = parseAndTypeCheck(code, 3)
+    expect(topLevelTypesToString(program)).toMatchInlineSnapshot(`
+      "arr: Array<number>
+      y: number
+      x: number"
+    `)
+    expect(parseError(errors)).toMatchInlineSnapshot(`
+      "Line 4: Array expected type: Array<number>
+          but got: boolean"
```
openorclose, line 1001: This error seems rather confusing.  Array<number> is the type of the expected array but boolean is the type of the expected array _element_ 

Ideally we should be consistent and display either the array type of both or the element type of both

PR #552, opened 2020-04-18
Add support for type inference display

BobbyZhouZijian:
Added a method in index.ts that transfers the node at a designated location to a string representation of the type inference.

### DIFF
--- a/src/editors/ace/modes/source.ts
+++ b/src/editors/ace/modes/source.ts
@@ -101,11 +101,12 @@ export function HighlightRulesSelector(id: number, variant: Variant = 'default')
         !(name in SourceDocumentation.builtins['3'])
       ) {
         func4 += '|' + name
+      } else if (SourceDocumentation.builtins['4'][name]['meta'] === 'const') {
+        constants += '|' + name
       }
     }
 
     func4 = func4.substr(1)
-    constants = constants.substr(1)
 
     const chapter4 = {
       keywords: '',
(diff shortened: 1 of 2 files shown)


martin-henz:
Having second thoughts about it. I don't see why we cant just take the type from the node that findNodeAt returns.

openorclose:
> Having second thoughts about it. I don't see why we cant just take the type from the node that findNodeAt returns.

Could this be ambiguous?

`|` represents the cursor.
```
1|+2;
```
should this report the type of + or of 1?

Similarly, where should the cursor be to report the type of the function call?

```
fn(x);
```

I think restricting to only names for now is reasonable.

BobbyZhouZijian:
I do have a question. If there is a syntax error in the program. Should I still try to give the type inference as best as the program can, or should we just display the a parse error reported by the typedCheck function?

BobbyZhouZijian:
Or maybe should we display both?

BobbyZhouZijian:
I think reporting both should be the better way of doing it.

martin-henz:
> I think reporting both should be the better way of doing it.

I think so too. But if it's too much trouble, I'm also ok with not displaying anything if the program isn't syntactically correct.

martin-henz:
> > Having second thoughts about it. I don't see why we cant just take the type from the node that findNodeAt returns.
> 
> Could this be ambiguous?
> 
> `|` represents the cursor.
> 
> ```
> 1|+2;
> ```
> 
> should this report the type of + or of 1?
> 
> Similarly, where should the cursor be to report the type of the function call?
> 
> ```
> fn(x);
> ```
> 
> I think restricting to only names for now is reasonable.

OK, let's stick to names for now. 

It would be interesting, later, to extend this to any expression and statements. If the begin/end position information is precise, I think we can establish an understanding what cursor position refers to what node, but that's for another day, then.

PR #561, opened 2020-04-20
Fix a bug in Buffer, implement VM-internal functions in the compiler, and expose SVML compile and assemble

angelsl:


angelsl:
I'll add some tests.

PR #564, opened 2020-04-22
Lazy lists support for lazy source 2

Ahmedjjj:
This pull request adds support for lazy lists for source chapter 2.
The pair and list operators are now lazy: pairs of thunks are created.
Head and tail will return the corresponding thunks.
Tests have been added in order to test the functionnality of these lazy pairs.


jansky:
I've gone ahead and untracked those files and resolved the merge conflicts.

PR #577, opened 2020-04-25
Fix circular referencing for display in Source 3 Concurrent

jonchan51:


martin-henz:
Fixes https://github.com/source-academy/js-slang/issues/578

martin-henz:
Fixes https://github.com/source-academy/cadet-frontend/issues/1089

PR #585, opened 2020-04-29
Better type inference display and highlighting support

BobbyZhouZijian:
* fix bug when the display of the function/variable declarations were shown as 'undefined'
* simplify code structure for the type inference method.
* change all builtin functions/constants to white color.
* need to make changes to cadet frontend correspondingly. Wait for js-slang to bump first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ node_modules
 *.map
 dist/
 .idea/
+coverage/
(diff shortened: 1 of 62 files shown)


BobbyZhouZijian:
need to bump the version before making pr in cadet-frontend

PR #587, opened 2020-04-30
cleaning up .gitignore

martin-henz:


Meowzz95:
image

I think we need to !ignore a few more files. Especially `scripts/updateAutocompleteDocs.js` this one which causes build failure when it is missing.

PR #632, opened 2020-06-12
Update dependencies again

angelsl:


angelsl:
This makes no sense, the coverage dropped in the push run ... but not the PR run. (And no code was changed anyway.)

Concurrency is hard. Sigh.

PR #656, opened 2020-07-05
Stepper: parentheses around operator combinations

zhaojj2209:
Fix for [cadet-frontend issue #1343](https://github.com/source-academy/cadet-frontend/issues/1343).

All binary/conditional expressions that are part of a binary expression will now be placed in brackets when it is being evaluated in the stepper.


martin-henz:
fixes https://github.com/source-academy/cadet-frontend/issues/1343

Issue #661, opened 2020-07-11
MISC library: rename runtime to get_time


martin-henz:
The proposed name is more appropriate, since it measures wall-clock time since Jan 1 1970 and not "runtime".

PR #666, opened 2020-07-14
renames runtime to get_time

martin-henz:
fixes https://github.com/source-academy/js-slang/issues/661


martin-henz:
I had to comment out test cases in the following files:
src/transpiler/__tests__/native.ts
src/__tests__/stringify.ts
Look for // MH: 14/7/2020

src/__tests__/block-scoping.ts
```diff
@@ -69,7 +69,10 @@ test('for loops use block scoping instead of function scoping', () => {
     test();
   `,
     { chapter: 3, native: true }
-  ).toMatchInlineSnapshot(`true`)
+  ).toMatchInlineSnapshot(`
+            "native:undefined
```
thomastanck, line 73: This looks strange, this change should not be affecting these snapshots.

thomastanck:
Failing test is due to #688, so it should be as good as passing.

PR #687, opened 2020-07-26
Centralize all infinite loop messages

JoeyChenSmart:
related: see https://github.com/source-academy/cadet-frontend/pull/1387

'Collecting' all infinite loop messages into a single Enum.
Mainly to create the `infiniteLoopErrorType` function that will be used in the above PR.


angelsl:
Ah, this wasn't exactly what I meant 😅 My bad.

I meant having something like this:

```ts
type InfiniteLoopError = 'stack_overflow_error' | 'no_base' | ...;
```

and then an extra field in the `Error` object so that we can compare on a type-checked value instead of the error message.

JoeyChenSmart:
> Ah, this wasn't exactly what I meant 😅 My bad.
> 
> I meant having something like this:
> 
> ```tsx
> type InfiniteLoopError = 'stack_overflow_error' | 'no_base' | ...;
> ```
> 
> and then an extra field in the `Error` object so that we can compare on a type-checked value instead of the error message.

Oh ok I see what you mean, but I don't think we can implement it this way (easily), because of how the infinite loop detector throws the error.

Basically the infinite loop error message is 'injected' into the source code, for example:
```
function f(x) {
  return f(x);
}
```
will have a few lines added before being executed:
```
function f(x) {
  if (true) {
    error("Infinite recursion (or runtime error) detected. Did you forget your base case?");
  } else {}
  return f(x);
}
```
so we won't be able to add the type information to the errors.

angelsl:
Oh I see. I just looked at how the infinite loop detection works&mdash;so it actually transforms the Source AST, so we're limited to Source. Oh well.

This will do then I guess.

PR #690, opened 2020-07-27
Fix backtracking for Source 3 non-det variant

### LINKED ISSUE #639: Source 3 Non-det: backtracking not working correctly
here is the example:
https://share.sourceacademy.nus.edu.sg/thestudentwiththecateats

If you run the program, you see that parsing is successful. The call of parse_verb_phrase successfully returns, and unparsed is null. But then, when parse_sentence returns, unparsed is back to ["with", ["the", ["cat", ["eats", null]]]]. That is I think a bug in the non-det interpreter.

### LINKED ISSUE #689: Source 3 Non-det: backtracking not working with loops
image

Here's the example shown:
https://share.sourceacademy.nus.edu.sg/3o

After the second backtracking, c seems to be set to an incorrect value of 2 for mysterious reasons.

anubh-v:
Fixes #639 
Fixes #689
I have adopted the examples in both bug reports as test cases for the non-det interpreter.


martin-henz:
Is this ready for review? (currently has draft status)

anubh-v:
> Is this ready for review? (currently has draft status)

I still need to finish verifying the behaviour of for-loops and while-loops, but other aspects can be reviewed.

martin-henz:
OK. Will wait then.

thomastanck:
https://github.com/source-academy/js-slang/blob/master/src/__tests__/non-det-sicp-examples.ts#L2

Could you move this helper function somewhere else (e.g. src/utils/testing.ts)? Currently by importing from non-det-interpreter, those tests are being run twice. Also, most other testing code (except the ones testing the newer projects) use one or more util functions from there already.

Also adding the examples from Martin and I to the tests would be great too (just call it regression or something).

anubh-v:
Thanks @thomastanck for the bug report.

> Also adding the examples from Martin and I to the tests would be great too (just call it regression or something).

I have added both examples as test cases.
They are working now.

> Could you move this helper function somewhere else (e.g. src/utils/testing.ts)? Currently by importing from non-det-interpreter, those tests are being run twice

Thanks for highlighting this. Can we do this in a separate PR? This way, each PR can be more atomic.

anubh-v:
The PR is ready for review

martin-henz:
Great to see it all working!

Issue #688, opened 2020-07-26
Fix unreliable transpiler time limit test


angelsl:
This test seems to fail [randomly](https://github.com/source-academy/js-slang/runs/911647214) (probably depending on the load of the server running the tests):

```
Summary of all failing tests
FAIL src/transpiler/__tests__/native.ts (26.967 s)
  ● test increasing time limit for mutual recursion

    expect(received).toBeLessThan(expected)

    Expected: < 2000
    Received:   2025

      316 |   await promise
      317 |   const timeTaken = Date.now() - start
    > 318 |   expect(timeTaken).toBeLessThan(timeout * 2)
          |                     ^
      319 |   expect(timeTaken).toBeGreaterThanOrEqual(timeout)
      320 |   return parseError(context.errors)
      321 | }

      at Object.<anonymous> (src/utils/testing.ts:318:21)
      at fulfilled (src/utils/testing.ts:5:58)
```

PR #697, opened 2020-07-31
Concurrent fix interleave test

thomastanck:
Also add a similar test using busy wait to ensure interleaving


thomastanck:
Test fails because of spurious test: #688.

Issue #583, opened 2020-04-29
Library: display(v, s) does not ensure that the second argument is a string


btzy:
Run the following code in Source Academy:

```
display("test",100);
display("test",true);
display("test",()=>{});
```

It runs without errors and produces:

```
100 "test"
true "test"
() => {} "test"
```

This is not the correct behaviour as per the library documentation.  The library documentation states that the second argument must be a string:

![image](https://user-images.githubusercontent.com/6948096/80560895-2f41a900-8a15-11ea-9019-57d648db07d1.png)

The correct behaviour should be a compilation or runtime error (at least, some error should be produced).

`error(v, s)` has the same issue.

martin-henz:
We could accept this as an undocumented feature. If we fix it, let's remember to change the type inferencer, too.

thomastanck:
> We could accept this as an undocumented feature. If we fix it, let's remember to change the type inferencer, too.

Looking at the type inferencer, supporting more than one type signature isn't a couple lines change. I think we can close this issue when #700 gets merged, and I'll make a new issue for type checking of `display`.

PR #700, opened 2020-07-31
Check that second argument of `display` is a string

dtlay:
fix and tests for issue #583


src/createContext.ts
```diff
@@ -147,10 +147,15 @@ export const importExternalSymbols = (context: Context, externalSymbols: string[
...
 export const importBuiltins = (context: Context, externalBuiltIns: CustomBuiltIns) => {
   ensureGlobalEnvironmentExist(context)
-
+  const placeholder = Symbol()
   const rawDisplay = (v: Value, s: string) =>
     externalBuiltIns.rawDisplay(v, s, context.externalContext)
-  const display = (v: Value, s: string) => (rawDisplay(stringify(v), s), v)
+  const display = (v: Value, s: any = placeholder) => {
+    if (s !== placeholder && typeof s !== 'string') {
+      throw new Error('display expects the second argument to be a string')
```
thomastanck, line 155: Is it possible to change `Error` to `TypeError`? That's the error thrown in https://github.com/source-academy/js-slang/blob/master/src/utils/rttc.ts

thomastanck:
@dtlay There's a failing test thanks to #346 being recently merged. Could you fix that by changing the failing test so it works?

PR #716, opened 2020-08-03
Allow runInContext option to not re-declare variables

anthony-halim:
`runInContext` to accept a fourth parameter, where it allows no creation of new variables. Technical changes mirrors [Native Same Env](https://github.com/source-academy/js-slang/tree/native_same_env) branch.

### DIFF
--- a/src/transpiler/transpiler.ts
+++ b/src/transpiler/transpiler.ts
@@ -133,9 +133,9 @@ function wrapWithPreviouslyDeclaredGlobals(
   nativeStorage: NativeStorage,
   globalIds: NativeIds
 ) {
-  let currentVariableScope = nativeStorage.globals!.previousScope
-  let timesToGoUp = 1
-  let wrapped = create.blockStatement(statements)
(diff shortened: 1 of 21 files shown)


openorclose:
Why did all the snapshots change? This shouldn't affect the running of any program unless the new flag is passed.

src/transpiler/transpiler.ts
```diff
@@ -133,9 +133,9 @@ function wrapWithPreviouslyDeclaredGlobals(
   nativeStorage: NativeStorage,
   globalIds: NativeIds
 ) {
-  let currentVariableScope = nativeStorage.globals!.previousScope
-  let timesToGoUp = 1
-  let wrapped = create.blockStatement(statements)
+  let currentVariableScope = nativeStorage.globals!
+  let timesToGoUp = 0
+  let wrapped = statements
```
thomastanck, line 138: @openorclose this seems to have removed a block layer. Looking at the first few edited snapshots, it looks like it was useless to begin with? Do you know why we need this layer?

PR #717, opened 2020-08-04
Source 4 GPU minor refactor

thomastanck:
Try to reduce the influence of GPU on transpiler's code.

The blacklist introduced by GPU doesn't seem easy to avoid, but certain things
could have simply used the existing transpiler framework.
- `native.gpu` to store gpu internal functions is not needed as we
  already have `native.operators`.
- There is no need for `transpileToGPU` to return statements that are
later injected -- just inject the code right away.

The current design now simply has GPU doing a first pass, with
transpiler doing a second pass, with all GPU-related calls getting blacklisted
and thus not processed by the transpiler.


thomastanck:
Closing as I did manage to figure out how to remove blacklist after all.

PR #721, opened 2020-08-05
Fix nested for-loop compilation

### LINKED ISSUE #669: SVML compiler miscompiles nested for loops
The following snippet produces the wrong result when run on both the Source §3 Concurrent VM and the Sinter VM:

```js
for (let j = 0; j < 10; j = j + 1) {
    for (let i = 0; i < 1000; i = i + 1) {}
    display(j);
}
```

so I assume it is a miscompilation.

jonchan51:
Fixes #669

### DIFF
--- a/src/vm/svml-compiler.ts
+++ b/src/vm/svml-compiler.ts
@@ -1001,20 +1001,16 @@ function transformForLoopsToWhileLoops(program: es.Program) {
       if (init!.type === 'VariableDeclaration') {
         const loopVarName = ((init as es.VariableDeclaration).declarations[0].id as es.Identifier)
           .name
+        const newLoopVarName = loopVarName + '-loop-control-var' // use '-' for invalid name
+        const copyOfNewLoopVarName = 'copy-of-' + newLoopVarName
         const innerBlock = create.blockStatement([
-          create.constantDeclaration(
-            loopVarName,
-            create.identifier('copy-of-loop-control-var') // purposely long to reduce unintentional clash
-          ),
+          create.constantDeclaration(loopVarName, create.identifier(copyOfNewLoopVarName)),
           body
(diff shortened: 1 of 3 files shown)


src/vm/svml-compiler.ts
```diff
@@ -1001,20 +1001,16 @@ function transformForLoopsToWhileLoops(program: es.Program) {
       if (init!.type === 'VariableDeclaration') {
         const loopVarName = ((init as es.VariableDeclaration).declarations[0].id as es.Identifier)
           .name
+        const newLoopVarName = loopVarName + '-loop-control-var' // purposely long to reduce unintentional clash
+        const copyOfNewLoopVarName = 'copy-of-' + newLoopVarName
```
martin-henz, line 1005: to be safe you might use a string that is not allowed as a name in JavaScript, maybe

"*" + loopVarName + "*"

jonchan51: Forgot to remove the comment, "-" is already an invalid character for identifiers

martin-henz: Oh, didn't look close enough. But then why so long?

martin-henz: I guess it's nice and clear for debugging. Just remove the comment, then.

thomastanck: Probably not a bad thing to have long identifiers. It might also come in handy if manually debugging the VM, maybe.

jonchan51:
Yea, this was a result of mixing renaming and creating new environments for loops. Previously, I did not rename variables within loops because a new environment was being created, which is a decision that didn't make any sense. I changed it such that renaming will happen in loops as well, which should address the issue.

PR #725, opened 2020-08-07
Support toReplString, allowing overriding of `stringify` default behaviour

thomastanck:
Needed for https://github.com/source-academy/cadet-frontend/pull/1441


src/utils/stringify.ts
```diff
@@ -117,6 +117,8 @@ ${indentify(indentString.repeat(indentLevel), valueStrs[1])}${arrSuffix}`
       return v.toString()
     } else if (ancestors.size > MAX_LIST_DISPLAY_LENGTH) {
       return '...<truncated>'
+    } else if (typeof v.toReplString !== 'undefined') {
```
angelsl, line 120: Would it be better to check if `typeof v.toReplString === 'function'`?

thomastanck: Done.

PR #726, opened 2020-08-08
Improve code quality for check that ensures display's 2nd argument is a String.

dtlay:
fixes #700


martin-henz:
Can we have a better name for this PR? Any description what issue it addresses would be helpful. @dtlay

btzy:
I think this PR highlights a related bug in `display` (and perhaps other stdlib functions):

Both the following calls should be an error according to the spec, but are currently not an error:
```
display();
display(1,"T",3);
```

dtlay:
this PR is too messy, sorry.

PR #729, opened 2020-08-10
Update Source §1 WebAssembly spec for Sourceror 0.7.0

btzy:
This PR updates the docs and specs for Source §1 WebAssembly.  It also splits out MATH from MISC in the spec for all variants (I think this might have been an oversight previously).


btzy:
Btw, the deployed version of Sourceror is currently not in sync with the docs and specs.  This PR will bring the spec back in sync, so I think it would be best not to leave this open for too long.

PR #732, opened 2020-08-12
Improve code quality

### LINKED ISSUE #583: Library: display(v, s) does not ensure that the second argument is a string

dtlay:
this pr improves code quality for #700 
#700 fixes #583 by including a check to ensure display's 2nd argument is a String

however, this PR seems to be failing a test suite for unrelated reasons

### DIFF
--- a/docs/specs/source_studio_2.tex
+++ b/docs/specs/source_studio_2.tex
@@ -1,0 +1,61 @@
+\input source_header.tex
+
+\begin{document}
+	%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%	\docheader{2021}{Source}{Studio 2}{Martin Henz}
+        
+  \thispagestyle{empty}
+  
+\markright{CS1101S AY2021/S1, Studio Week 2}
+\begin{center}
+  {\Large {\bf Background for Studio Week 2}}\\[10mm]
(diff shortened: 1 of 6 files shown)


btzy:
Looks okay in principle, but I'm not familiar with most of js-slang so I would prefer someone else (e.g. @thomastanck) to look at it.

Also @dtlay it should be possible to tag me from this PR even though I'm not previously involved in it.  You don't need to reply to one of my issues to catch my attention.

btzy:
Whoops, I did `git push -f dtlay HEAD~2:ownself_check_ownself` to see if this PR is causing the test failure, but apparently GitHub will close the issue automatically when there are no more commits.  So now I'm locked out of your branch and can't push the two commits back.

If you don't mind, can you push those two commits back?  (If you've lost them, I've also pushed them onto [this branch](https://github.com/btzy/js-slang/tree/dtlay-display) so you can grab them from there.)

PR #741, opened 2020-08-25
Separate running of benchmarks to its own command `yarn benchmark` 

thomastanck:
Also increase `list-benchmark.ts` timeout to 1000s.


openorclose (approved):

.github/workflows/nodejs.yml
```diff
@@ -29,6 +29,6 @@ jobs:
     - run: yarn autocomplete
     - run: yarn format:ci
     - run: yarn tslint
-    - run: yarn test --testPathIgnorePatterns='.*benchmark.*'
```
angelsl, line 32: This is not necessary, `yarn test` runs exactly the script you change in `package.json`.

thomastanck: not sure what you mean. before, the flags were being added here. i've moved the flags to be added in package.json. isn't this the same as what you're suggesting?

angelsl: Oops sorry. Misread the diff..

PR #747, opened 2020-08-28
Fix infinite loop detection false positives in functions shadowed by their parameters

JoeyChenSmart:
Functions shadowed by their parameters cause false positives in the infinite loop detector. E.g:
```
function f(f){
    return f(1);
}

f((x)=>x);
```

will result in an `Error: "Infinite recursion (or runtime error) detected. Did you forget your base case?"`.

This was caused by a small bug in the symbolic executor where shadowed parameters were overlooked.

Also added test cases for this.


angelsl:
Whoa, what's with the crazy commit history here o_o

But it's OK, it'll get squashed.

JoeyChenSmart:
Yeah sorry, I'm a bit new to SE and git :S, I'll learn how to make cleaner PRs in the future.

martin-henz (approved):
Looks good!

PR #749, opened 2020-08-29
Fix equals

### LINKED ISSUE #748: equal in Source §2
Currently we have a discrepancy between the specs and the implementation.

Specs:
https://source-academy.github.io/source/source_2.pdf#page=9

Implementation:
Screenshot 2020-08-29 at 4 29 25 PM

The difference shows up here, for example:

equal(1, x => x);
false
equal(x => x, 1);
Line 27: Expected string or number on left hand side of operation, got function.

openorclose:
Fixes #748


martin-henz (approved):
Yes, that's the right equal

thomastanck (approved):
Yes, agreed! very nice.

Issue #736, opened 2020-08-20
Stepper: Parentheses missing in some cases


martin-henz:
https://share.sourceacademy.nus.edu.sg/9c

<img width="824" alt="Screenshot 2020-08-20 at 10 54 02 AM" src="https://user-images.githubusercontent.com/30220768/90711638-d3f13a80-e2d3-11ea-806e-b093dadaf376.png">

parentheses disappear in the next step:

<img width="824" alt="Screenshot 2020-08-20 at 10 53 03 AM" src="https://user-images.githubusercontent.com/30220768/90711629-ca67d280-e2d3-11ea-87e9-6756ac523e3c.png">

PR #764, opened 2020-09-05
Stepper error check

zhaojj2209:
Added error checking to stepper: the code is run normally in the editor first, then if the result returns an error, an error is returned in the stepper as well.

The error message directs students to check their code in the editor, which is more useful compared to before, when the stepper would just not run if the code is erroneous. The error check also prevents issues where the stepper evaluates the code wrongly (due to issues such as variable capture) when the editor throws an error.

image

Also included: fix to #736 that makes bracketing more consistent.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -83,6 +83,17 @@ if (typeof window !== 'undefined') {
 
 let verboseErrors = false
 const resolvedErrorPromise = Promise.resolve({ status: 'error' } as Result)
+const stepperErrorPromise = Promise.resolve({
+  status: 'finished',
+  value: [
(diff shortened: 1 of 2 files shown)


martin-henz (changes requested):
I don't think this is a good idea. For example, we want to run programs in the Stepper that do infinite recursion. We will just run into the step limit. This would not be possible any longer.

Can you be more selective? Only do the checking phase of the transpiler, and not the execution.

martin-henz:
fixes https://github.com/source-academy/cadet-frontend/issues/1508

PR #790, opened 2020-10-11
Updates to Source §4 GPU

rrtheonlyone:
After further testing with Source §4 GPU, some issues have creeped up. This PR will help to fix these and discuss potential consequences:

## Assignment 

We initially allowed for Array Assignment for the GPU result. However, this runs into an issue as follows:  

```javascript
const N = 400;

const arr = [];
for (let i = 0; i < N; i = i + 1) {
    arr[i] = [i, i + 1, i + 2, i + 3];
}


const res = [];
for (let i = 0; i < N; i = i + 1) {
    res[i] = arr[N - 1 - i];
}

res;
``` 

When we run on the GPU, we actually require the exact dimensions of the output array to be passed in (and it has be constant e.g 10 x 10 array). This can be inferred for simple cases but for cases like the above, it becomes a non-trivial problem (esp if it is a nested array). 

In order to prevent laborious runtime checks being done by the language, we leave it up to the programmer to tell us the dimensions through the for loops. **We ban assignment done to anything that is not a number** (the current implementation allows for number or array of numbers). 

Hence the above code should be written as follows:

```javascript
const N = 400;

const arr = [];
for (let i = 0; i < N; i = i + 1) {
  for (let j = 0; j < 4; j = j + 1) {
     arr[i][j] = i + j;
  }
}

const res = [];
for (let i = 0; i < N; i = i + 1) {
  for (let j = 0; j < 4; j = j + 1) {
     res[i][j] = arr[N - 1 - i][j];
  }
}

res;
``` 

The documentation and code has been updated for this case.

## Quick Return 

Another issue that might cause difference in behaviour is an early return. When the assignment statement is written somewhere else, the program might return earlier than expected. 

An example is as follows:

```javascript
function builder(res) {
    for (let i = 0; i < 400; i = i + 1) {
        const row = [];
        res[i] = row;
        row[0] = 1;
    }
    return res;
}

const res = [];
builder(res);
```

We expect `res` to have 400 elements each with `[1]`. However with the current version of the GPU language, we will get 400 elements each with `[]`. This is because we return early for each GPU kernel thread. This issue will also be conveniently solved by the above fix.


limhawjia (approved):
LGTM

PR #827, opened 2020-12-12
Make assignment an expression

### LINKED ISSUE #820: Make assignment expressions
> That is consistent with JavaScript, and makes it easier to explain why they have a value, and declarations dont.

From @martin-henz

openorclose:
Fixes #820


martin-henz:
So basically: no change in the evaluators? Including the interpreter?

openorclose:
> So basically: no change in the evaluators? Including the interpreter?

Yeah no change, but I'll add a test for the evaluation of assignment statements as a basic consistency check

martin-henz (approved):
Looks great. No change!

PR #839, opened 2021-01-14
Updates README.md to use yarn link rather than manual copying

Gopiandcode:
The standard way of using local versions of JS packages in a project is through the `yarn link` mechanism. This pull request updates the js-slang readme to use this process rather than manually copying the files (as was previously recommended).


martin-henz (approved):
Great. Works!

Issue #765, opened 2020-09-08
Stepper: wrong result


martin-henz:
telegram "CS1101S 2021 All Staff": uwu, [8 Sep 2020 at 1:24:59 PM]:
Hi all, student discovered a bug in the stepper

```
function f(w) { 
    return 2 * g(w);
}
function g(x) {
     return w + x;
}

const w = 5;
f(w + 2);
```

(RA1, question 2)

Answer: 24
REPL returns 24
Stepper returns 28

For some reason when the stepper is invoked, the function call g(7) is replaced by return 7 + 7 instead of return 5 + 7

martin-henz:
fixed by #844

PR #844, opened 2021-01-25
Capture-avoiding Substitution

Zacchua:
Adds a test case under js-slang/src/stepper/__tests__/stepper.ts named scoping test, to test for the scoping issues found in issue #765


martin-henz (approved):
Very good!

PR #855, opened 2021-02-01
Updated to import new modules in js-slang

MarcusTXK:
Modified the intepreter.ts and transpiler.ts to use the new modules. 

In intepreter.ts, `functions` are now from destructured from the new module object being returned from `loadModule` instead of being directly returned. 

In transpiler.ts, the functions from the imported `moduleText` now have to be accessed from the new module object, as the functions are no longer directly returned. A constant `MODULE_FUNCTIONS` was added to access the functions. `substring` was used to remove the last character of the `moduleText `, as after transpiling it in babel, there would be a semi-colon there and leaving the semicolon causes an error when the `sandboxedEval` function when it tries to evaluate it.

In __tests__/moduleLoader.ts, the tests were updated to reflect the changes to the new module system

New module object structure:
```
    (_params) => {
      ...
      return {
        functions: {
          ...
        },
        sideContents: [...],
      };
    }
```

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -7,6 +7,7 @@ export const TRY_AGAIN = 'try_again' // command for Source 4.3
 export const GLOBAL = typeof window === 'undefined' ? global : window
 export const NATIVE_STORAGE_ID = 'nativeStorage'
 export const MODULE_PARAMS_ID = 'moduleParams'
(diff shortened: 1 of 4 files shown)


openorclose (approved):
Looks good

PR #891, opened 2021-03-02
Stepper: various bug fixes

### LINKED ISSUE #872: Stepper: Renaming for BlockStatements occur before substitution
Capture Avoiding Substitution is implemented in blocks by renaming variables when an expression to be substituted into the block has a free variable with the same name. However the renaming occurs in the first step of the substitution, instead of at the step where substitution occurs
Screenshot 2021-02-17 at 4 41 58 PM

in the above case constant "z" in function f is renamed to "z (const)" as z occurs free in the body of g, and g is called in f, and the body of g is hence substituted into f.

### LINKED ISSUE #873: Stepper: Fresh names during renaming
Currently to implement capture-avoiding substitution, alpha-renaming is used when an expression is substituted into a function, and the function has the parameters with the same name as free variables in the expression. The parameters are renamed from "name" to "name + (param)".

Screenshot 2021-02-17 at 4 57 32 PM
Screenshot 2021-02-17 at 4 57 45 PM

The same occurs when substituting into another block, constants with the same name as free variables are renamed to "name + (const)".

Screenshot 2021-02-17 at 4 41 58 PM

Variable renaming can become confusing, and in recursive functions, the " (param)" tag could become long and cumbersome. These new names are also not proper syntax, and may confuse users.

Therefore there is a need for fresh names to be 
- syntactically correct
- fresh
- identifiable as "generated"
- as short as possible
- visibly associated with the original

### LINKED ISSUE #874: Stepper: Incorrect result
Stepper gives an incorrect result of "g();" instead of "1;" in the following function 

Screenshot 2021-02-17 at 5 56 41 PM

Screenshot 2021-02-17 at 5 39 02 PM

### LINKED ISSUE #890: Stepper: return in block not working correctly
function f(x) {{ return 1; }}
f(3);

should give 1, but currently gives undefined.

### LINKED ISSUE #894: Stepper: Renaming to non-fresh names
The stepper currently does not rename to fresh names, effectively failing to avoid variable capture.

image

image

image

The REPL gives 1 where as the stepper gives 0.

Although `x_2` is renamed to a fresh name (`x_3`), `x_1` is renamed to `x_2` effectively capturing `x_2` in `() => x_1 + x_2`.

Zacchua:
Summary:

fixes #872
fixes #873
fixes #874 
fixes #890 
fixes #894

Modified BlockStatement reducer such that return in blocks works correctly (issue #890) , also addresses fresh names issue (#873 ), as well as incorrect result issue (#874 ), however does not resolve substitution into infinite structure issue.


martin-henz (changes requested):
You say you fixed:
https://github.com/source-academy/js-slang/issues/890
but there is no corresponding test case added!

martin-henz (approved):

PR #907, opened 2021-03-18
Stepper: Fixed issue #898, renaming not occuring in blocks

### LINKED ISSUE #898: Stepper: Renaming not occurring in blocks
The renaming for constant declarations and function declarations does not occur in blocks.

<br>

**Constant declaration**

REPL gives `1`, stepper gives `0`.

`x_1` of `const x_1 = 0;` should be renamed, as `x_1` occurs free in `g`.

<br>

**Function declaration**

REPL gives `1`, stepper gives `x_1`.

`x_1` of `function x_1() {...}` should be renamed, as `x_1` occurs free in `g`.

Zacchua:
fixed #898 and abstracted away renaming process into function getFreshNames


martin-henz (changes requested):
Regarding
```
function scanOutDeclarations(
  node: es.BlockStatement | BlockExpression | es.Expression
): es.Identifier[] {
  const declaredIds: es.Identifier[] = []
  if (node.type == 'ArrowFunctionExpression') {
    for (const param of node.params) {
      declaredIds.push(param as es.Identifier)
    }
  } else if (node.type == 'BlockExpression' || node.type == 'BlockStatement') {
    for (const stmt of node.body) {
      // if stmt is assignment or functionDeclaration
      // add stmt into a set of identifiers
      // return that set
      if (stmt.type === 'VariableDeclaration') {
        stmt.declarations
          .map(decn => (decn as es.VariableDeclarator).id as es.Identifier)
          .forEach(name => declaredIds.push(name))
      } else if (stmt.type === 'FunctionDeclaration' && stmt.id) {
        declaredIds.push(stmt.id)
      }
    }
  }
  return declaredIds
}
```

What is the purpose of the first case:  `if (node.type == 'ArrowFunctionExpression')`?

Is `scanOutDeclarations` ever called on an `ArrowFunctionExpression`?

Zacchua:
> Regarding
> 
> ```
> function scanOutDeclarations(
>   node: es.BlockStatement | BlockExpression | es.Expression
> ): es.Identifier[] {
>   const declaredIds: es.Identifier[] = []
>   if (node.type == 'ArrowFunctionExpression') {
>     for (const param of node.params) {
>       declaredIds.push(param as es.Identifier)
>     }
>   } else if (node.type == 'BlockExpression' || node.type == 'BlockStatement') {
>     for (const stmt of node.body) {
>       // if stmt is assignment or functionDeclaration
>       // add stmt into a set of identifiers
>       // return that set
>       if (stmt.type === 'VariableDeclaration') {
>         stmt.declarations
>           .map(decn => (decn as es.VariableDeclarator).id as es.Identifier)
>           .forEach(name => declaredIds.push(name))
>       } else if (stmt.type === 'FunctionDeclaration' && stmt.id) {
>         declaredIds.push(stmt.id)
>       }
>     }
>   }
>   return declaredIds
> }
> ```
> 
> What is the purpose of the first case: `if (node.type == 'ArrowFunctionExpression')`?
> 
> Is `scanOutDeclarations` ever called on an `ArrowFunctionExpression`?
> 
> Also the comment in the second case is wrong: The comment talks about assignment, but the implementation is about `VariableDeclaration`.
> 
> And what about constant declarations?

The first case is used when substituting into a nested lambda function eg. x => x_1 => { .. }, to ensure that x is not renamed to x_1. scanOutBoundNames, similar to scanOutDeclarations, but used on functions and lambda functions, is called on the lambda function's body, which in the example above, is itself a lambda function. scanOutBoundNames is a new function added that returns all the bounded names within a function block.

I believe constant declarations are handled in the same block as variable declarations.

martin-henz:
> I believe constant declarations are handled in the same block as variable declarations.

Then it's not properly named. Can you fix the naming?

martin-henz:
> The first case is used when substituting into a nested lambda function eg. x => x_1 => { .. }, to ensure that x is not renamed to x_1. scanOutBoundNames, similar to scanOutDeclarations, but used on functions and lambda functions, is called on the lambda function's body, which in the example above, is itself a lambda function. scanOutBoundNames is a new function added that returns all the bounded names within a function block.

Are you following your specs on this?

Zacchua:
> Are you following your specs on this?

Yes, I believe so prof, because it ensures that parameters would not be renamed to any name that occurs in the block, including parameters of functions declared within the block, which may cause it to be captured. There is a testcase for this, "renaming clash with parameter of lambda function/function declaration in block".

Zacchua:
> Then it's not properly named. Can you fix the naming?

Will do, prof

martin-henz:
> > Are you following your specs on this?
> 
> Yes, I believe so prof, because it ensures that parameters would not be renamed to any name that occurs in the block, including parameters of functions declared within the block, which may cause it to be captured. There is a testcase for this, "renaming clash with parameter of lambda function/function declaration in block".

OK.

PR #908, opened 2021-03-21
Modify context to pass in historically created frames

CrownKira:
### Description
This pull request contains some modifications to `createContext.ts` and `interpreter.ts` to prebuild and pass in the environments, to the context, in the form of a tree structure, namely the `environmentTree` (encapsulated in an `EnvTree` class)

An `EnvTree` is a tree structure in which each node contains a pointer to an environment, with the parent being a node that contains the tail of this environment and the children being the nodes that each points to an environment that has this environment as their tail environment. 

It is quite a mouthful but it is very intuitive. An environment is a linked list of frames. You can just imagine each pair in the linked list as a node in our `EnvTree` and each arrow being an edge in the tree that links the two `EnvTreeNode` ...and tadaaa you get the tree!!

The Environment Visualizer Tool currently makes use of the linked list environment to draw the environment model, which does not make sense in terms of the nature of the environment model itself (which is inherently a tree). This has resulted in costly preprocessing on our side. With this new structure, a huge portion of preprocessing can be cut and it will make our lives a lot easier. 

### How to test 
Jest tests have been added under `src/__test__/environmentTree.ts`, so there is not much to test actually.


openorclose (approved):
Overall looks fine, it really does seem a bit redundant though since we're storing the same information twice.

Is there no way to do the preprocessing on the frontend side all at one go?

src/__tests__/environmentTree.ts
```diff
@@ -0,0 +1,66 @@
...
+
+test('EnvTree root should be null upon instantiation', () => {
+  const envTree = new EnvTree()
+  expect(envTree.root).toBeNull()
+})
+
+test('EnvTree::insert should insert the globalEnvironment as the root', () => {
+  const envTree = new EnvTree()
+  envTree.insert(createGlobalEnvironment())
+  expect(envTree.root).toMatchSnapshot()
```
openorclose, line 13: I think we should check whether it is === the env directly rather than using snapshots

src/__tests__/environmentTree.ts
```diff
@@ -0,0 +1,66 @@
...
+})
+
+test('EnvTree::getTreeNode should return the tree node that contains a pointer to the given environment', () => {
+  const envTree = new EnvTree()
+  const globalEnvironment = createGlobalEnvironment()
+  envTree.insert(globalEnvironment)
```
openorclose, line 19: here also, just check directly

PR #926, opened 2021-04-04
Updated the specifications for the stepper

petermonky:
Updated the specfications for the stepper, namely adding sections for **Substitution** and **Free names**.

**Substitution** explains how substitution of expressions occurs during the reduction process. The _function declaration_ subsection uses multiline fractions to keep expressions within the text width. The notation for substitution has also been changed from "←" to ":=" as to stay true to lambda calculus conventions.

**Free names** explains the rules for which the stepper extracts free names for an expression. A new relation "▷" is defined to draw the relationship between an expression and the set of free names of said expression.

The header was also updated accordingly.


martin-henz (approved):
Looks good!

PR #928, opened 2021-04-04
Refactor parser and format parser tests

thomastanck:
This is in preparation for an future change to add support for Typescript syntax.

This PR first enforces consistent `looseParse` use to decrease API surface area and hopefully make the transition to Typescript slightly smoother. Currently, only the name extractor requires `looseParse`. Since Typescript syntax is either for advanced users or for teaching the later parts of the module, the name extractor does not have to support Typescript yet.

Secondly, due to differences in column number handling when tabs are present, the existing parser tests have to be formatted correctly in order to ensure that the line and column numbers can be compared across the future syntax change.


martin-henz:
I think the best reviewer would be @openorclose . He has another review on his desk. What's your timeline for this?

openorclose (approved):
Looks simple enough, refactoring from parse to looseParse and changing tests.

PR #950, opened 2021-05-01
Fix problem with source modules import aliases

Anonymxtrix:
Fix problem with source modules import aliases.

```javascript
import { baz as foo } from "bar";

foo(); // Now works
```

Fixes source-academy/modules#59.

## Testing

To test the fixes mentioned above, follow the steps listed below. 

Fetch and checkout to the `modules/importalias` branch in `js-slang` code repository.

Build and use the new `js-slang` branch on your local `cadet-frontend` instance. 

Go to the Source Academy Playground and run the following code to test the Source transpiler for Source Module import aliases.
```javascript
import { make_empty_tree as foo } from "binary_tree";

foo(); // REPL should display "null"
``` 

Run the following code to test the Source imterpreter for Source Module import aliases.
```javascript
"enable verbose";
import { make_empty_tree as foo } from "binary_tree";

foo(); // REPL should display "null"
```

### DIFF
--- a/src/interpreter/interpreter.ts
+++ b/src/interpreter/interpreter.ts
@@ -623,11 +623,22 @@ export const evaluators: { [nodeType: string]: Evaluator<es.Node> } = {
 
   ImportDeclaration: function*(node: es.ImportDeclaration, context: Context) {
     const moduleName = node.source.value as string
-    const neededSymbols = node.specifiers.map(spec => spec.local.name)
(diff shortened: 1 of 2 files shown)


Anonymxtrix:
Do not merge first, I have only fixed the issue for the Source transpiler. I am working on the interpreter fix now.

MarcusTXK (approved):
Great work, tested and working well.

PR #956, opened 2021-05-10
Stepper: Fix substPredefined* by making Program create new scope

### LINKED ISSUE #772: Stepper: Redeclaration of predeclared functions
The stepper currently does not deal with redeclaration of predeclared functions properly. If the student declares a function with the same name as a predeclared function, the stepper will continue to use the Source predeclared function in its evaluation.

thomastanck:
Resolves #772.

In the following test case taken from #772,

```
const reverse = x => x;
reverse(0);
```

The issue was that substPredefinedFns was substituting `reverse` at the very start. When the time came for `const reverse = ...` to be substituted, the `reverse` on line 2 already had its own "name", distinct from the Program-scoped `reverse` on line 1.

The fix is to treat substituting into a program the same as BlockStatements. However, because block statements and programs have different fields in the AST type, I simply copy-pasted instead of trying to convert back and forth between the two.

Now, when substPredefinedFns tries to substitute into the program, it will first scanOutDeclarations, same as with BlockStatements. This causes the substitution to stop because reverse is bound. Then, when the time comes for `const reverse = ...` to be substituted, the `reverse` on lines 1 and 2 share the same name as they are both untouched, and the substitution succeeds.

### DIFF
--- a/src/stepper/stepper.ts
+++ b/src/stepper/stepper.ts
@@ -85,10 +85,14 @@ function scanOutBoundNames(
 }
 
 function scanOutDeclarations(
-  node: es.BlockStatement | BlockExpression | es.Expression
+  node: es.BlockStatement | BlockExpression | es.Expression | es.Program
 ): es.Identifier[] {
   const declaredIds: es.Identifier[] = []
-  if (node.type == 'BlockExpression' || node.type == 'BlockStatement') {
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
yes, looks right.

PR #963, opened 2021-05-11
Specs: reserved words of ECMAScript 2020 strict mode, in alphabetical order

martin-henz:


martin-henz:
easier to check

martin-henz:
Here are the reserved (or otherwise restricted) names of ECMAScript 2020:

await, arguments, break, case, catch, class, const, continue, 
debugger, default, delete, do, else, enum, eval, export, 
extends, false, finally, for, function, if, implements, import,
in, instanceof, interface, let, new, null, package, private, 
protected, public, return, static, super, switch, this, 
throw, true, try, typeof, var, void, while, with, yield

docs/specs/source_names_lang.tex
```diff
@@ -16,7 +16,7 @@ \subsection*{Names}
...
..., $\textbf{\texttt{delete}}$, $\textbf{\texttt{do}}$, $\textbf{\texttt{else}}$, $\textbf{\texttt{eval}}$, $\textbf{\texttt{finally}}$, $\textbf{\texttt{for}}$, $\textbf{\texttt{function}}$, $\textbf{\texttt{if}}$, $\textbf{\texttt{in}}$, $\textbf{\texttt{instanceof}}$, $\textbf{\texttt{new}}$, $\textbf{\texttt{return}}$, $\textbf{\texttt{switch}}$, $\textbf{\texttt{this}}$, $\textbf{\texttt{throw}}$, $\textbf{\texttt{try}}$, $\textbf{\texttt{typeof}}$, $\textbf{\texttt{var}}$, $\textbf{\texttt{void}}$, $\textbf{\texttt{while}}$, $\textbf{\texttt{with}}$, $\textbf{\texttt{class}}$, $\textbf{\texttt{const}}$, $\textbf{\texttt{enum}}$, $\textbf{\texttt{export}}$, $\textbf{\texttt{extends}}$, $\textbf{\texttt{import}}$, $\textbf{\texttt{super}}$, $\textbf{\texttt{implements}}$, $\textbf{\texttt{interface}}$, $\textbf{\texttt{let}}$, $\textbf{\texttt{package}}$, $\textbf{\texttt{private}}$, $\textbf{\texttt{protected}}$, $\textbf{\texttt{public}}$, $\textbf{\texttt{static}}$, $\textbf{\texttt{yield}}$, $\textbf{\texttt{null}}$, $\textbf{\texttt{true}}$, $\textbf{\texttt{false}}$. These are all words that cannot be used without restrictions as names in the strict mode of ECMAScript 2020.
```
thomastanck, line 19: class, const are not in alphabetical order

thomastanck: eval isn't an ECMAScript 2020 keyword. I disabled the noEval syntax validator in a local build, and it actually runs as expected:

image

possible new issue?

thomastanck: Just to note, the current error is "eval is not allowed", checked by us at parse time.

thomastanck: enum, export, extends are not in alphabetical order

thomastanck: Err, and from that point onwards looks like there's a lot of out of order stuff. Might need a re-sort?

martin-henz: Regarding eval: It meets my definition of "restricted word". E.g. Chrome:

Screenshot 2021-05-12 at 4 56 34 PM

martin-henz: Since strict-mode ES2020 does not allow declaration of eval, we also shouldn't. So we should list eval as "restricted word".

thomastanck: You're right, now I'm looking at the right set of docs

thomastanck (approved):

PR #965, opened 2021-05-12
svml-compiler: fix scope renaming bug by always renaming

### LINKED ISSUE #784: SVML compiler allows names declared in inner scope to be accessed from outer scope
```js
{
  let i = 5;
}
display(i);
```

displays `5`.

```js
display(i);
{
    let i = 1;
    display(i);
    {
        let i = 2;
        display(i);
    }
    i = 3;
    display(i);
}
display(i);
```

displays

```
undefined
1
2
3
3
```

<details>
<summary>Original issue</summary>

```js
for (let i = 0; i < 5; i = i + 1) {
    display(i);
}
display(i);
```

compiles and prints `5` at the end. `i` is not supposed to exist outside the `for` loop.
</details>

thomastanck:
... except top level names for prelude, which is needed to link the prelude with the main code.

Resolves #784.

The one detail that wasn't obvious is that the code to transform for loops to while loops assumes that the renaming code will be able to handle different identifiers sharing the same `loc`. So the renaming code had to be changed just a tiny bit to make this assumption correct, by working harder -- it now ensures that the renamed name has no clash with any outer scopes.


martin-henz (approved):
looks good!

PR #968, opened 2021-05-15
greatly simplify variable access/storage in transpiler

### LINKED ISSUE #538: Transpiler: Evaluation of toplevel statement sequences
See https://github.com/source-academy/js-slang/issues/537

openorclose:
this should fix #538  too

# old

first run: 
```js
let a = 1;
{
 const b = 1; // not stored
}
```
transpile->
```js
let a = 1;
{
 const b = 1; // not stored
}
globals.set("a", {getValue = () => a, setValue: v=> a = v});
```

second run:

```js
a = 2;
```
transpile->
```js
globals.get("a").setValue(1); // basically need to replace all references to `a` with this
```

# after this refactor

we expose a evaller function that gets replaced each time to close around the values

(evaller is simply `eval` the first time a program is run)
first run:
```js
let a = 1;
{
 const b = 1; // not stored
}
```
transpile->
```js
evaller = nextProgram => eval(nextProgram);
let a = 1;
{
 const b = 1; // not stored
}
```

second run:

```js
a = 2;
```
transpile->
```js
evaller = nextProgram => eval(nextProgram);
a = 2;
```
note that on the second run, this program is run in the scope of the previous program using the replaced `evaller`, so no complicated code is needed to transform top level vars into object accesses.


thomastanck (approved):
LGTM, theory checks out, tests pass, and coverage is good. Did a tiny bit of manual testing as well.

Issue #311, opened 2019-07-29
NATIVE: stack size exceeded in RUNES hangs browser


martin-henz:
Chrome in MacOS:

https://sourceacademy.nus.edu.sg/playground#chap=2&ext=RUNES&prgrm=PTAEENQBwSwDwKYBsD6BnAxgCwQWwaDGhKAGYCuAdhgC4wD2lZATvbiVPWjHY6DfVDMqCAFAhoXHgyZFQAcxgA3BEwDuPLKCXgk5BMRigAFKzUBKIeEqKb4sKVbsADP0FMAtKACMAGgiUACagAFYmGPR6uJSWzNa28ixs9qCuAqDsXt6iohTUvEwARkjgGADWKLCIqJg4%2BMYw-iH%2BlP64lgDeoqA9Qgg05MxFJeXGaACO5ODMCOYA3KIAvjl5tDKgEUj0zBSoVcjo2HgIDU0tbZ3dvVc9MwNDG5HbY5PTCE09YK2Enxn%2BzgA6ACs8yWKyoaz4m22uxQGBgzAw5BKzEq8AOtWOp1C5wyl16j0oaBooECcFAAF4MuAaFgUOBCmgGqAvEwwAAmUEEiJEkmBACelOptPpjOMYS87A5XN6POJpKFuBpdImzBoxjJACoyaAANSk-na-ky279QZMaHMF5TGb%2BYJff7%2Bbyg5a5CEFaAwWiDE7fXD%2BfY1I74fG9VYepVlBAoMzYkKhgmm%2B5MMIUtOpUAAflAExtYkTBdAAC5QIUDDBAtHHKVjOKWT5LGBmjdC63W5Ho7HGqF687fC224PeoHDnUTt3mqA-eYTaBlgTw%2BsO%2BhXjMGgnE3dzT801TXNnc28B4WS8TShVqxha0Yso3CP5j0P2%2BAoyu88zb-2n9-l12Lo-v0TWd516LcHmXQ81xiBZXTQLB6DUYxYG9NcgWcZx-DQjDHi2HZkTRapR2OGcFiAA

gives this in JavaScript console and browser hangs:

<img width="598" alt="Screenshot 2019-07-30 at 7 21 34 AM" src="https://user-images.githubusercontent.com/30220768/62089212-b6050200-b29a-11e9-8f3d-8a00d4900feb.png">

martin-henz:
This program now produces a nice error message in the REPL:

[...]

PR #969, opened 2021-05-19
Improve stringify performance

### LINKED ISSUE #371: Printing of data structures needs to be restricted
Some programs make NATIVE hang because they generate large data structures. Example:

```
function make_k_list(k, d) {
    // WRITE YOUR SOLUTION HERE.
    const degree = k;
    const depth = d;
    let repeat = 0;
    function helper(k, d, to_repeat) {
        if (d === 0 && k === 0) {
            return null;
        } else if (k === 0) {
            return helper(degree, d - 1, repeat);
        } else {
            repeat = pair(to_repeat, helper(k - 1, d, to_repeat));
            return pair(to_repeat, helper(k - 1, d, to_repeat));
        }
    }
    return helper(k, d, 0);
}

make_k_list(2,2); 
```

We need to restrict the number of nodes that are printed. This will allow us also to catch infinite data structures (data structures with loops).

thomastanck:
Resolves #371. The benchmark in the issue is still slow due to a different issue https://github.com/source-academy/cadet-frontend/issues/1494. However, the js-slang part of the benchmark now runs in around .2s on my machine.

maim9958


martin-henz:
Does this fix #311 ?

src/utils/stringify.ts
```diff
@@ -211,3 +77,579 @@ function niceTypeToString(type: Type, nameMap = { _next: 0 }): string {
...
+        suffixLast: ''
+      }
+    } else if (dag.type === 'pair') {
+      const headTree = format(dag.head)
+      const tailTree = format(dag.tail)
+      // - 2 is there for backward compatibility
+      if (
+        dag.length - 2 > splitlineThreshold ||
+        headTree.type !== 'line' ||
+        headTree.type !== 'line'
```
openorclose, line 512:
```suggestion
        tailTree.type !== 'line'
```

openorclose (approved):

PR #970, opened 2021-05-19
Make sicp module

Hungkhoaitay:
Whenever `createContext` is called, the functions and names are generated inside `createContext.ts`. Thus, they are all collected together in a hash-map `dict`, then be exported to outside.

It also satisfy the "single source of truth", since whenever any function or its name is changed, we can still collect and use them without any further changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Using your js-slang as a package
 
 To install this package
 ``` {.}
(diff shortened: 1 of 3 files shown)


martin-henz:
Separate telegram discussion led to clarified requirements. For now, I'm making this PR a "draft". No need to review yet.

PR #990, opened 2021-06-04
Improves type error message for arrays

### LINKED ISSUE #730: Source §2: Wording of error message for arrays
Screenshot 2020-08-11 at 5 47 01 PM

Currently, the error mentions "array" but Source §2 does not have arrays. For Source §2, we should say "pair" instead of "array" because all arrays in Source §2 are pairs.

martin-henz:
Uses string "compound data" instead of "array" across the board in type error messages.

In addition, when the chapter is known to be <= 2, uses string "pair" instead of "array" (all arrays in Chapters 1 and 2 are pairs).

fixes #730

### DIFF
--- a/docs/lib/list.js
+++ b/docs/lib/list.js
@@ -79,7 +79,7 @@ function list(value1, value2, ...values ) {}
  * themselves pairs), and <CODE>false</CODE> otherwise. For the "identical",
  * the values need to have the same type, otherwise the result is 
  * <CODE>false</CODE>. If corresponding leaves are boolean values, these values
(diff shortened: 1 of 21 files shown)


martin-henz:
The PR is minimalistic. I've tried to change as little as possible. In order to do a comprehensive job, the changes are far-reaching. I don't think it's worth it. So I've resorted to adding optional parameters to a couple of functions and do a best effort to find out whether we are in Source <= 2. If we don't have that info, we just use a generic term "compound data" that works for all Source languages.

Issue #410, opened 2020-02-24
Runtime error (stack overflow) prints '-1' for line number


btzy:
**Test case:**

Use the following code.
```
function test(){
    test();
}
test();
```
Then run the code.

**Actual output:**

`Line -1: RangeError: Maximum call stack size exceeded`

**Expected output:**

Line number should not be negative.

martin-henz:
This is fixed.

PR #1004, opened 2021-06-12
Avoid error messages with "Line -1"

martin-henz:


martin-henz:
Fixes #410

openorclose:
image

PR seems fine in general, but we should probably try fixing these specific errors since they shouldn't happen

PR #1008, opened 2021-06-13
Allow if without else

### LINKED ISSUE #993: else clause in conditional statements should be optional from Source §3 onwards

openorclose:
Fixes #993 

Needs an update of the syntax too


martin-henz (approved):
works as advertized. also updated the specs.

PR #1013, opened 2021-06-16
Fix SICP

Hungkhoaitay:
I had a problem with testing the package and did not detect the error in the previous version. I have fixed and now the package is working fine.


martin-henz (approved):
Looks fine. Still probably a bit roundabout, with the generation of the txt file, but acceptable.

PR #1041, opened 2021-07-09
Update infinite loop detector

### LINKED ISSUE #606: Infinite Loop Detector and number overflows
Example code:
```
function fac(x) {
    if (x===4503599627370495) {
        return 1;
    } else {
        return 2*fac(x+1-1);
    }
}
fac(4503599627370495.5);
```
This code terminates, but is detected by the Infinite Loop Detector (false positive).

JoeyChenSmart:
Replace the old infinite loop detector with a much better version.
Also closes #606 because of the new implementation.

---
Note: included the [Alt Ergo](https://alt-ergo.ocamlpro.com/) SMT solver as a local package. The file is compiled from OCaml bytecode so it's pretty unreadable (also adds 28k LoC to this PR :/).

The license for Alt Ergo is found [here](https://ocamlpro.github.io/alt-ergo/About/licenses/index.html), which is some combination of [Apache 2.0](https://ocamlpro.github.io/alt-ergo/_downloads/014a25ba9fea849c303e53622b903252/Apache-License-2.0.txt) and [their own license](https://ocamlpro.github.io/alt-ergo/_downloads/3fe24c1fd9d6eae73db94a27a8d23ece/OCamlPro-Non-Commercial-License.pdf).

EDIT: Checked the licenses, we should have no issue in using this.

---

Some examples of it working (on the frontend for nice formatting)

image

image

image

image

---

I've included below a (very) rough explanation of how this works below. Hopefully it contains enough to understand the gist of how the system works but I probably missed out many details. Do let me know if there is anything that needs more elaboration.

---
# Small explanation of what's going on

Whenever a program executed in `runInContext` encounters a stack overflow or timeout (source infinite recursion/loop protection error), the infinite loop detector will be run on the code. If an infinite loop was indeed found, it will replace the error with a more descriptive error message about why the program runs into an infinite loop.

## How infinite loops are detected

The basic idea is that we run the program until it reaches some 'infinite looping state', and collect information from some iteration of this 'infinite looping state' to prove that future iterations are also 'infinite looping'. An example will probably be helpful here.

Consider the program:
```
function fac(x) {
  return x===0?1: x * fac(x-1);
}
fac(-1);
```
We want to show that when `x` is negative, the function will always take the second branch, which will result in fac being called with a negative value for `x` again.

During execution, we will take note of all the branches taken in conditionals/if statements in one iteration before `fac` calls itself again. We will call this the _Path_ of the iteration.

In the example, the path taken was `x !== 0`. Note that `x !== 0` is equivalent to saying `x<0 || x>0`, and in the example since `x` was negative, we can rewrite the path from `x !== 0` to the more specific `x < 0`. This will be helpful later.

During execution, we also want to take note of how each variable is changing. We will call this the _Transition_ of the iteration.

In the example, the transition will be `x' = x-1`. We refer to the updated `x` as `x'`.

Now we want to (automatically) check if that whenever `x<0` and `x' = x-1`, we have that `x'<0`. We do this by combining all
the information into a single sentence in first-order logic, and check the sentence's validity using an SMT solver.

In this case, the formula is `forall x, x': [(x<0 and x' = x-1) -> (x'<0)]`, which is valid.

### Other methods of detection

A special case of the above is when the path is empty, i.e. no conditional/if statement was encountered before a function calls itself again. This is when a function has no base case. If we have this we can bypass going to the SMT solver and immediately produce an error.

Another way is to detect cycles in the program's (global) state. Besides storing transitions as the symbolic `x' = x-1`, we also store the raw numerical values, i.e. `x=-1`, etc. If no variables are updated, or if there is a cycle in the global state between iterations, we have encountered an infinite loop. (This is how we detect infinite loops involving arrays as variables. It is surprisingly effective.)

## Collecting the information

In the above sections, I magically conjured up the expressions `x !== 0`, and `x - 1`. As a human observer this is obvious, but this will require some work to be done automatically. We will use _concolic execution_ (a combination of concrete and symbolic execution) to achieve this.

We want to track how each argument in the function changes/is being used. Take the following example below:
```
function f(x) {
  let y = x + 1;
  return y === 0? 0 : f(y+1);
}
f(1);
```
We want to 'track' how `x` changes in `f`. To do this we replace `x` with an object containing the concrete value of `x` and a symbolic representation of `x` (In Haskell this would be a monad). Whenever we perform binary/unary operations on `x`, we update the symbolic representation accordingly. Using the above example code:

`let y = x + 1;` -> `let y = [1, 'x'] + 1`.

Instead of substituting `x` with `1`, we substitute it with the above representation using the concrete `1` and symbolic `x` values.

`let y = [1, 'x'] + 1` -> `let y = [2, 'x+1']`

`return y===0? 0 : f(y+1);` -> `return [2, 'x+1'] === 0? 0: f(y+1)`

We check `2===0`, and take the second branch of the conditional. We save `x+1 > 0` in the Path.

-> `return f(y+1)` -> `return f([3, 'x+2'])`

Now we save `x'=x+2` in the transition. Proceed to the next iteration.

## Implementation

In practice, to implement this I decided to instrument the source code and run it using some predefined objects to handle what happens during the runtime. This is similar to what happens in the transpiler.

For example in `let y = x + 1`,

First we have to wrap `x` in a function that turns it into our hybrid object `[1,'x']`. 

`let y = fns.hybridize(x, 'x') + 1`

Then we also have to change the `+` operator into a function that accepts the hybrid objects as arguments:

`let y = fns.evalB( '+', fns.hybridize(x, 'x'), 1)`.

`fns` is a javascript object containing all the above functions (and more) defined in `runtime.ts`. It will be added to the environment of the instrumented program, which will be run in native js.

This is just a small example which (hopefully) gives the gist of what's going on in `instrument.ts` and `runtime.ts`, but in reality there are many more hacks going on that are needed to get things working.


JoeyChenSmart:
Hi @thomastanck this PR should be ready for review. Its pretty big so do let me know if theres anything you need me to explain.

src/infiniteLoops/detect.ts
```diff
@@ -0,0 +1,362 @@
...
+  const s1 = temp.slice(ix1 - period, ix1)
+  const s2 = temp.slice(ix1, -1)
+  if (s1.length != period) return undefined
+  for (let i = 0; i < period; i++) {
+    if (s1[i] != s2[i]) return undefined
+  }
+  return s1.concat(s1[0])
+}
+
+function stringifyCircular(x: any) {
```
thomastanck, line 104: I think we discussed not using any kind of stringify at all, but if we're using one, what's wrong with using the standard library's stringify? see `utils/stringify.ts`

src/infiniteLoops/detect.ts
```diff
@@ -0,0 +1,362 @@
...
+  try {
+    const input = JSON.stringify({ content: [code] })
+    const out = JSON.parse(runAltErgo(input, options))
+    return out.results[0]
+  } catch (e) {
+    return e.toString()
+  }
+}
+
+type Triple = [number[], number[], [string, any, number][]]
```
thomastanck, line 139: Could you add a comment above this line describing the fields? Best case is to use an object here (but only if it doesn't further complicate the code).

src/infiniteLoops/instrument.ts
```diff
@@ -0,0 +1,606 @@
...
+          const newName = genId(decl.id.name)
+          decl.id.name = newName
+        }
+      }
+    }
+    for (const stmt of stmts) {
+      callback(stmt, undefined)
+    }
+  }
+  recursive(program, [{}], {
```
thomastanck, line 82: Just checking, have you tested renaming with if statements, which can also introduce new envs? My guess is that it's not a problem due to to the BlockStatement introducing a new env, but it's probably prudent to check it in a test anyway.

src/infiniteLoops/runtime.ts
```diff
@@ -0,0 +1,295 @@
...
+    } else {
+      newBuiltins.set(name, (...args: any[]) => fun(...args.map(sym.shallowConcretize)))
+    }
+  }
+  newBuiltins.set('undefined', undefined)
+  return newBuiltins
+}
+
+function nothingFunction(...args: any[]) {
+  return nothingFunction
```
thomastanck, line 230: Why do you return `nothingFunction` instead of, say, `args[0]` or `undefined`? Returning `args[0]` would be more in line with the type of `display`.

thomastanck (approved):
Thanks for formatting the tests. Just had a couple more comments I found during code reading, and it should be good to merge after those are fixed!

Really appreciate the work you put in to integrating the SMT solver, I'm sure it was a little tricky...

src/index.ts
```diff
@@ -468,23 +472,22 @@ export async function runInContext(
...
-    addInfiniteLoopProtection(program, context.chapter === 2)
-  }
   const isNativeRunnable = determineExecutionMethod(theOptions, context, program)
-  if (context.prelude !== null) {
-    const prelude = context.prelude
-    context.prelude = null
-    await runInContext(prelude, context, { ...options, isPrelude: true })
-    return runInContext(code, context, options)
-  }
+  if (context.nativeStorage.evaller === null || context.numberOfOuterEnvironments)
```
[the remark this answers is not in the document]
JoeyChenSmart: Seems like I had some code after line 476 that I deleted but I forgot to delete line 476 along with it.
Line 476 doesn't do anything (the next if statement tests for the same thing) so it didn't break anything but I removed it in the new PR just to be safe.

src/infiniteLoops/runtime.ts
JoeyChenSmart: The `nothingFunction` is used as a stub when some library functions (e.g. from RUNES etc) are used in the code. I didn't want the return value of these functions to be used at all (e.g. `x_of(point) + 1`) and just throw an error when used.

src/infiniteLoops/detect.ts
JoeyChenSmart: Didn't know there was already a stringify implementation... In any case I'll just leave this bit in for now until https://github.com/source-academy/js-slang/issues/1074 is resolved.

PR #1056, opened 2021-07-16
Source non-det: rename command try_again to retry

martin-henz:


martin-henz:
fixes https://github.com/source-academy/sicp/issues/569

PR #1057, opened 2021-07-20
Implement rest/spread

### LINKED ISSUE #1049: Rest and spread for Source §3
Source follows the need of SICP JS. SICP JS mentions rest and spread:

https://sourceacademy.nus.edu.sg/sicpjs/3.4.2#h4

Therefore Source §3 should include rest and spread syntax.

This feature would be crucial if we were to adapt
Software Design for Flexibility
https://mitpress.mit.edu/books/software-design-flexibility
to JavaScript

openorclose:
fixes #1049 

should we ban `[...a]`? currently not yet implemented so this would result in an error in the interpreter


martin-henz (approved):
Tested and documented

PR #1070, opened 2021-07-26
Move SMT solver to external package

JoeyChenSmart:
Refer to the PR here: https://github.com/source-academy/frontend/pull/1869

It seems the local packages don't work properly when published. Fixed this by publishing the SMT solver separately and then importing that instead.


JoeyChenSmart:
Oops was rushing and forgot to rebase. Ignore all but the last commit

martin-henz (approved):
ok

PR #1098, opened 2021-08-13
Fix import statements breaking infinite loop detector

JoeyChenSmart:
With the new change where students manually import libraries in their code, wild `import` statements in the program will break the infinite loop detector (in particular when we use `eval` on the instrumented code).

e.g.
```
import {turn_upside_down,beside,heart,stackn,show} from 'rune';
function besiden(n, rune) {
  // edit the return expression
    return besiden(n,rune);
  }
  
show(besiden(5, heart));
```
This will currently not be detected unless you remove the first line.

EDIT:
Fixed this by actually importing the module into the code.
One hiccup is that modules like `sound` might try to play sound in the browser's `window` so to prevent this we shadow the original window with an empty object (before eval), i.e. `{let window = {}; return eval(code)}`

### DIFF
--- a/src/infiniteLoops/runtime.ts
+++ b/src/infiniteLoops/runtime.ts
@@ -12,6 +12,7 @@ import {
 } from './instrument'
 import { parse } from '../parser/parser'
 import { createContext } from '../index'
+import { MODULE_PARAMS_ID } from '../constants'
 
 function checkTimeout(state: st.State) {
   if (state.hasTimedOut()) {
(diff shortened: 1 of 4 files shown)


src/infiniteLoops/runtime.ts
```diff
@@ -279,10 +280,18 @@ export function testForInfiniteLoop(code: string, previousCodeStack: string[]) {
...
 
-  const sandboxedRun = new Function('code', functionsId, stateId, builtinsId, 'return eval(code)')
+  const sandboxedRun = new Function(
+    'code',
+    functionsId,
+    stateId,
+    builtinsId,
+    MODULE_PARAMS_ID,
+    // redeclare window so modules don't do anything funny like play sounds
+    '{let window = {}; return eval(code)}'
```
thomastanck, line 290: Have you tested this in browser? I'm not convinced it'll work, actually. A permanent fix would probably need a change in the modules side.

It's good enough for now, though. If something is timing out, their code will probably be stuck before any side effects occur.

thomastanck (approved):

src/infiniteLoops/runtime.ts
JoeyChenSmart: yup I tested in the browser

I ran something like
```import {play, sine_sound} from 'sound';
play(sine_sound(1200,1));
function f(x){f(x);}
f(1);
```
and the sound only played once. The infinite loop wasn't detected though since `play` threw an error first (in the instrumented code)

thomastanck: Hmm on closer inspection, it might actually work, I misread some parts of the module loading process.

Ignore this, then.

PR #1099, opened 2021-08-17
Make typechecking lazy

thomastanck:
Should speed up Source 2 to Source 4 by approx 100ms (~150ms to ~25ms).

Test with tiny programs like

`1;`


thomastanck:
Note that typecheck errors still work as usual, they're invoked in a different path in the frontend.

openorclose (approved):

PR #1100, opened 2021-08-18
Disable math functions in infinite loop detector

JoeyChenSmart:
The infinite loop detector interprets the return value of math functions (and in general builtin functions that return numbers) as constants which leads to false positives/incorrect error messages, like in the following code:
```
function cc(n) {
    return n===0
        ? "hi"
        : n % 2 === 0
        ? cc(math_floor(n/2))
        : cc(n*3+1);
}

cc(99);
```

Fixed this by marking the output of math (and other numeric builtins) as 'invalid' symbolically so the infinite loop detector does not use that iteration in its analysis in the SMT solver (no base case and cycle detection still work).

Also marked output from functions such as `math_random` as non-deterministic and exclude functions using these values from cycle detection

Also refactored tuple data structures to objects for better readability

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-slang",
-  "version": "0.5.14",
+  "version": "0.5.15",
   "license": "Apache-2.0",
   "description": "Javascript-based implementations of Source, written in Typescript",
   "keywords": [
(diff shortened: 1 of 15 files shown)


martin-henz (approved):
Yes, inevitable to have undetected cases, in absence of including the primitive operations in the reasoning of the prover.

PR #1123, opened 2021-09-05
Stepper: Fix redex substitution

### LINKED ISSUE #1112: Stepper: Predicate repeated
Screenshot 2021-08-30 at 3 00 13 PM

Note that this example uses the predeclared function length.

petermonky:
Should fix #1112.

Stepper now declares the detects `$redex` as the redex marker. This way the marker doesn't collide with the `$` character in the prelude functions as before.

Stepper redex fix

### DIFF
--- a/src/stepper/stepper.ts
+++ b/src/stepper/stepper.ts
@@ -2453,8 +2453,8 @@ function pathifyMain(
   let path = paths[0]
   let redex = ast.program([]) as substituterNodes
   let endIndex = path === undefined ? 0 : path.length - 1
-  const redexMarker = ast.identifier('$') as substituterNodes
-  const withBrackets = ast.identifier('($)') as substituterNodes
+  const redexMarker = ast.identifier('$redex') as substituterNodes
+  const withBrackets = ast.identifier('($redex)') as substituterNodes
 
   const pathifiers = {
     ExpressionStatement: (target: es.ExpressionStatement): es.ExpressionStatement => {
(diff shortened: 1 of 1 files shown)


martin-henz (changes requested):
It may be best to use a syntax that cannot be used in JavaScript programs. You could use '@redex' instead of '$redex'. Otherwise you'll have the same problem when the name $redex occurs in a user program.

petermonky:
> It may be best to use a syntax that cannot be used in JavaScript programs. You could use '@redex' instead of '$redex'. Otherwise you'll have the same problem when the name $redex occurs in a user program.

Just pushed a new change for this; the redex marker should now be `@redex`.

src/stepper/stepper.ts
```diff
@@ -2877,6 +2877,9 @@ export const getRedex = (node: substituterNodes, path: string[][]): substituterN
 // then we return it to the getEvaluationSteps
 function substPredefinedFns(program: es.Program, context: Context): [es.Program, Context] {
   if (context.prelude) {
+    // replace all occurences of '$' with 'helper_' to
+    // prevent collision with redex (temporary solution)
+    // context.prelude = context.prelude.replace(/\$/gi, 'helper_')
```
thomastanck, line 2882: is this comment stale?

petermonky: ohh yea, that comment can be removed.

PR #1152, opened 2021-10-25
Implement Typescript-esque predicate tests

thomastanck:
Depends on #1151

Caveats: lack of union types means that we don't quite follow the same
semantics as typescript, notably when it comes to disjunctions of
predicate tests as well as the lack of constraints applied to the
alternate of a predicate test conditional

Example 1:

```
if (is_pair(xs) || is_null(xs)) {
  // no constraint added to xs here
  // Typescript's bidirectional type system would infer
  //   xs: Pair<T1, T2> | null
}
```

Example 2:

```
if (is_pair(xs) || is_null(xs)) {
  if (is_pair(xs)) {
    // xs known to be pair here
  } else {
    // no constraint added to xs here
    // Typescript's bidirectional type system would infer
    //   xs: null
  }
}
```

This PR contains:

* Implementation of predicate tests as well as documentation updates for type inference.
* Tweaked implementation of some list prelude functions (e.g. equal and remove) so that they now have correct types.
  * In this version of js-slang, the entire list prelude finally has correct types!
* Changed typing of `&&` and `||` during type checking. Now they require both inputs to be `boolean`, in order to be consistent with conditionals. The "human" type given in the Source chapter 1-4 specifications have not been changed.
* Moved `is_list` from being a predeclared function to a builtin, which allows it to have a predicate type.


martin-henz (approved):
Looks good. Looked over the changes and they make sense to me. Trusting the test cases.

thomastanck:
Will merge after fixing the merge conflict and removing some commented out code.

PR #1170, opened 2021-11-16
Implement arity

### LINKED ISSUE #1160: add function arity to Source §1
arity(f): returns the number of arguments that the function f takes. If applied to non-function, an error is issued.

teekaytai:
This will resolve #1160.
Could not run the test cases for arity though. If I'm not mistaken some gitignored Source .json files are needed?


martin-henz:
About testing, you say:
Could not run the test cases for arity though. If I'm not mistaken some gitignored Source .json files are needed?

Not sure if you understand the jest setup. Can you take a look at this:

https://github.com/source-academy/js-slang#testing

martin-henz:
Currently the PR is failing CI (continuous integration) aparently because of linting. Try
yarn format

teekaytai:
To clarify about running the tests, when I enter `yarn run test`, some tests still run.
image

But some of them fail to run, seemingly because some json files are missing:

image

Also, I corrected the formatting failing the check.

thomastanck:
Run

yarn # get packages
yarn build # generates json files
yarn test -u # tests and updates snapshots

teekaytai:
Thanks Thomas! It works now. Sorry for the trouble, I'm new to yarn and jest.

Regarding the tests, I ran into an issue where some tests failed because the "native" and "interpreted" results of a program were different. Not sure how to proceed now, could someone explain what that means? Thanks.

image

martin-henz:
"interpreted" is the variant of the Source Academy system that is used for the environment visualizer. It's a bit like our MCE, so functions are represented by some data structures. The arity function needs to be adapted so that it distinguishes between the two cases.

teekaytai:
Thanks for the hint prof! Everything should be fine now.

martin-henz:
Ah it would be great if we could get arity to work in embedded versions, such as our robots. For that, we'd need to modify this folder: https://github.com/source-academy/js-slang/tree/master/src/vm
Alternatively, we can file an issue about this gap. Up to you how much time you want to spend on this.

Yes, Source Academy has become quite complex, with so many runtime options to consider. But arity was important, because otherwise is_stream is really weird.

teekaytai:
Perhaps an issue can be filed for now, I've got to worry about finals 😬

martin-henz (approved):
Looks good!

PR #1185, opened 2021-12-13
Documentation: display and error

### LINKED ISSUE #733: Missing run-time check for correct number of arguments for display() and error()
Both the following calls should be an error according to the spec, but are currently not an error:

```
display();
display(1,"T",3);
```

It is a problem (and perhaps confusing for students) because:
- It does not follow the spec
- It hints that missing arguments turn into `undefined`, which is not taught anywhere and not mentioned anywhere in the spec
- Source-defined functions and other stdlib functions are checked as per #346, so not checking display() and error() is even more surprising

martin-henz:
to allow (and ignore) arguments beyond the second; fixes #733
by declaring that the bug is a feature.

### DIFF
--- a/docs/lib/misc.js
+++ b/docs/lib/misc.js
@@ -101,7 +101,7 @@ function prompt(s) {
 /**
  * Optional second argument. If present,
  * displays the given string <CODE>s</CODE>, followed by a space character, followed by the
- * value <CODE>v</CODE> in the console.
+ * value <CODE>v</CODE> in the console. Arguments after the second argument are ignored.
  * If second argument not present,
  * just displays the value <CODE>v</CODE> in the console.
  * The notation used for the display of values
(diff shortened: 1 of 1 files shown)


martin-henz:
@btzy what do you think about this solution? This way, we can treat variadic functions in a more uniform and less complicated manner across the different implementations.

btzy:
Well, imho this is just a hack, and a good linter would then be expected to emit warnings when `display()` or `error()` is called with more than two arguments.

martin-henz:
I agree it's a hack. We should do this properly, and implement the two functions correctly, across the implementation of Source. 

(I'm just looking for opportunities to simplify things. Concerned about long-term sustainability of the project. All depends whether we can motivate enough talented people to join and contribute. Hopefully, with SICP JS coming out in 2022, we get more eyeballs.)

angelsl:
From the Sinter point of view: Sinter already confirms to this behaviour. I do think there is value in reporting extraneous arguments though, rather than silently discarding them.

martin-henz:
OK. Let's try to get all implementations properly report any number of arguments of display and error less than 1 and greater than 2.

openorclose:
Should we separate these into 2 functions instead, a version that only accepts 1 argument and another that only accepts 2?

martin-henz:
@openorclose said:
> Should we separate these into 2 functions instead, a version that 
> only accepts 1 argument and another that only accepts 2?

That gate has closed yesterday: We squeezed a final bug fix into the PDF of SICP JS for printing by MIT Press. That's it now: Any other changes need to go into an errata sheet and can only be corrected in the next printing.

The bottom line: The design of Source §1-4 is practically frozen now. Most features are used / mentioned in the book, and we cannot chance them any longer. Changing display and error would confuse readers and disturb our community of learners unnecessarily.

PR #1193, opened 2021-12-29
typeChecker: import should bind local name, not imported name

### LINKED ISSUE #1191: "misconfiguration bug": type checking
https://share.sourceacademy.org/5ng75

leads to:

Line 14: Expected 2 arguments, but got 1.
Hints:
Line 11: One or more undeclared names detected (e.g. 'sound_play').
If there aren't actually any undeclared names, then is either a Source or misconfiguration bug.
Please report this to the administrators!

thomastanck:
Fixes #1191 by creating a binding for the local field and not the imported field.

When an import doesn't rename the imported object, e.g. `import { show } from 'runes';`, the `imported` and `local` fields are identical, which is why the bug wasn't caught sooner. In `import { show as tell } from 'runes';`, show is the imported field while tell is the local field.

### DIFF
--- a/src/typeChecker/__tests__/typeChecker.test.ts
+++ b/src/typeChecker/__tests__/typeChecker.test.ts
@@ -2069,21 +2069,23 @@ describe('Type context from previous executions get saved', () => {
 describe('imported vars have any type', () => {
   it('import', () => {
     const code1 = `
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Looks good!

PR #1201, opened 2022-01-06
`stringify` non-literal object support

shenyih0ng:
Fixes: https://github.com/source-academy/frontend/issues/2056


martin-henz (approved):
nice catch

PR #1207, opened 2022-01-31
Fix for reduction of empty block expressions

### LINKED ISSUE #1109: Stepper: stops at applying empty function body
image

This program is able to run normally.

However, the stepper stops at trying to run `a()` where the function has an empty body, displaying an empty output message on the right:

image

The stepper also stops when using such a lambda:

image

Explicitly returning `undefined` seems to let it proceed:

image

However, this may be a thing different from #1108, despite that also depicting empty function bodies and `undefined`. Since it seems to be about the stepper understanding empty function bodies, rather than about the stepper displaying `undefined` when ending on some *declaration*. Eg, the function run (*application*?) here doesn't need to return `undefined` as well, it can do something else like return `1` and proceed, as long as its body is not empty:

image

The empty output message is reminiscent of how the stepper acts when a program actually encounters an error:

image

Submitted with reference to my [Piazza post](https://piazza.com/class/krvh5l6ipib380?cid=115_f1). I'm not completely certain if this isn't just another effect stemming from #1108, where said issue spawned from [the same thread](https://piazza.com/class/krvh5l6ipib380?cid=115).

parnikkapore:
Fixes #1109

Follows suggested fix by prof Henz; it also turns out that reduction of block expressions is missing the empty block handling that's present in the block statement code, so I copied that over too

### DIFF
--- a/src/stepper/__tests__/__snapshots__/stepper.ts.snap
+++ b/src/stepper/__tests__/__snapshots__/stepper.ts.snap
@@ -1,5 +1,91 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`#1109: Empty function bodies don't break execution Constant declaration of lambda 1`] = `
+"const a = () => {};
+\\"other statement\\";
+a();
+\\"Gets returned by normal run\\";
+
+const a = () => {};
+\\"other statement\\";
+a();
+\\"Gets returned by normal run\\";
+
+\\"other statement\\";
+(() => {})();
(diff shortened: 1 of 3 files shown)


martin-henz:
@parnikkapore Continuous Integration (CI) is failing for this PR. The reason is that it doesn't pass the linting tools. You should run 
yarn format
before you commit and create the PR. For this PR, you can run
yarn format
and then push the changes to the same branch.

martin-henz (approved):
Looks good! Nice work!

PR #1216, opened 2022-02-05
Enforce auto pre-commit/pre-push formatting and testing via Husky

salty-flower:
## Description
Husky is already added in #167, but not automatically installed to the working directory. It is a bit annoying to find commits failed the tests with a red cross only due to format issues after pushing to GitHub.

## Change
I added one line to `package.json` under `scripts` and Husky seems to work fine.
https://github.com/source-academy/js-slang/blob/659ecd8d0d2f0f5da4b36a7b77c112089e0333ab/package.json#L53
Also, `eslint-plugin-simple-import-sort` is added as a dependency to enforce import sorting, since this repository's Action warns this.
https://github.com/source-academy/js-slang/blob/659ecd8d0d2f0f5da4b36a7b77c112089e0333ab/.eslintrc.json#L46
This PR also formats all existing code with respect to current rules.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,7 +14,10 @@
     "project": "tsconfig.json",
     "sourceType": "module"
   },
-  "plugins": ["@typescript-eslint"],
+  "plugins": [
+    "@typescript-eslint",
+    "simple-import-sort"
+  ],
   "rules": {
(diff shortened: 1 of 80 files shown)


martin-henz (approved):
Great idea. Thanks!

Issue #1097, opened 2021-08-13
Multiple import statements result in multiple tabs generated


chownces:
This bug was discovered by a student. See https://share.sourceacademy.nus.edu.sg/tworunes. There are also 2 runes generated instead of 1.

leeyi45:
# Proposed solution
## Outcome
- This solution should cause modules' tabs to be generated only once regardless of the number of times it is imported.

Add an array containing the names of modules previously imported to the evaluation context.
```ts
export interface Context<T = any> {
  /**
   * Names of modules that have already been imported
   */
  importedModules?: string[]
}
```

When `appendModulesToContext` is called to add the module's tabs, check if they have already been previously added.

```ts
function appendModulesToContext(program: Program, context: Context): void {
  if (context.modules == null) { 
    context.modules = []
    context.importedModules = []
  }

  for (const node of program.body) {
    if (node.type !== 'ImportDeclaration') break
    const moduleName = (node.source.value as string).trim()

    // Check if the module's content has already been imported before
    // Avoid loading its tabs again if it has
    if(!context.importedModules?.includes(moduleName)) { 
        context.importedModules?.push(moduleName);
        Array.prototype.push.apply(context.modules, loadModuleTabs(moduleName))
    }
  }
}
```

This prevents the module's tabs from being added to the context multiple times.

Issue #29, opened 2018-08-15
Rest parameter syntax not caught as SyntaxError


ningyuansg:
```
function f(x, ...rest){
    return rest;
}
```

should raise a syntax error, but does not.

martin-henz:
Haha, I beat you to this!

ningyuansg:
oops

ningyuansg:
Closing in favour of #28

PR #1220, opened 2022-02-25
Introduction of Modules Context objects

leeyi45:
This branch changes how modules are imported by js-slang.

### Modules State
This branch introduces the module state interface for modules to store their own state information that is accessible wherever `Context` is accessible
```.ts
interface ModuleState {}
```
Example:
```.ts
class CurvesModuleState implements ModuleState {
    public drawnCurves: ShapeDrawn[];
}
```

### Module Context
`ModuleState` objects are stored in `ModuleContext`s, which also store the module's tabs.
```.ts
type ModuleContext = {
    state?: ModuleState;
    tabs: any[]
}
```
### Context
`ModuleContext`s are stored in the `Context` object in a map, the key of the map being the name of the module (e.g. `curve`). This map is passed to modules during initalization, making the context of other modules available to the module being imported. The previous `moduleParams` object is still passed to modules.
```.ts
type Context <T = any> = {
    moduleContexts: Map<string, ModuleContext>
    moduleParams: any
}
```

To make use of the new context object, modules should take two parameters:
```.ts
export function curves(params, contexts: Map<string, ModuleContext>) {
    const moduleContext = contexts.get('curve');
    moduleContext.state = {
        drawnCurves: []
    };
}
```

### Issues Resolved
- Fixes [Modules Issue #65](https://github.com/source-academy/modules/issues/65) - Drawing should not depend on evaluation result
- Fixes [Modules Issue #118](https://github.com/source-academy/modules/issues/118) - Importing pixnflix and sound crashes the frontend
    - Also fixes a possibly related issue where importing the curves and sound modules crashes the frontend
- Fixes [Modules Issue #29](https://github.com/source-academy/modules/issues/29) - Module errors not caught in interpreter mode 
    - The interpreter will now also properly load side content tabs
- Fixes [js-slang Issue #1097](https://github.com/source-academy/js-slang/issues/1097) - Multiple import statements cause multiple tabs to be spawned
- Import statements will now be hoisted to the top of the source program to follow Javascript's behaviour, and multiple import statements to the same module will be treated as a single import statement.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -8,6 +8,7 @@ export const TRY_AGAIN = 'retry' // command for Source 4.3
 export const GLOBAL = typeof window === 'undefined' ? global : window
 export const NATIVE_STORAGE_ID = 'nativeStorage'
 export const MODULE_PARAMS_ID = 'moduleParams'
+export const MODULE_STATE_ID = 'moduleContext'
 export const MAX_LIST_DISPLAY_LENGTH = 100
 export const UNKNOWN_LOCATION: es.SourceLocation = {
   start: {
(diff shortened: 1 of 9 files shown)


martin-henz (approved):
Nice work!

PR #1231, opened 2022-03-12
Source Native Support

shenyih0ng:
## Description

Added `Native` variant to all existing Source chapters

### `Native` variant

- Adheres to respective Source chapter parser and validator rules
- **Does not** use Source transpiler rules instead make use of `fullJS` transpiler rules to handle modules
- Tools such as the **stepper, env viz and data viz** will work as expected

> To put it simply, it runs respective `Source` parsers and hands it over to `fullJSRunner()`

## TODOs
- [X] Source Native Unit Tests


salty-flower (commented):
Looks good to me!

martin-henz (approved):
Looks fine!

PR #1232, opened 2022-03-17
Fix inconsistency of json imports by brute force

thomastanck:
There seems to be an inconsistency between how jest and how typescript
behaves when encountering imports of the form `import * as x from 'x.json'`
jest will set x = jsonobject,
but typescript will instead set x = { default: jsonobject }

This means that under typescript, we want `import x from 'x.json'`,
while under jest, we want `import * as x from 'x.json'`

This problem was hidden when transpiling to CommonJS modules before, which
behaves similarly to jest. But now that we are transpiling to es6,
typescript projects that depend on js-slang may now be exposed to this
inconsistency.

For now, we use brute force until the landscape changes or someone thinks of
a proper solution.

Resolves https://github.com/source-academy/frontend/issues/2123

### DIFF
--- a/src/editors/ace/docTooltip/index.ts
+++ b/src/editors/ace/docTooltip/index.ts
@@ -6,16 +6,44 @@ import * as source_3_concurrent from './source_3_concurrent.json'
 import * as source_3_non_det from './source_3_non-det.json'
 import * as source_4 from './source_4.json'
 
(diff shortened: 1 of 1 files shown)


martin-henz (approved):
Nice explanation. This will help in the future.

PR #1233, opened 2022-03-18
Fix `ChromeEvalErrorLocator` regex

shenyih0ng:
Fixes https://github.com/source-academy/frontend/issues/2124

## Description

The existing regex used to extract error messages fails to get the correct error for the following snippet:

```javascript
function h() {
    g();
}
h();
```
Expected: **Line 2: Name g not declared.**
Return: **Line 4: Name h not declared.**

The error stack is as follows:

```
at g (eval at nativeStorage.evaller (eval at fullJSEval (http://localhost:8000/static/js/main.chunk.js:14490:12)), <anonymous>:2:3)
at eval (eval at nativeStorage.evaller (eval at fullJSEval (http://localhost:8000/static/js/main.chunk.js:14490:12)), <anonymous>:4:1)
```

The problem lies in the fact that the current regex matches the prefix `at eval` which only captures errors that occur in the first stack frame of the evaluation.

Therefore, the fix will instead match error messages during the entire `eval` process and return the first error on the top of the error stack.

```diff
- `at eval.+<anonymous>:(\d+):(\d+)`
+ `eval at.+<anonymous>:(\d+):(\d+)`
```

### DIFF
--- a/src/runner/__tests__/fullJSRunner.ts
+++ b/src/runner/__tests__/fullJSRunner.ts
@@ -124,7 +124,19 @@ describe('Error locations are handled well in fullJS', () => {
     ]
   }
 
-  it.each([UNDEFINED_VARIABLE, SYNTAX_ERROR])(
+  const REFERENCE_ERROR: CodeSnippetTestCase = {
+    name: 'REFERENCE ERROR',
+    snippet: `
+            function h() {
+              g();
+            }
+            h();
+            `,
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Great. Looks like this will do it. I'm leaving the issue in the frontend repo so that we remember to check (in Chrome and Firefox) once js-slang is bumped and merged.

PR #1243, opened 2022-04-11
Allow js-slang to load tabs that require ReactDOM

leeyi45:
Modify `convertRawTabToFunction` to handle tabs that require `React` to be passed to them, and also both `React` and `ReactDOM`. This will allow tabs to make use of components like Tooltip2 from blueprint/js.

Refer to [this](https://github.com/source-academy/modules/pull/138) pull request for examples of tabs that utilize such components.


martin-henz (approved):
Looks good.

PR #1244, opened 2022-04-12
Lodash Memoize Fix

leeyi45:
The [`memoize`](https://lodash.com/docs/#memoize) function provided by lodash only memoizes based on the first parameter of the function. This occasionally causes incorrect values to be passed to `loadModuleTabs` and `loadModuleBundle`, creating errors when evaluating tabs or bundles.

```ts
const memoizedGetModuleFileInternal = memoize(getModuleFile)
export const memoizedGetModuleFile = (name: string, type: 'tab' | 'bundle') =>
  memoizedGetModuleFileInternal({ name, type })

// Make getModuleFile a single parameter function instead
function getModuleFile({ name, type }: { name: string; type: 'tab' | 'bundle' }): string {
  return httpGet(`${MODULES_STATIC_URL}/${type}s/${name}.js`)
}
```
The above changes should fix this issue from occurring in the future.

### DIFF
--- a/src/modules/moduleLoader.ts
+++ b/src/modules/moduleLoader.ts
@@ -55,8 +55,9 @@ function getModuleManifest(): Modules {
  */
 
 const memoizedGetModuleFileInternal = memoize(getModuleFile)
-export const memoizedGetModuleFile = (name: string, type: 'tab' | 'bundle') => memoizedGetModuleFileInternal({name, type})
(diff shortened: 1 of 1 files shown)


leeyi45:
I encountered this issue while expanding upon `loadModuleTabs` and `loadModuleBundle`, where calls to the latter were returning me the tabs. Not very sure why this issue hasn't really been noticed before.

PR #1245, opened 2022-04-16
Fix sicp package build

algtan:
I would run into a "MODULE_NOT_FOUND" error when trying to use 'sicp' per the instructions on npm: https://www.npmjs.com/package/sicp

I rebuilt the 'sicp' package with some changes, and now it seems to work when I use it in Node (my repo for 'sicp': https://github.com/algtan/sicp-algtan)

### Changes
`build_sicp_package.sh`:
It seems like items were missing when trying to import 'sicp' when working in Node, so I updated the script to include all dependencies.

`.eslintignore`:
I would run into problems running the `build_sicp_package` script because the linter would find problems in the /dist/ folder, so added this file.

`sicp_publish/package.json` and `sicp_publish/yarn.lock`:
There were missing dependencies for `@joeychenofficial/alt-ergo-modified` and `source-map`.

`src/createContext.ts`:
Kept running into a ESLint error with this file, so changed the problem type to 'string'.

`tsconfig.json`:
Updated the module to `commonjs` to match how 'sicp' was packaged in npm. When building the package with module set to 'es2015', I would run into issues with the 'import' syntax.


martin-henz (approved):
Looks good!

PR #1283, opened 2022-08-25
Fix passing of module params

gok99:
The game module, which relies on module parameters, was receiving `IOptions` instead of the module params passed from frontend. This seems to be because of a `sandboxedEval` call in `sourceRunner.ts`. Changing the argument seems to fix the issue and allows the game module to function correctly.

There were also 2 calls to `fullJSEval` where options are passed for the moduleParams argument. Changing these resulted in the `native.ts` test non-deterministically getting stuck and not being able to complete (running the test individually does not seem to cause the same issue however). I'm not sure if this change needs to be made.


martin-henz:
Just a heads-up: This will require bumping js-slang in frontend. Once this PR is approved, I'll do the bumping. Then this should work in sourceacademy.org. For sourceacademy.nus.edu.sg, we'll need to redeploy the frontend in AWS.

martin-henz (approved):
Looks good!

PR #1285, opened 2022-08-26
Refactor full JS language representation

zhaojj2209:
As HTML will soon be added SA, after discussion, it was agreed to have languages that are not part of Source (e.g. full JS, HTML) share the chapter number (-1), and differ in variant. This PR is to make the necessary refactoring first before adding HTML support.

### DIFF
--- a/src/runner/utils.ts
+++ b/src/runner/utils.ts
@@ -10,8 +10,12 @@ import { simple } from '../utils/walkers'
 
 // Context Utils
 
-export function isFullJSChapter(chapter: number): boolean {
-  return chapter == -1
+export const isOtherLanguage = (chapter: number) => {
+  return chapter === -1
+}
+
+export const isFullJSLanguage = (chapter: number, variant: Variant) => {
+  return isOtherLanguage(chapter) && variant === 'js'
 }
 
 /**
--- a/src/types.ts
+++ b/src/types.ts
@@ -59,8 +59,19 @@ export interface Comment {
 }
 
 export type ExecutionMethod = 'native' | 'interpreter' | 'auto'
-export type Variant = 'native' | 'wasm' | 'lazy' | 'non-det' | 'concurrent' | 'gpu' | 'default' // this might replace EvaluationMethod
 
+// Includes languages that are not part of Source (e.g. full JS)
+export type Variant =
+  | 'native'
+  | 'wasm'
+  | 'lazy'
+  | 'non-det'
+  | 'concurrent'
+  | 'gpu'
+  | 'default'
+  | 'js'
+
+// Languages that are not part of Source have chapter set to -1
 export interface SourceLanguage {
   chapter: number
   variant: Variant
--- a/src/createContext.ts
+++ b/src/createContext.ts
@@ -2,7 +2,7 @@
 
 import { GLOBAL, JSSLANG_PROPERTIES } from './constants'
 import * as gpu_lib from './gpu/lib'
-import { isFullJSChapter } from './runner'
+import { isFullJSLanguage } from './runner'
 import { AsyncScheduler } from './schedulers'
 import { lazyListPrelude } from './stdlib/lazyList.prelude'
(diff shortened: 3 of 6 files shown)


src/runner/utils.ts
```diff
@@ -10,8 +10,12 @@ import { simple } from '../utils/walkers'
 
 // Context Utils
 
-export function isFullJSChapter(chapter: number): boolean {
-  return chapter == -1
+export const isOtherLanguage = (chapter: number) => {
```
shenyih0ng, line 13: Perhaps it will be clearer if we have a `isSource`/`isSourceLanguage` instead? My concern is that `OtherLanguage` might be a little vague

src/types.ts
```diff
@@ -59,8 +59,19 @@ export interface Comment {
...
+// Includes languages that are not part of Source (e.g. full JS)
+export type Variant =
+  | 'native'
+  | 'wasm'
+  | 'lazy'
+  | 'non-det'
+  | 'concurrent'
+  | 'gpu'
+  | 'default'
+  | 'js'
```
shenyih0ng, line 72: I am not quite sure if `js` should be treated a `Variant`. My intuition is that we have a language (`source`, `full JS`, `HTML`) and each language has zero or more variants.

Perhaps there will be a need for us to have a explicit `Language` construct/type to better handle more languages. Also, right now `chapter` seems like is a purely `source`/`sicp` concept, might not be the best way to represent different languages.

Let me know what do you think of this?

zhaojj2209: I have considered this as well! I have thought of the following methods to refactor the code:
- Keep full JS and HTML as different languages with negative chapter numbers (full JS = chapter -1, HTML = chapter -2)
- Group full JS and HTML as variants of chapter -1 (current implementation)
- Add an `OtherLanguage` type for full JS/HTML

@martin-henz Would like your input on which method would be best!

Side note, we could also refactor `Variant` into an enum in the future.

martin-henz: I think the problem lies in the numeric indexing at the top level. How about we index using strings instead of numbers -1, 0, 1, 2, 3, 4 ?
We could have top-level strings like:
"Source §1", "Source §2", "Source §3", "Source §4", "HTML", "full JavaScript", "TSX", etc
For each of those we could have optional variants, also indexed using strings.

What do you all think about this?

zhaojj2209: I think that is a good idea! I think I can go one step further and refactor both the top-level languages and variants to make use of string enums (e.g. `Language`, `Variant`). If everyone is alright with this I will make a new PR for the refactoring and close this one.

shenyih0ng: > I think the problem lies in the numeric indexing at the top level. How about we index using strings instead of numbers -1, 0, 1, 2, 3, 4 ? We could have top-level strings like: "Source §1", "Source §2", "Source §3", "Source §4", "HTML", "full JavaScript", "TSX", etc For each of those we could have optional variants, also indexed using strings.
> 
> What do you all think about this?

this sounds good to me! we will perhaps still need keep the idea of chapters and numeric indexing for SA

zhaojj2209: After digging deeper into the codebase, I came across the [following code](https://github.com/source-academy/js-slang/blob/e5647b5b5a1e37f5f74c368c2d358ad088f605d9/src/createContext.ts#L378) which checks for `chapter >=100`. Unless this block of code is legacy and can be safely removed, this makes it difficult for the chapters to be anything other than numeric.

Perhaps as a workaround I can define a `Chapter` enum with numeric indexing? Though it means that we will still have additional languages be represented by negative chapter numbers, the mapping will at least make it clearer which language corresponds to which number.

martin-henz: This was a hack that I added to support OOP in js-slang, which I needed for CS4215 last time. We should come up with a better way. So yea, you can consider "chapter >= 100" as "legacy".

zhaojj2209:
Closing as implementation has been replaced by #1286.

PR #1289, opened 2022-09-06
Improvements to the Module System 2

### LINKED ISSUE #1246: Transpiled names may conflict with user declared names when using transpiled Source
The following Source program should run without errors.
```ts
import { twice } from 'repeat'; // any module will work

twice(x => x);
const __MODULE_0__ = "random_str";
```

Because user variables are transpiled as is, there is potential for naming conflicts with variables declared during transpilation. This results in errors such as `ReferenceError: Cannot access '__MODULE_0__' before initialization` In this case, the name `__MODULE_0__` is used by the transpiler to prefix the `repeat` module.
To resolve this issue, user variables should be appropriately namespaced to prevent naming conflicts from being possible during transpilation.

leeyi45:
In conjunction with this [PR](https://github.com/source-academy/modules/pull/153), change how bundles are loaded, making the evaluation context available to bundles.

This PR should also fix #1246.


martin-henz (approved):
Looks good. This PR is needed for a major upgrade of the module system. See
https://github.com/source-academy/modules/pull/153 and
https://github.com/source-academy/frontend/pull/2208

PR #1290, opened 2022-09-07
Fix bug in stringifying objects

zhaojj2209:
Fixed inconsistent naming of function arguments in `convertObject` which causes objects to be wrongly memoized when the `stringify` util function is called, resulting in code related to objects to be wrongly displayed on the frontend.

Before fix:
image

After fix:
image


martin-henz (approved):
That was fast!

PR #1296, opened 2022-09-13
Fix `build` command

### LINKED ISSUE #1264: yarn build does not build the dist directory
After I deleted the dist directory manually, running `yarn build` does not build the dist directory again. The README is also not clear on what to do.

shenyih0ng:
Fixes #1264 

## Problem

When the `dist` directory is deleted or removed after a initial build via `yarn build`, subsequent builds will not rebuild the `dist` directory.

This is due to the presence of `tsconfig.tsbuildinfo` which is generated when the `incremental` flag is enabled.

## Fix

- Instead of disabling the `incremental` flag, `yarn build` will do a force rebuild via `tsc --build --force`.
- To continue leveraging on the benefits that the `incremental` flag offers (faster compile times during development), a new `compile` command is added
    - `yarn compile` simply triggers a compilation but it respects the `incremental` flag

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,21 +37,22 @@
     "js-slang": "dist/repl/repl.js"
   },
   "scripts": {
+    "autocomplete": "node ./scripts/updateAutocompleteDocs.js",
+    "benchmark": "jest --runInBand --testPathPattern='.*benchmark.*' --testPathIgnorePatterns='/dist/'",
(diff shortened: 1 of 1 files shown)


martin-henz:
Great. This finally fixes a long-standing issue in js-slang.

PR #1297, opened 2022-09-15
Add HTML runner

zhaojj2209:
Adds HTML runner to js-slang. 

The HTML runner currently takes the input code, injects a script to the top of the code that handles error handling, and returns the code without any further checks. This is because the bulk of error handling is done when the document is rendered in the frontend iframe component; the injected script serves to route any console errors to the frontend.

More substantial changes will likely be made once module support is considered.

### DIFF
--- a/src/runner/__tests__/runners.ts
+++ b/src/runner/__tests__/runners.ts
@@ -5,6 +5,7 @@ import { FatalSyntaxError } from '../../parser/parser'
 import { Chapter, Variant } from '../../types'
 import { locationDummyNode } from '../../utils/astCreator'
 import { CodeSnippetTestCase } from '../../utils/testing'
+import { ERROR_HANDLING_SCRIPT } from '../htmlRunner'
 
 const JAVASCRIPT_CODE_SNIPPETS_NO_ERRORS: CodeSnippetTestCase[] = [
   {
@@ -176,4 +177,15 @@ describe('Additional JavaScript features are not available in Source Native', ()
       }
     }
   )
+})
+
+// HTML Unit Tests
+
+test('Error handling script is injected in HTML code', async () => {
+  const htmlDocument: string = '<p>Hello World!</p>'
+  const htmlContext: Context = mockContext(Chapter.HTML, Variant.DEFAULT)
+  const result = await runInContext(htmlDocument, htmlContext)
+
+  expect(result.status).toStrictEqual('finished')
+  expect((result as any).value).toStrictEqual(ERROR_HANDLING_SCRIPT + htmlDocument)
 })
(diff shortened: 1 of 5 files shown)


src/runner/__tests__/runners.ts
```diff
@@ -177,3 +178,14 @@ describe('Additional JavaScript features are not available in Source Native', ()
...
+
+// HTML Unit Tests
+
+test('Error handling script is injected in HTML code', async () => {
+  const htmlDocument: string = '<p>Hello World!</p>'
+  const htmlContext: Context = mockContext(Chapter.HTML, Variant.DEFAULT)
+  const result = await runInContext(htmlDocument, htmlContext)
+
+  expect(result.status).toStrictEqual('finished')
+  expect((result as any).value).toStrictEqual(ERROR_HANDLING_SCRIPT + htmlDocument)
```
ianyong, line 190:
```suggestion
  expect((result as Finished).value).toStrictEqual(ERROR_HANDLING_SCRIPT + htmlDocument)
```

Let's avoid using `any`. In fact, we should not even be performing any type casts because once we introduce a type cast into the codebase, we lose any guarantee of type safety. Ideally, we would make use of [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to avoid having to type cast, but I understand that this would be rather out of scope for this PR + the entire codebase unfortunately uses type casts everywhere.

src/runner/htmlRunner.ts
```diff
@@ -0,0 +1,20 @@
+import { IOptions, Result } from '..'
+import { Context } from '../types'
+
+export const ERROR_HANDLING_SCRIPT = `<script>
+  window.onerror = (msg, url, lineNum) => {
+    window.parent.postMessage("Line " + Math.max(lineNum - 5, 0) + ": " + msg, "*");
+  };
+</script>\n`
```
ianyong, line 8: Is there a reason why this particular newline is represented by `\n` while the previous newlines are represented by literal line breaks?

zhaojj2209: Mainly for readability; the last `\n` is to signal that the user's code begins on the next line.

src/runner/htmlRunner.ts
```diff
@@ -1,12 +1,18 @@
...
-export const ERROR_HANDLING_SCRIPT = `<script>
+const ERROR_HANDLING_SCRIPT_TEMPLATE = `<script>
   window.onerror = (msg, url, lineNum) => {
-    window.parent.postMessage("Line " + Math.max(lineNum - 5, 0) + ": " + msg, "*");
+    window.parent.postMessage("Line " + Math.max(lineNum - %d, 0) + ": " + msg, "*");
   };
 </script>\n`
 
+const errorScriptLines = ERROR_HANDLING_SCRIPT_TEMPLATE.split('\n').length - 1
+export const htmlErrorScript = ERROR_HANDLING_SCRIPT_TEMPLATE.replace(
```
ianyong, line 11: Can we standardise the names? Why does `ERROR_HANDLING_SCRIPT_TEMPLATE` produce an `htmlErrorScript`?

ianyong (changes requested):

ianyong (approved):
LGTM!

PR #1298, opened 2022-09-16
bumping version

martin-henz:


martin-henz:
We need to wait for 
https://github.com/source-academy/modules/pull/153
and
https://github.com/source-academy/frontend/pull/2208
for this to be merged, otherwise there will be incompatible js-slang changes in the frontend.

PR #1299, opened 2022-09-20
Modify SICP package to use ES6 imports

zhaojj2209:
Modifies the `build_sicp_package` script to change `global` to `exports`; this allows functions in the package to be imported ES6-style. Also changed documentation accordingly.

Note that `import 'sicp'` no longer works now; if we wish to support both import styles, the script needs to be edited such that the functions are written both to `global` and `exports`.

More work is still needed before SICP can work with TypeScript; we will likely need to look into automatically generating/manually adding a TypeScript declaration file.

### DIFF
--- a/scripts/build_sicp_package.sh
+++ b/scripts/build_sicp_package.sh
@@ -42,7 +42,7 @@ write() {
     #Write prelude names
     while read -r CURRENT_LINE
     do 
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Seems to work. Thanks for improving this package.

PR #1305, opened 2022-09-30
parser upgrade

martin-henz:
Cleanup of parse function un preparation for MCE, CS4215, and the LEGO Spike implementation.


martin-henz:
Bumping version of js-slang to 1.0.1.

PR #1311, opened 2022-10-12
Fix exposed global functions in Source Native

shenyih0ng:
## Description

Functions such as `compose_filter` is accessible globally on client side (see below). Since source native variants, relies on browser `eval` source native programs will be able to call such global functions.

image

## Fix

- Add a new check during source native transpilation to check for undefined variables
    - This will do a check against the source program context (imports, precludes, builtins) and not the global browser context

https://github.com/source-academy/js-slang/blob/46c02cd45bc5b7077aa93bfb41aa04d410be54ad/src/transpiler/transpiler.ts#L663-L664

### DIFF
--- a/src/runner/fullJSRunner.ts
+++ b/src/runner/fullJSRunner.ts
@@ -2,6 +2,7 @@
 import { Options, parse } from 'acorn'
 import { generate } from 'astring'
 import * as es from 'estree'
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Looks good.

PR #1314, opened 2022-12-05
Add Source 1 Typed variant

zhaojj2209:
Changes:
-
- Added ability to parse TS code to ESTree AST format, using a combination of [Babel Parser](https://babeljs.io/docs/en/babel-parser) and [acorn Typescript plugin](https://github.com/patternfly/patternfly-org/blob/main/packages/ast-helpers/acorn-typescript.js).
- Added a type error checker, which recurses through the parsed AST and adds any detected type errors to the context. The type error checker is designed such that all types are assumed to be "any" unless specifically declared (e.g. declared by the user using type annotations, pre-declared Source 1 prelude functions, binary operations such as "+" being restricted to only strings and numbers in Source 1, etc). Note that this type error checker is separate from the existing type checker used in type inference, however functions shared between the two checkers have been abstracted out into a separate utils file.
- Added types used in the type error checker. The types for the types are mostly the same as those used in the type inference checker, but additional ones have been added as per necessary and certain existing types were renamed for clarity.
- Modified the Source mode for the frontend Ace Editor to support syntax highlighting differences in the Source Typed variants.
- Added unit tests to cover most of Source 1 Typed functionality.

How to Test this PR:
-
`src/parser/typeParser.ts` can be ignored as it is the code for the [acorn Typescript plugin](https://github.com/patternfly/patternfly-org/blob/main/packages/ast-helpers/acorn-typescript.js). This reduces about 1k LoC of code to review, but that still makes this PR about 3k LoC large, apologies in advance.

Given the size of the PR, the main focus would therefore be to test the functionality of Source Typed to see if there are any bugs/inconsistencies. This can be done in the following ways:
1. Without frontend: Run Source 1 Typed in the js-slang REPL.
```
$ js-slang -c 1 -v typed # if js-slang is added to PATH
$ node dist/repl/repl.js -c 1 -v typed # if js-slang is not added to PATH
```
2. With frontend: [Link local js-slang to local frontend](https://github.com/source-academy/js-slang#using-your-js-slang-in-your-local-source-academy) and checkout local frontend to source-academy/frontend#2264.
3. Check the existing unit tests in `src/typeChecker/__tests__/typeErrorChecker.test.ts`.
```
$ yarn test src/typeChecker/__tests__/typeErrorChecker.test.ts
```


ianyong (commented):
@zhaojj2209 Have yet to look through everything; hoping that this initial round of review will help me better understand what is going on once my comments have been responded to.

src/editors/ace/modes/source.ts
```diff
@@ -93,12 +93,22 @@ export function HighlightRulesSelector(
...
-        return 'while|for|break|continue|let'
-      } else {
-        return ''
+        forbiddenWords += 'while|for|break|continue|let'
+      }
+      if (variant !== Variant.TYPED) {
+        forbiddenWords += '|typeof|void'
+      }
+      return forbiddenWords
+    }
```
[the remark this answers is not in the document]
zhaojj2209: Good catch! I'll make the same refactoring to `ChapterKeywordSelector` as well since it uses similar logic.

src/editors/ace/modes/source.ts
```diff
@@ -93,12 +93,22 @@ export function HighlightRulesSelector(
...
+
+    const VariantForbiddenRegexSelector = () => {
+      if (variant === Variant.TYPED) {
+        return /\.{3}|--+|\+\++|\^|(==|!=)[^=]|[$%&*+\-~\/^]=+|[^&]*&[^&]/
       }
+      return /\.{3}|--+|\+\++|\^|(==|!=)[^=]|[$%&*+\-~\/^]=+|[^&]*&[^&]|[^\|]*\|[^\|]/
```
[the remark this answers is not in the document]
zhaojj2209: The change here is to remove the highlighting of singular `|` as forbidden for Source Typed, due to the presence of union types. I will add a comment here to explain the change.

src/types.ts
```diff
@@ -291,11 +293,116 @@ export interface BlockExpression extends es.BaseExpression {
...
 
-export type TypeAnnotatedFuncDecl = TypeAnnotatedNode<es.FunctionDeclaration> & TypedFuncDecl
+export type ContiguousArrayElementExpression = Exclude<es.ArrayExpression['elements'][0], null>
+
+export type ContiguousArrayElements = ContiguousArrayElementExpression[]
 
-export type TypeAnnotation = Untypable | Typed | NotYetTyped
+// =======================================
+// Types used in type checker for type inference/type error checker for Source Typed variant
+// =======================================
```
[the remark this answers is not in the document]
zhaojj2209: Yes, I moved the exports up so that the bottom of the file consists only of types for the type inference/type error checker.

src/editors/ace/modes/source.ts
```diff
@@ -551,6 +560,30 @@ export function HighlightRulesSelector(
         if (!options || options.jsx != false)
           // @ts-ignore
           JSX.call(this)
+
+        // Adding of highlight rules for Source Typed
+        // Code referenced from https://github.com/ajaxorg/ace-builds/blob/master/src/mode-typescript.js
+        if (variant === Variant.TYPED) {
+          // @ts-ignore
+          this.$rules.no_regex.unshift(
```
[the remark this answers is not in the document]
zhaojj2209: Not really, it's mostly for consistency with the rest of the file.

src/types.ts
```diff
@@ -336,55 +443,52 @@ export interface FunctionType {
...
-export type BindableType = Type | ForAll | PredicateType
+export interface PredicateType {
+  kind: 'predicate'
+  ifTrueType: Type | ForAll
+}
 
+/**
+ * Stores the type/declaration type of variables.
+ * The array distinguishes between the different scopes in the program
+ * (e.g. first element is the global scope, last element is the closest).
```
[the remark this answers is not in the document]
zhaojj2209: It's referring to the how `TypeEnvironment` is an array of objects; will reword this comment for clarity.

src/types.ts
```diff
@@ -336,55 +443,52 @@ export interface FunctionType {
...
 }
 
-export type BindableType = Type | ForAll | PredicateType
+export interface PredicateType {
+  kind: 'predicate'
+  ifTrueType: Type | ForAll
+}
 
+/**
+ * Stores the type/declaration type of variables.
```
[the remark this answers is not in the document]
zhaojj2209: Type is actual type of the constant/variable being declared, declaration type is whether the declaration is a constant or variable.

src/types.ts
```diff
@@ -291,11 +293,116 @@ export interface BlockExpression extends es.BaseExpression {
...
+  parameters: NodeWithDeclaredTypeAnnotation<es.Identifier>[]
+  typeAnnotation: AnnotationTypeNode
+}
+
+export interface UnionTypeNode extends BaseTypeNode {
+  type: TSTypeAnnotationType.TSUnionType
+  types: BaseTypeNode[]
+}
+
+export interface TypeAliasDeclarationNode extends BaseTypeNode {
```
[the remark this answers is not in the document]
zhaojj2209: Good catch, will modify the code to use this type for better typing.

src/types.ts
```diff
@@ -291,11 +293,116 @@ export interface BlockExpression extends es.BaseExpression {
...
+  TSSymbolKeyword = 'TSSymbolKeyword',
+  TSUndefinedKeyword = 'TSUndefinedKeyword',
+  TSUnknownKeyword = 'TSUnknownKeyword',
+  TSVoidKeyword = 'TSVoidKeyword',
+  // Nodes
+  TSTypeAliasDeclaration = 'TSTypeAliasDeclaration',
+  TSAsExpression = 'TSAsExpression',
+  TSInterfaceDeclaration = 'TSInterfaceDeclaration',
+  TSTypeReference = 'TSTypeReference'
+}
```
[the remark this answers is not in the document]
zhaojj2209: This enum is meant to encompass all the TS-related node types in a ESTree AST that supports type syntax. Will rename this enum to `TSNodeType` for clarity.

[Example of ESTree AST with TS support](https://typescript-eslint.io/play#ts=4.8.3&sourceType=module&showAST=es&code=C4TwDgpgBMEM7CgXiggTgSwHYHMDcAsAFADGA9lgjPMAFzVUoDkAjE3kA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA)

src/types.ts
```diff
@@ -291,11 +293,116 @@ export interface BlockExpression extends es.BaseExpression {
...
+
+// Types for parsed TS nodes used in Source Typed variants
+export type NodeWithDeclaredTypeAnnotation<T extends es.Node> = DeclaredTypeAnnotation & T
+
+export type DeclaredTypeAnnotation = {
+  typeAnnotation?: AnnotationTypeNode
```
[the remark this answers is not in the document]
zhaojj2209: I'm guessing the name here is causing some confusion; the initial idea was to distinguish between `InferredTypeAnnotation` (originally named `TypeAnnotation`) that's used by the type checker for type inference, and `DeclaredTypeAnnotation` that's used by the type error checker for the Source Typed variant. I'll change the naming to `InferredType` and `TypeAnnotation` instead for clarity.

src/editors/ace/modes/source.ts
```diff
@@ -551,6 +560,30 @@ export function HighlightRulesSelector(
         if (!options || options.jsx != false)
           // @ts-ignore
           JSX.call(this)
+
+        // Adding of highlight rules for Source Typed
+        // Code referenced from https://github.com/ajaxorg/ace-builds/blob/master/src/mode-typescript.js
+        if (variant === Variant.TYPED) {
+          // @ts-ignore
```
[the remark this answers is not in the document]
ianyong: Hmm okay, I think the context in which it is used is similar enough to the existing instances, so we can leave it as is. Main idea being that we don't want to leave the codebase in a worse state than when we started.

src/types.ts
```diff
@@ -291,11 +293,116 @@ export interface BlockExpression extends es.BaseExpression {
...
+  NEVER = 'never',
+  NULL = 'null',
+  NUMBER = 'number',
+  OBJECT = 'object',
+  STRING = 'string',
+  SYMBOL = 'symbol',
+  UNDEFINED = 'undefined',
+  UNKNOWN = 'unknown',
+  VOID = 'void'
+}
```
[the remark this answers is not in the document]
zhaojj2209: Will refactor this enum to the following:
- `PrimitiveType`: number, boolean, string, undefined, null
- `TSAllowedTypes`: void, any
- `TSDisallowedTypes`: everything else

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+    }
+    case 'Program':
+    case 'BlockStatement': {
+      let returnType: Type = tVoid
+      pushEnv(env)
+
+      if (node.type === 'Program') {
+        // Import statements should only exist in program body
+        handleImportDeclarations(node, context, env)
+      }
```
ianyong, line 138: If `import` statements that are not in the top-level program body are already flagged as an error by the parser, maybe we don't need to specifically perform this check? My concern here is that this is an implicit dependency on the parser. What if one day the parser is changed to allow `import` statements in functions? The `import` statements would silently fail. I think we can trust the parser with the responsibility of determining whether `import` statements are valid.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+        if (node.declarations[0].id.type !== 'Identifier') {
+          throw new TypeError(node, 'Variable declaration ID should be an identifier')
+        }
+        const id = node.declarations[0].id as NodeWithTypeAnnotation<es.Identifier>
+        const expectedType = getAnnotatedType(id.typeAnnotation, context, env)
+
+        // Save variable type and decl kind in type env
+        setType(id.name, expectedType, env)
+        setDeclKind(id.name, node.kind, env)
+      default:
```
ianyong, line 434: There is no `break` or `return` statement here, so this case will fallthrough to the default case. Is this intended?

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+) {
+  node.body.forEach(node => {
+    switch (node.type) {
+      case 'FunctionDeclaration':
+        if (node.id === null) {
+          // Block should not be reached since node.id is only null when function declaration
+          // is part of `export default function`, which is not used in Source
+          throw new TypeError(node, 'Function declaration should always have an identifier')
+        }
+        const params = node.params as NodeWithTypeAnnotation<es.Identifier>[]
```
ianyong, line 403: `node.params` has type `es.Pattern[]` where `Pattern` is `Identifier | ObjectPattern | ArrayPattern | RestElement | AssignmentPattern | MemberExpression`. Is it correct to typecast a `Pattern` it as just an `Identifier`? Can function arguments be any of the other types? If not, maybe a comment can be added here to explain this.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+        const returnType = getAnnotatedType(
+          (node as NodeWithTypeAnnotation<es.FunctionDeclaration>).returnType,
```
ianyong, line 405: Thanks to the use of a discriminant property `type`, TypeScript is able to infer that the type of `node` is `es.FunctionDeclaration` at this point based off the original type of `es.Program | es.BlockStatement`. However, my question is why we can cast to a `NodeWithTypeAnnotation` without any runtime errors. This seems to imply that `es.Program | es.BlockStatement` is already a mistyping.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+  context: Context,
+  env: TypeEnvironment
+): Type {
+  if (!annotationNode) {
+    return tAny
+  }
+  const annotatedTypeNode = annotationNode.typeAnnotation
+  switch (annotatedTypeNode.type) {
+    case TSNodeType.TSFunctionType:
+      const fnTypeNode = annotatedTypeNode as FunctionTypeNode
```
ianyong, line 673: The typecasting here is unnecessary if we make use of TypeScript's [discriminated unions](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions).

```ts
@@ -373,39 +373,37 @@ export type TypeAnnotation = {
   returnType?: AnnotationTypeNode
 }
 
-export interface BaseTypeNode extends es.BaseNode {
-  type: TSNodeType
-}
+export type BaseTypeNode = AnnotationTypeNode | FunctionTypeNode | UnionTypeNode | TypeAliasDeclarationNode | AsExpressionNode | TypeReferenceNode
 
-export interface AnnotationTypeNode extends BaseTypeNode {
+export interface AnnotationTypeNode extends es.BaseNode {
   type: TSNodeType.TSAnnotationType
   typeAnnotation: BaseTypeNode | FunctionTypeNode | UnionTypeNode | TypeReferenceNode
 }
 
-export interface FunctionTypeNode extends BaseTypeNode {
+export interface FunctionTypeNode extends es.BaseNode {
   type: TSNodeType.TSFunctionType
   parameters: NodeWithTypeAnnotation<es.Identifier>[]
   typeAnnotation: AnnotationTypeNode
 }
 
-export interface UnionTypeNode extends BaseTypeNode {
+export interface UnionTypeNode extends es.BaseNode {
   type: TSNodeType.TSUnionType
   types: BaseTypeNode[]
 }
 
-export interface TypeAliasDeclarationNode extends BaseTypeNode {
+export interface TypeAliasDeclarationNode extends es.BaseNode {
   type: TSNodeType.TSTypeAliasDeclaration
   id: es.Identifier
   typeAnnotation: BaseTypeNode
 }
 
-export interface AsExpressionNode extends BaseTypeNode {
+export interface AsExpressionNode extends es.BaseNode {
   type: TSNodeType.TSAsExpression
   expression: es.Node
   typeAnnotation: BaseTypeNode
 }
 
-export interface TypeReferenceNode extends BaseTypeNode {
+export interface TypeReferenceNode extends es.BaseNode {
   type: TSNodeType.TSTypeReference
   typeName: es.Identifier
 }
```

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+      const unionTypeNode = annotatedTypeNode as UnionTypeNode
+      const unionTypes = unionTypeNode.types.map(type =>
+        getPrimitiveType(annotationNode, type, context)
+      )
+      return mergeTypes(...unionTypes)
+    case TSNodeType.TSIntersectionType:
+      throw new TypeError(
+        annotationNode as unknown as es.Node,
+        'Intersection types are not allowed'
+      )
```
ianyong, line 688: A benefit of making use of discriminated unions is that we will find out that this case is impossible! Unless of course everything is already mistyped at this point...

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+  const annotatedTypeNode = annotationNode.typeAnnotation
+  switch (annotatedTypeNode.type) {
+    case TSNodeType.TSFunctionType:
+      const fnTypeNode = annotatedTypeNode as FunctionTypeNode
+      const fnTypes = getParamTypes(fnTypeNode.parameters, context, env)
+      // Return type will always be last item in types array
+      fnTypes.push(getAnnotatedType(fnTypeNode.typeAnnotation, context, env))
+      return tFunc(...fnTypes)
+    case TSNodeType.TSUnionType:
+      const unionTypeNode = annotatedTypeNode as UnionTypeNode
```
ianyong, line 679: Same here, the typecasting wouldn't be necessary if discriminated unions were used.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+        getPrimitiveType(annotationNode, type, context)
+      )
+      return mergeTypes(...unionTypes)
+    case TSNodeType.TSIntersectionType:
+      throw new TypeError(
+        annotationNode as unknown as es.Node,
+        'Intersection types are not allowed'
+      )
+    case TSNodeType.TSTypeReference:
+      const typeReferenceNode = annotatedTypeNode as TypeReferenceNode
```
ianyong, line 690: Here too.

ianyong (commented):
@zhaojj2209 Submitting another round of review to get your thoughts on the comments before I continue.

src/errors/typeErrors.ts
```diff
@@ -398,3 +409,146 @@ export class InconsistentPredicateTestError implements SourceError {
...
+  public severity = ErrorSeverity.ERROR
+
+  constructor(
+    public node: NodeWithTypeAnnotation<es.Node>,
+    public actualTypeString: string,
+    public expectedTypeString: string
+  ) {}
+
+  get location() {
+    return this.node.loc!
```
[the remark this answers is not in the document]
zhaojj2209: The acorn parser has an option to add location info to the parsed AST, that's why `loc` is typed as being possibly `null` or `undefined`. We have this location info option turned on (`locations` is set to `true`) for the parser, which should guarantee that `loc` always has a value, but this unfortunately is not reflected int he type, resulting in the use of non-null assertions.

src/parser/parser.ts
```diff
@@ -115,7 +118,22 @@ export function parseAt(source: string, num: number) {
 export function parse(source: string, context: Context) {
   let program: es.Program | undefined
   try {
-    program = acornParse(source, createAcornParserOptions(context)) as unknown as es.Program
+    if (context.variant === Variant.TYPED) {
+      // For Source Typed variant, the code is parsed twice, first using the custom TypeParser and then using Babel Parser.
+      // This is a workaround as the custom TypeParser does not cover all type annotation cases needed for Source Typed
+      // and the Babel Parser does not allow for no semicolon/trailing comma errors when parsing.
+      program = TypeParser.parse(source, createAcornParserOptions(context)) as unknown as es.Program
```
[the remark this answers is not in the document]
zhaojj2209: `acorn.Node` has its `type` attribute being typed as `string`, so it can refer to any node in an ESTree AST. However, since we know that the AST returned by `acorn.parse` should always start with type `program`, we cast it to `es.Program` from the `estree` library so as to make use of its much more detailed typing.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+  // Deep copy type environment to avoid modifying type environment in the context,
+  // which might affect the type inference checker
+  const env: TypeEnvironment = cloneDeep(context.typeEnvironment)
+  // Override predeclared function types
+  for (const [name, type] of source1TypeOverrides) {
+    setType(name, type, env)
+  }
+  try {
+    typeCheckAndReturnType(program, context, env)
+    return removeTSNodes(program)
```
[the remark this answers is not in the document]
zhaojj2209: Thanks for the catch! Will make the edit

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,856 @@
...
+      // Infers type
+      if (node.value === undefined) {
+        return tUndef
+      }
+      if (node.value === null) {
+        // TODO: Handle null literal types for Source 2 and above
+        return tAny
+      }
+      const typeOfLiteral = typeof node.value as PrimitiveType
+      if (Object.values(PrimitiveType).includes(typeOfLiteral)) {
```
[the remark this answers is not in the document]
zhaojj2209: There should be no issue in removing this check. Will double-check and make the edit.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: Thanks for the catch! Though the return type is the same as default case, it is still best to return the value instead of falling through.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: This is a special case where the type of annotationNode is not `es.Node` per se, but the shared attributes are enough such that the casting is safe. I'll try to find an alternative that doesn't involve typecasting though.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: Turns out removing this check actually causes type errors.

src/typeChecker/typeErrorChecker.ts
ianyong: For posterity, this was discussed offline and resolved by correcting the mistyped `node` argument in the function declaration.

src/parser/parser.ts
```diff
@@ -115,7 +126,26 @@ export function parseAt(source: string, num: number) {
...
+      // This is a workaround as the custom TypeParser does not cover all type annotation cases needed for Source Typed
+      // and the Babel Parser does not allow for no semicolon/trailing comma errors when parsing.
+      program = TypeParser.parse(
+        source,
+        createAcornParserOptions(context)
+      ) as unknown as NodeWithTypeAnnotation<es.Program>
+      program = babelParse(source, {
+        sourceType: 'module',
+        plugins: ['typescript', 'estree']
+      }).program as unknown as NodeWithTypeAnnotation<es.Program>
```
ianyong, line 141: Maybe it would be good to add a comment here explaining why you cast the Babel types to estree types as we discussed offline?

src/parser/parser.ts
```diff
@@ -115,7 +126,26 @@ export function parseAt(source: string, num: number) {
 export function parse(source: string, context: Context) {
   let program: es.Program | undefined
   try {
-    program = acornParse(source, createAcornParserOptions(context)) as unknown as es.Program
+    if (context.variant === Variant.TYPED) {
+      // For Source Typed variant, the code is parsed twice, first using the custom TypeParser
+      // (Acorn parser with plugin that allows for parsing of TS syntax), then using Babel Parser.
+      // This is a workaround as the custom TypeParser does not cover all type annotation cases needed for Source Typed
+      // and the Babel Parser does not allow for no semicolon/trailing comma errors when parsing.
+      program = TypeParser.parse(
```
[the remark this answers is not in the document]
zhaojj2209: Good catch!

zhaojj2209:
As discussed with @ianyong offline, the `estree` types have been cloned into a separate file `tsESTree.ts` and modified to support type syntax. Further tightening of the types will be done once Source 2-4 is added and all types are finalised.

src/types.ts
```diff
@@ -291,11 +293,60 @@ export interface BlockExpression extends es.BaseExpression {
...
+  OBJECT = 'object',
+  SYMBOL = 'symbol',
+  UNKNOWN = 'unknown'
+}
+
+export const TSBasicType = {
+  ...PrimitiveType,
+  ...TSAllowedTypes,
+  ...TSDisallowedTypes
+}
```
ianyong, line 338:
```suggestion
export type PrimitiveType =
  | 'boolean'
  | 'null'
  | 'number'
  | 'string'
  | 'undefined';

export type TSAllowedTypes =
  | 'any'
  | 'void'

export type TSDisallowedTypes =
  | 'bigint'
  | 'never'
  | 'object'
  | 'symbol'
  | 'unknown'
```

Using union types might result in easier handling of types as compared to enums, and is consistent with the ESTree types.

src/typeChecker/tsESTree.ts
```diff
@@ -0,0 +1,778 @@
...
+  TSAsExpression = 'TSAsExpression',
+  TSInterfaceDeclaration = 'TSInterfaceDeclaration',
+  TSTypeReference = 'TSTypeReference'
+}
+
+export const TSNodeType = {
+  ...TSTypeAnnotationType,
+  ...TSTypeKeyword,
+  ...TSOnlyNode
+}
```
ianyong, line 705:
```suggestion
export type TSTypeAnnotationType =
  | 'TSAnnotationType'
  | 'TSFunctionType'
  | 'TSUnionType'
  | 'TSIntersectionType'
  | 'TSLiteralType'

export type TSTypeKeyword =
  | 'TSAnyKeyword'
  | 'TSBigIntKeyword'
  | 'TSBooleanKeyword'
  | 'TSNeverKeyword'
  | 'TSNullKeyword'
  | 'TSNumberKeyword'
  | 'TSObjectKeyword'
  | 'TSStringKeyword'
  | 'TSSymbolKeyword'
  | 'TSUndefinedKeyword'
  | 'TSUnknownKeyword'
  | 'TSVoidKeyword'

export type TSOnlyNode =
  | 'TSTypeAliasDeclaration'
  | 'TSAsExpression'
  | 'TSInterfaceDeclaration'
  | 'TSTypeReference'
}
```

Same here.

src/typeChecker/utils.ts
```diff
@@ -0,0 +1,391 @@
...
+  [tsEs.TSNodeType.TSNeverKeyword]: TSBasicType.NEVER,
+  [tsEs.TSNodeType.TSNullKeyword]: TSBasicType.NULL,
+  [tsEs.TSNodeType.TSNumberKeyword]: TSBasicType.NUMBER,
+  [tsEs.TSNodeType.TSObjectKeyword]: TSBasicType.OBJECT,
+  [tsEs.TSNodeType.TSStringKeyword]: TSBasicType.STRING,
+  [tsEs.TSNodeType.TSSymbolKeyword]: TSBasicType.SYMBOL,
+  [tsEs.TSNodeType.TSUndefinedKeyword]: TSBasicType.UNDEFINED,
+  [tsEs.TSNodeType.TSUnknownKeyword]: TSBasicType.UNKNOWN,
+  [tsEs.TSNodeType.TSVoidKeyword]: TSBasicType.VOID
+}
```
ianyong, line 44:
```suggestion
export const typeAnnotationKeywordToBasicTypeMap: Record<tsEs.TSTypeKeyword, TSBasicType> = {
  'TSAnyKeyword': 'any',
  'TSBigIntKeyword': 'bigint',
  'TSBooleanKeyword': 'boolean',
  'TSNeverKeyword': 'never',
  'TSNullKeyword': 'null',
  'TSNumberKeyword': 'number',
  'TSObjectKeyword': 'object',
  'TSStringKeyword': 'string',
  'TSSymbolKeyword': 'symbol',
  'TSUndefinedKeyword': 'undefined',
  'TSUnknownKeyword': 'unknown',
  'TSVoidKeyword': 'void'
}
```

A benefit of using union types over enums is that your typing can be tighter here. Instead of using `TSNodeType` which is a union type of your 3 enums `TSTypeAnnotationType | TSTypeKeyword | TSOnlyNode`, you can use `TSTypeKeyword`.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,914 @@
...
+        // Type error where function does not return anything when it should
+        context.errors.push(new FunctionShouldHaveReturnValueError(node))
+      } else {
+        checkForTypeMismatch(node, actualReturnType, expectedReturnType, context)
+      }
+
+      // No need to save variable type again, return undefined type
+      return tUndef
+    case 'VariableDeclaration': {
+      const id = node.declarations[0].id as tsEs.Identifier
```
ianyong, line 236: Would be good to add a comment explaining why this casting is safe. Alternatively, you can add a check to narrow the type, throwing an error if it's an unexpected type.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,914 @@
...
+      const init = node.declarations[0].init!
```
ianyong, line 237: Would be good to add a comment explaining why this non-null assertion is safe. Alternatively, you can add a check to narrow the type, throwing an error if it's `null` or `undefined`.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,914 @@
...
+      // Return type number
+      checkForTypeMismatch(node, leftType, tNumber, context)
+      checkForTypeMismatch(node, rightType, tNumber, context)
+      return tNumber
+    case '+':
+      // Both sides can only be number, string, or any
+      // However, case where one side is string and other side is number is not allowed
+      if (leftTypeString === 'number' || leftTypeString === 'string') {
+        checkForTypeMismatch(node, rightType, leftType, context)
+        // If string + number, return string; else return left type
```
ianyong, line 457: So, is one side being string and the other side being number allowed or not allowed? These comments seem to contradict each other.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,914 @@
...
+    context.errors.push(new FunctionShouldHaveReturnValueError(node))
+  } else {
+    checkForTypeMismatch(node, actualReturnType, expectedReturnType, context)
+  }
+
+  const types = getParamTypes(params, context, env)
+  // Return type will always be last item in types array
+  types.push(expectedReturnType)
+  const fnType = tFunc(...types)
+  return fnType
```
ianyong, line 544: Minor nit: Possible to inline variable.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,914 @@
...
+      for (const type of currType.types) {
+        if (!containsType(mergedTypes, type)) {
+          mergedTypes.push(type)
+        }
+      }
+    } else if (!containsType(mergedTypes, currType)) {
+      mergedTypes.push(currType)
+    } else {
+      // Duplicate type, do nothing
+    }
```
ianyong, line 796:
```suggestion
    } else {
      if (!containsType(mergedTypes, currType)) {
        mergedTypes.push(currType)
      }
    }
```

Minor nit: Control flow might be easier to understand if each level does only one thing. Here, the first level is checking the `kind` of `currType`, while the second level is checking whether to push the type into `mergedTypes`.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,914 @@
...
+    case 'Literal':
+    case 'Identifier': {
+      return node
+    }
+    case 'Program':
+    case 'BlockStatement': {
+      const newBody: tsEs.Statement[] = []
+      node.body.forEach(stmt => {
+        const type = stmt.type
+        if (type.startsWith('TS')) {
```
ianyong, line 849: Minor nit: This code can be made more robust to future changes by declaring some container containing all the types you want to match against. For example, if I were to add a new type that starts with `TS`, but is not meant to be matched here, I wouldn't know that it will be treated differently just by looking at the types.

ianyong (changes requested):
The improved types look good! Thanks for making the change. I believe that moving forward, the improved typing will make development much easier.

This should be my final round of comments on code quality. It's up to you whether you want to address the minor nits or not; they fall more on the subjective side so I won't force them.

I will be reviewing the test cases + testing functionality now - posting this set of reviews first so that you can work on them earlier.

Note that as per the PR description, I am not reviewing `src/parser/typeParser.ts` since it is the code for the acorn TypeScript plugin.

src/typeChecker/tsESTree.ts
```diff
@@ -0,0 +1,777 @@
...
+  | 'TSVoidKeyword'
+
+export type TSOnlyNode =
+  | 'TSTypeAliasDeclaration'
+  | 'TSAsExpression'
+  | 'TSInterfaceDeclaration'
+  | 'TSTypeReference'
+
+// All node types in a ESTree AST with TS support
+export type TSNodeType = TSTypeAnnotationType | TSTypeKeyword | TSOnlyNode
```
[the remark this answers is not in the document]
zhaojj2209: Will keep for abstraction purposes.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -0,0 +1,914 @@
...
+/**
+ * Entry function for type error checker.
+ * Checks programs for type errors, and returns the program with all TS-related nodes removed.
+ */
+export function checkForTypeErrors(program: tsEs.Program, context: Context): es.Program {
+  // Deep copy type environment to avoid modifying type environment in the context,
+  // which might affect the type inference checker
+  const env: TypeEnvironment = cloneDeep(context.typeEnvironment)
+  // Override predeclared function types
+  for (const [name, type] of source1TypeOverrides) {
```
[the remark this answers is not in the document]
zhaojj2209: Yep, will change this accordingly when adding source 2-4.

src/typeChecker/__tests__/typeErrorChecker.test.ts
```diff
@@ -0,0 +1,984 @@
...
+
+    const code = `const x1: string | number = 1;
+      const x2: string = x1 as string;
+    `
+
+    const program = parse(code, context)
+    expect(program).toMatchSnapshot() // Should not contain TSAsExpression node
+  })
+
+  it('only supports downcasting', () => {
```
ianyong, line 456: My understanding of downcasting is in the context of OOP, where a parent object is typecast to a child object. What does it mean here since we're only dealing with primitives?

I see things like
```ts
const x3: string = x1 as string; // no error
```
which is a narrowing of types.

At the same time, I see things like
```ts
const x8 = x1 as any; // error
```
which is a widening of types.

What exactly does "down" mean here?

src/typeChecker/__tests__/typeErrorChecker.test.ts
```diff
@@ -0,0 +1,984 @@
+import { parseError } from '../..'
+import { mockContext } from '../../mocks/context'
+import { parse } from '../../parser/parser'
+import { Chapter, Variant } from '../../types'
+
+describe('primitive types', () => {
+  it('does not throw errors for allowed primitive types', () => {
+    const context = mockContext(Chapter.SOURCE_1, Variant.TYPED)
```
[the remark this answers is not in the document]
zhaojj2209: I agree, the test files are getting too big for all the tests for source 1-4 to be in one file. Will make the changes and put the tests for each Typed variant in a separate file.

src/typeChecker/__tests__/typeErrorChecker.test.ts
zhaojj2209: The only form of typecasting allowed in Source would be to typecast a union type to its subset (e.g. typecasting `string | number` to `number`), this is what "downcasting" refers to. If needed, I can reword this (e.g. `only supports casting to a sub-type`).

ianyong: Should casting to `any` be handled in a separate test case then?

zhaojj2209:
> Random weird things I found while manual testing:
> 
> 1. `const x: void = null;` results in `Line 1: Type 'object' is not assignable to type 'void'.` instead of `Line 1: Type 'null' is not assignable to type 'void'.`.
> 2. The function argument name can be omitted if it is primitive type, but not otherwise:
>    ```ts
>    const f: (number) => number = x => x + 1;
>    // Works
>    const f2: (f: (number) => number, number) => number = (f, x) => f(x);
>    // Does not work
>    // const f2: ((number) => number, number) => number = (f, x) => f(x);
>    f2(f, 6);
>    ```

I am unable to replicate point 1. This may have been an error from an old commit, as I have since patched a fix.
Point 2 is unfortunately a parser limitation, so I am unable to do much about it.

ianyong:
> I am unable to replicate point 1. This may have been an error from an old commit, as I have since patched a fix.

Oops, my bad, you are right. This no longer occurs on the latest version after rebuilding js-slang.

PR #1320, opened 2022-12-18
Add Source 2 Typed Variant

zhaojj2209:
Changes:
-
- Added support for `null` to type error checker.
- Added `Pair` and `List` types to type error checker.
- Added type checking for Source 2 library functions (e.g. `pair`, `list`, `head`, `tail`).
- Added tests for Source 2 Typed.

How to Test this PR:
-
1. Without frontend: Run Source 2 Typed in the js-slang REPL.
```
$ js-slang -c 2 -v typed # if js-slang is added to PATH
$ node dist/repl/repl.js -c 2 -v typed # if js-slang is not added to PATH
```
2. With frontend: [Link local js-slang to local frontend](https://github.com/source-academy/js-slang#using-your-js-slang-in-your-local-source-academy) and checkout local frontend to source-academy/frontend#2293.
3. Check the existing unit tests in `src/typeChecker/__tests__/source2Typed.test.ts`.
```
$ yarn test src/typeChecker/__tests__/source2Typed.test.ts
```

### DIFF
--- a/src/errors/typeErrors.ts
+++ b/src/errors/typeErrors.ts
@@ -594,7 +594,7 @@ export class InvalidNumberOfArgumentsTypeError implements SourceError {
   }
 }
 
-export class IncorrectNumberOfTypeArgumentsError implements SourceError {
+export class InvalidNumberOfTypeArgumentsForGenericTypeError implements SourceError {
   public type = ErrorType.TYPE
   public severity = ErrorSeverity.ERROR
   public calleeStr: string
--- a/src/errors/typeErrors.ts
+++ b/src/errors/typeErrors.ts
@@ -613,3 +613,23 @@ export class InvalidNumberOfTypeArgumentsForGenericTypeError implements SourceEr
     return this.explain()
   }
 }
+
+export class TypeAliasNameNotAllowedError implements SourceError {
+  public type = ErrorType.TYPE
+  public severity = ErrorSeverity.ERROR
+  public calleeStr: string
+
+  constructor(public node: tsEs.TSTypeAliasDeclaration, public name: string) {}
+
+  get location() {
+    return this.node.loc!
+  }
+
+  public explain() {
+    return `Type alias name cannot be '${this.name}'.`
+  }
+
+  public elaborate() {
+    return this.explain()
+  }
+}
--- a/src/errors/typeErrors.ts
+++ b/src/errors/typeErrors.ts
@@ -593,3 +593,23 @@ export class InvalidNumberOfArgumentsTypeError implements SourceError {
     return `Try calling function ${calleeStr} again, but with ${this.expected} argument${pluralS} instead. Remember that arguments are separated by a ',' (comma).`
   }
 }
+
+export class IncorrectNumberOfTypeArgumentsError implements SourceError {
+  public type = ErrorType.TYPE
+  public severity = ErrorSeverity.ERROR
+  public calleeStr: string
+
+  constructor(public node: tsEs.TSTypeReference, public name: string, public expected: number) {}
+
+  get location() {
+    return this.node.loc!
+  }
+
+  public explain() {
+    return `Generic type '${this.name}' requires ${this.expected} type argument(s).`
+  }
+
+  public elaborate() {
+    return this.explain()
+  }
+}
--- a/src/typeChecker/__tests__/source2Typed.test.ts
+++ b/src/typeChecker/__tests__/source2Typed.test.ts
@@ -11,10 +11,11 @@ beforeEach(() => {
 
 describe('null type', () => {
   it('handles type mismatches correctly', () => {
-    const code = `const x1: null = null;
-      const x2: null = '1';
-      const x3: boolean = null;
-      const x4: undefined = null;
+    const code = `const x1: null = null; // no error
+      const x2: null = '1'; // error
+      const x3: boolean = null; // error
+      const x4: undefined = null; // error
+      const x5: null = list(); // no error as null is empty list
     `
 
     parse(code, context)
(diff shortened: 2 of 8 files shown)


src/typeChecker/typeErrorChecker.ts
```diff
@@ -251,7 +257,61 @@ function typeCheckAndReturnType(node: tsEs.Node, context: Context, env: TypeEnvi
...
+          checkForTypeMismatch(node, actualType, expectedType, context)
+          if (context.errors.length > numErrors) {
+            // If errors were found, return any type
+            return tAny
+          }
+          if (actualType.kind === 'pair') {
+            return fnName === 'head' ? actualType.headType : actualType.tailType
+          }
+          if (actualType.kind === 'list') {
+            return fnName === 'head' ? actualType.elementType : actualType
```
ianyong, line 309: Given that `List` is defined as such:
```ts
export interface List {
  kind: 'list'
  elementType: Type
  // Used in Source Typed variants to check for type mismatches against pairs
  typeAsPair?: Pair
}
```
If we take the `tail` of a list, the `typeAsPair` remains unchanged despite the list itself changing. Is this intended?

src/typeChecker/typeErrorChecker.ts
```diff
@@ -397,8 +456,21 @@ function addTypeDeclarationsToEnvironment(
         setDeclKind(id.name, node.kind, env)
         break
       case 'TSTypeAliasDeclaration':
+        const alias = node.id.name
+        if (context.chapter >= 2 && (alias === 'Pair' || alias === 'List')) {
+          throw new TypecheckError(
+            node,
+            `Type '${alias}' is a reserved type and cannot be redeclared`
+          )
+        }
```
ianyong, line 465: After looking at this, I realised that I can do things like:
```ts
type string = number;
```

For reference, TypeScript disallows this:
```
TS2457: Type alias name cannot be 'string'.
```

src/typeChecker/utils.ts
```diff
@@ -352,16 +362,41 @@ export const temporaryStreamFuncs: [string, BindableType][] = [
...
+  ['member', tFunc(tAny, tList(tAny), tList(tAny))],
+  ['remove', tFunc(tAny, tList(tAny), tList(tAny))],
+  ['remove_all', tFunc(tAny, tList(tAny), tList(tAny))],
+  ['reverse', tFunc(tList(tAny), tList(tAny))],
+  // misc functions
+  // TODO: Add support for type checking of functions with variable no. of args
+  ['display_list', tForAll(tAny)],
+  ['draw_data', tForAll(tAny)],
+  ['equal', tFunc(tAny, tAny, tBool)]
 ]
```
ianyong, line 400: To confirm, this means that once I run any kind of operation on lists, I lose the type of its elements?

src/typeChecker/__tests__/source2Typed.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    `
+
+    parse(code, context)
+    expect(parseError(context.errors)).toMatchInlineSnapshot(`
+      "Line 2: Expected 2 arguments, but got 1.
+      Line 3: Expected 2 arguments, but got 3."
+    `)
+  })
+
+  it('lists are pairs', () => {
```
ianyong, line 61: Is this always true? What about the empty list?

src/typeChecker/__tests__/source2Typed.test.ts
```diff
@@ -0,0 +1,196 @@
...
+describe('list', () => {
+  it('handles type mismatches correctly', () => {
+    const code = `const x1: List<number> = list(1, 2, 3); // no error
+      const x2: List<number> = list(); // no error
+      const x3: List<number> = list('1'); // error
+      const x4: List<number> = 1; // error
+      const x5: List<number> = list(1, '2'); // error
+      const x6: List<number | string> = list(1, '2'); // no error
+      const x7: List<number, string> = list(1, '2'); // error
+    `
```
ianyong, line 80: Maybe can add a test case for a list of `null`s for completeness?

src/typeChecker/typeErrorChecker.ts
ianyong: Would also be good to add a test case that can catch this issue as a form of regression testing.

ianyong (changes requested):

src/errors/typeErrors.ts
```diff
@@ -593,3 +593,23 @@ export class InvalidNumberOfArgumentsTypeError implements SourceError {
     return `Try calling function ${calleeStr} again, but with ${this.expected} argument${pluralS} instead. Remember that arguments are separated by a ',' (comma).`
   }
 }
+
+export class IncorrectNumberOfTypeArgumentsError implements SourceError {
```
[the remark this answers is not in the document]
zhaojj2209: `InvalidNumberOfArguments` is for arguments to function calls, while `IncorrectNumberOfTypeArgumentsError` is for type parameters to generic types (e.g. `Pair`, `List`). Will rename this error for clarity.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: Thanks for the catch! Will make the necessary changes.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: Got it, will follow TS and disallow overwriting of basic types using type aliases.

src/typeChecker/utils.ts
zhaojj2209: Unfortunately as of right now the type error checker is not yet powerful enough to infer the return type based on the types of its arguments. I will KIV this for future enhancements.

src/typeChecker/__tests__/source2Typed.test.ts
zhaojj2209: I don't think `null` is considered a pair, given that the definition of a list is "either `null` or a pair whose tail is a list". Will rename this for clarity and also add testcases for the case where `null` should not be typed as `Pair`.

ianyong (approved):

PR #1321, opened 2022-12-25
Ignore unused arguments starting with underscore

ianyong:
This is useful for when we need to declare functions that take in unused arguments. For example, the argument `context` is not used in the following code snippet, yet cannot be removed:
https://github.com/source-academy/js-slang/blob/e549211a7356708e5df34f86c56897e4da55dbdd/src/interpreter/interpreter.ts#L336-L340

This results in a warning being thrown by ESLint.

By convention, arguments are denoted as unused by prepending their names with `_`:
```
Literal: function*(node: es.Literal, _context: Context) { 
  return node.value
}, 
```

We can make ESLint ignore arguments that start with an underscore by making use of the `argsIgnorePattern` option (see [here](https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern) for more information).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -37,6 +37,7 @@
     "@typescript-eslint/no-unsafe-call": "off",
     "@typescript-eslint/no-unsafe-member-access": "off",
     "@typescript-eslint/no-unsafe-return": "off",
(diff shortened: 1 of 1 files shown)


ianyong:
This change is motivated by the fact that my upcoming PR adds (necessary) unused arguments which ESLint throws warning for.

shenyih0ng (approved):

PR #1323, opened 2022-12-28
Add documentation and specs for Source 1 & 2 Typed 

zhaojj2209:
Adds pages on Source 1 & 2 Typed to the SA docs website, including the detailed specifications.

How to test: Follow the instructions on [building the documentation](https://github.com/source-academy/js-slang#documentation), then check the PDF specs and docs website generated. Looks like links to the PDFs are broken when building the docs website on local, though, so the PDF specs and docs webpages need to be checked separately.


martin-henz (approved):
Looks good. I'm approving and merging, and will ask the team in Sweden to take a look. I'll take a closer look in the next few days as well.

PR #1327, opened 2023-01-02
Add Source 3 Typed Variant

zhaojj2209:
Changes:
-
- Added support for array types to type error checker.
- Added support for array access, variable reassignment and loops to type error checker.
- Added type checking for Source 3 library functions.
- Added tests for Source 3 Typed.

How to test:
-
1. Without frontend: Run Source 3 Typed in the js-slang REPL.
```
$ js-slang -c 3 -v typed # if js-slang is added to PATH
$ node dist/repl/repl.js -c 3 -v typed # if js-slang is not added to PATH
```
2. With frontend: [Link local js-slang to local frontend](https://github.com/source-academy/js-slang#using-your-js-slang-in-your-local-source-academy) and checkout local frontend to source-academy/frontend#2299.
3. Check the existing unit tests in `src/typeChecker/__tests__/source3Typed.test.ts`.
```
$ yarn test src/typeChecker/__tests__/source3Typed.test.ts
```

### DIFF
--- a/src/typeChecker/__tests__/source3Typed.test.ts
+++ b/src/typeChecker/__tests__/source3Typed.test.ts
@@ -25,6 +25,22 @@ describe('array type', () => {
       Line 4: Type '(number | string)[]' is not assignable to type 'string[]'."
     `)
   })
+
+  it('handles nested array types', () => {
+    const code = `const arr1: number[][] = [[1], [2], [3]];
+      const arr2: string[][] = [['1'], ['2'], ['3']];
+      const arr3: number[][] = [[1], ['2'], [3]];
+      const arr4: string[][] = [[1], ['2'], [3]];
+      const arr5: (number | string)[][] = [[1], ['2'], [3]];
+      const arr6: number[][] = [];
+    `
+
+    parse(code, context)
+    expect(parseError(context.errors)).toMatchInlineSnapshot(`
+      "Line 3: Type '(number[] | string[])[]' is not assignable to type 'number[][]'.
+      Line 4: Type '(number[] | string[])[]' is not assignable to type 'string[][]'."
+    `)
+  })
 })
 
 describe('array access', () => {
--- a/src/typeChecker/__tests__/source3Typed.test.ts
+++ b/src/typeChecker/__tests__/source3Typed.test.ts
@@ -1,0 +1,181 @@
+import { parseError } from '../..'
+import { mockContext } from '../../mocks/context'
+import { parse } from '../../parser/parser'
+import { Chapter, Variant } from '../../types'
+
+let context = mockContext(Chapter.SOURCE_3, Variant.TYPED)
+
+beforeEach(() => {
+  context = mockContext(Chapter.SOURCE_3, Variant.TYPED)
+})
+
+describe('array type', () => {
+  it('handles type mismatches correctly', () => {
+    const code = `const arr1: number[] = [1, 2, 3];
+      const arr2: string[] = ['1', '2', '3'];
+      const arr3: number[] = [1, '2', 3];
+      const arr4: string[] = [1, '2', 3];
+      const arr5: (number | string)[] = [1, '2', 3];
+      const arr6: number[] = [];
+    `
+
+    parse(code, context)
+    expect(parseError(context.errors)).toMatchInlineSnapshot(`
+      "Line 3: Type '(number | string)[]' is not assignable to type 'number[]'.
+      Line 4: Type '(number | string)[]' is not assignable to type 'string[]'."
+    `)
+  })
+})
+
+describe('array access', () => {
+  it('index must be number', () => {
+    const code = `const arr: number[] = [1, 2, 3];
+      arr[0];
+      arr['1'];
+      arr[true];
+      arr[undefined];
+      arr[null];
+    `
+
+    parse(code, context)
+    expect(parseError(context.errors)).toMatchInlineSnapshot(`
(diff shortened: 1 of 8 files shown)


src/typeChecker/__tests__/source3Typed.test.ts
```diff
@@ -0,0 +1,181 @@
...
+import { parse } from '../../parser/parser'
+import { Chapter, Variant } from '../../types'
+
+let context = mockContext(Chapter.SOURCE_3, Variant.TYPED)
+
+beforeEach(() => {
+  context = mockContext(Chapter.SOURCE_3, Variant.TYPED)
+})
+
+describe('array type', () => {
```
ianyong, line 12: What about more than 1-dimensional arrays? The current type system doesn't seem to work.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -208,23 +212,38 @@ function typeCheckAndReturnType(node: tsEs.Node, context: Context, env: TypeEnvi
       return typeCheckAndReturnArrowFunctionType(node, context, env)
     }
     case 'FunctionDeclaration':
-      // Only identifiers are used as function params in Source
-      const params = node.params as tsEs.Identifier[]
+      if (node.id === null) {
+        // Block should not be reached since node.id is only null when function declaration
+        // is part of `export default function`, which is not used in Source
+        throw new TypecheckError(node, 'Function declaration should always have an identifier')
+      }
```
ianyong, line 219: For documentation, this can be done because `FunctionDeclaration` AST nodes without a name are banned in https://github.com/source-academy/js-slang/commit/6ff73b19490239486aba533deaf609c5ccdfeabf, although it has not been merged into `master` yet as it is part of a larger PR to add local module imports.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -208,23 +212,38 @@ function typeCheckAndReturnType(node: tsEs.Node, context: Context, env: TypeEnvi
...
+
+      // Only identifiers/rest elements are used as function params in Source
+      const params = node.params as (tsEs.Identifier | tsEs.RestElement)[]
```
ianyong, line 222: Instead of type casting, is it possible to filter using type guards? You can then compare the length of the filtered array against that of the original array to throw an error if there is a mismatch (thus indicating that the invariant `Only identifiers/rest elements are used as function params in Source` is not true). This would help greatly in the event that this invariant is broken (whether on purpose or not) in the future.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -342,6 +384,39 @@ function typeCheckAndReturnType(node: tsEs.Node, context: Context, env: TypeEnvi
...
+        context.errors.push(new ConstNotAssignableTypeError(node, node.left.name))
+      }
+      checkForTypeMismatch(node, actualType, expectedType, context)
+      return tUndef
+    case 'ArrayExpression':
+      // Casting is safe here as Source disallows use of spread elements and holes in arrays
+      const elements = node.elements as Exclude<
+        tsEs.ArrayExpression['elements'][0],
+        tsEs.SpreadElement | null
+      >[]
```
ianyong, line 401: Same thing here.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -436,17 +535,26 @@ function addTypeDeclarationsToEnvironment(
           // is part of `export default function`, which is not used in Source
           throw new TypecheckError(node, 'Function declaration should always have an identifier')
         }
-        // Only identifiers are used as function params in Source
-        const params = node.params as tsEs.Identifier[]
+        // Only identifiers/rest elements are used as function params in Source
+        const params = node.params as (tsEs.Identifier | tsEs.RestElement)[]
```
ianyong, line 539: Here too.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -463,6 +571,7 @@ function addTypeDeclarationsToEnvironment(
         }
         const id = node.declarations[0].id as tsEs.Identifier
         const expectedType = getTypeAnnotationType(id.typeAnnotation, context, env)
+        // Type is not checked here as it will be checked in typeCheckAndReturnType
```
ianyong, line 574: What's this mean?

src/typeChecker/typeErrorChecker.ts
```diff
@@ -580,14 +689,21 @@ function typeCheckAndReturnArrowFunctionType(
   node: tsEs.ArrowFunctionExpression,
   context: Context,
   env: TypeEnvironment
-): FunctionType {
-  // Only identifiers are used as function params in Source
-  const params = node.params as tsEs.Identifier[]
+): Type {
+  // Only identifiers/rest elements are used as function params in Source
+  const params = node.params as (tsEs.Identifier | tsEs.RestElement)[]
```
ianyong, line 694: Here too.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: `addTypeDeclarationsToEnvironment` only deals with hoisting, so there are no type checks done. Will reword this comment accordingly.

src/typeChecker/__tests__/source3Typed.test.ts
zhaojj2209: Thanks for the catch! Will add support + tests for 2D+ arrays.

ianyong (approved):

PR #1330, opened 2023-01-05
Add Source 4 Typed variant; modify typechecking for + operator

zhaojj2209:
Adds support for Source 4 predeclared functions to Source Typed. The `parse` function is typed loosely for now; a separate PR will be made down the line to add support for Source 4 parse tree types. Tests and documentation will also only be added once the parse tree types are added.

Also included in the PR is a modification to the behavior of the `+` operator. The new behavior is as follows:
- If the expression on either side is `string` primitive/literal type, check that the other expression is a subtype of `string`, then return type `string`.
- If the expression on either side is `number` primitive/literal type, check that the other expression is a subtype of `number`, then return type `number`.
- If neither expression can be narrowed to either `string` or `number`, check that both sides are subtypes of `string | number`, then return type `any`.

Specs for Source Typed 1-3 have also been updated accordingly.

### DIFF
--- a/docs/specs/source_1_typed_typing.tex
+++ b/docs/specs/source_1_typed_typing.tex
@@ -76,13 +76,13 @@ type environment $\Gamma_0$ as follows:
 && \hspace{2mm} [ \texttt{typeof} \leftarrow \Any \rightarrow \String]\Gamma_{-2}
 \end{eqnarray*}
 
-The overloaded binary primitives are handled as follows:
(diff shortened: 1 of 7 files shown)


ianyong (approved):
Code changes look good (did not check documentation changes).

Should `src/typeChecker/__tests__/source4Typed.test.ts` be created though? At least the scaffolding, so that the lack of any new test cases for Source 4 Typed is explicit. It will also make it easier for test cases to be added in the future should the need arise.

martin-henz (approved):
Looks good. Rules for + to be discussed in more detail.

PR #1334, opened 2023-01-16
Modify + operator to prioritise left type in Source Typed

zhaojj2209:
Currently, in both Source Typed and Source, the `+` operator checks the right type against the left type if the left type is `string` or `number`. Therefore, it is expected for the type of the result to also prioritise left type over right type. However, the actual behaviour prioritises `string` over `number`. This PR corrects this inconsistency.

### DIFF
--- a/src/typeChecker/__tests__/source1Typed.test.ts
+++ b/src/typeChecker/__tests__/source1Typed.test.ts
@@ -655,7 +655,7 @@ describe('binary operations', () => {
       const x11: string = x3 + x2; // error, boolean + string, return type string
       const x12: number = x1 + x4; // no error, number + any, return type number
       const x13: string = x5 + x2; // no error, any + string, return type string
-      const x14: string = x1 + x2; // error, number + string, return type string
+      const x14: number = x1 + x2; // error, number + string, return type number (follows left side)
       const x15: string = x4 + x5; // no error, any + any, return type any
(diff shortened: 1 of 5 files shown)


src/typeChecker/__tests__/source1Typed.test.ts
```diff
@@ -655,7 +655,7 @@ describe('binary operations', () => {
       const x11: string = x3 + x2; // error, boolean + string, return type string
       const x12: number = x1 + x4; // no error, number + any, return type number
       const x13: string = x5 + x2; // no error, any + string, return type string
-      const x14: string = x1 + x2; // error, number + string, return type string
+      const x14: number = x1 + x2; // error, number + string, return type number (follows left side)
```
ianyong, line 658: Should the other lines that follow right side also have `(follows right side)` for consistency?

ianyong (commented):
Only reviewed the code changes.

src/typeChecker/__tests__/source1Typed.test.ts
zhaojj2209: The extra info is added because this one line is the special case where one side is `number` and the other side is `string`, in which case the left type will be used. The other cases are more straightforward (follow the side where the type is valid, else return `any`), so the extra info is not necessary.

ianyong (approved):

martin-henz (approved):
Great. As discussed.

PR #1335, opened 2023-01-16
Update file path to function name translation

ianyong:
Prof @martin-henz observed that with the current file path to function name translation, we would still have name collisions. For example, both `/a.js` and `/a/dot/js` would be translated into `__$a$dot$js__`.

To work around this, we impose an additional constraint on the file paths such that file paths should not contain consecutive slashes (`//`). This means that we can update the encoding of `.` to `$$dot$$` and `-` to `$$dash$$`, both of which cannot be obtained from file paths that do not themselves contain `.` or `-`.

On top of that, the underscore character (`_`) is now allowed in file paths, and is translated to function names as itself.

Part of https://github.com/source-academy/frontend/issues/2176. Related to #1332.

### DIFF
--- a/src/errors/localImportErrors.ts
+++ b/src/errors/localImportErrors.ts
@@ -2,22 +2,38 @@ import { UNKNOWN_LOCATION } from '../constants'
 import { nonAlphanumericCharEncoding } from '../localImports/filePaths'
 import { ErrorSeverity, ErrorType, SourceError } from '../types'
 
-export class InvalidFilePathError implements SourceError {
+export abstract class InvalidFilePathError implements SourceError {
   public type = ErrorType.TYPE
   public severity = ErrorSeverity.ERROR
   public location = UNKNOWN_LOCATION
 
   constructor(public filePath: string) {}
 
+  abstract explain(): string
+
+  abstract elaborate(): string
+}
+
+export class IllegalCharInFilePathError extends InvalidFilePathError {
   public explain() {
     const validNonAlphanumericChars = Object.keys(nonAlphanumericCharEncoding)
       .map(char => `'${char}'`)
       .join(', ')
-    return `'${this.filePath}' must only contain alphanumeric chars or one of ${validNonAlphanumericChars}, and must not contain consecutive slashes '//'.`
+    return `File path '${this.filePath}' must only contain alphanumeric chars and/or ${validNonAlphanumericChars}.`
   }
 
   public elaborate() {
-    return 'Rename the offending file path to only use valid chars and remove consecutive slashes.'
+    return 'Rename the offending file path to only use valid chars.'
+  }
+}
+
+export class ConsecutiveSlashesInFilePathError extends InvalidFilePathError {
+  public explain() {
+    return `File path '${this.filePath}' cannot contain consecutive slashes '//'.`
+  }
+
+  public elaborate() {
+    return 'Remove consecutive slashes from the offending file path.'
   }
 }
 
--- a/src/errors/localImportErrors.ts
+++ b/src/errors/localImportErrors.ts
@@ -1,5 +1,5 @@
 import { UNKNOWN_LOCATION } from '../constants'
-import { charEncoding } from '../localImports/filePaths'
+import { nonAlphanumericCharEncoding } from '../localImports/filePaths'
 import { ErrorSeverity, ErrorType, SourceError } from '../types'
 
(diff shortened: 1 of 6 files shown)


ianyong:
@zhaojj2209 Marking this PR as draft for now. I would like to get #1336 merged in first so that I can add the verbose versions of the files test cases.

src/errors/localImportErrors.ts
```diff
@@ -10,14 +10,14 @@ export class InvalidFilePathError implements SourceError {
   constructor(public filePath: string) {}
 
   public explain() {
-    const validNonAlphanumericChars = Object.keys(charEncoding)
+    const validNonAlphanumericChars = Object.keys(nonAlphanumericCharEncoding)
       .map(char => `'${char}'`)
       .join(', ')
-    return `'${this.filePath}' must only contain alphanumeric chars or one of ${validNonAlphanumericChars}.`
+    return `'${this.filePath}' must only contain alphanumeric chars or one of ${validNonAlphanumericChars}, and must not contain consecutive slashes '//'.`
```
zhaojj2209, line 16:
```suggestion
    return `File path '${this.filePath}' must only contain alphanumeric chars or one of ${validNonAlphanumericChars}, and must not contain consecutive slashes ('//').`
```
Minor rewording.
As an aside, we can consider splitting this into two separate errors, one for file path containing invalid characters and one for consecutive slashes, if it helps reduce the length of error messages/makes the error clearer to students.

ianyong: Wouldn't adding `File path` at the start make the length of the error message even longer? I'm not sure what value it has since it's rather obvious that the file path is a file path (nothing else would be named with slashes in them).

Regarding the addition of brackets, this means that the the error message will be something like `must only contain one of '_', '/', '.', '-', and must not contain consecutive slashes ('//')`. Is this inconsistency intended?

Splitting the error message would require having two different kinds of errors that deal with invalid file paths and moving the error handling logic into the file path validation function which isn't as clean. I agree that it makes for a better user experience though. Let me think about this more.

zhaojj2209: Adding `File path` to the error message makes it clearer that the rules apply to all file paths, so that students are less likely to make the same errors for other file paths. This also makes the error message read more similarly to other error messages in Source. (e.g. `Redeclaring name ${this.name}.` in `VariableRedeclaration`).

The inconsistency is intended as the former is referring to the characters directly but the latter describes the characters first, hence the brackets were added to indicate that the term "consecutive slashes" and the charactes `//` refer to the same thing. Perhaps changing it to `(e.g. '//')` would make it clearer?

Speaking of, `one of '_', '/', '.', '-',` can also be reworded to `the chars '_', '/', '.' or '-',`. Though this means there will be two "or"s in the sentence, it reduces the chances of misinterpreting the error message that the file path can only contain one of the special chars.

ianyong: > Adding `File path` to the error message makes it clearer that the rules apply to all file paths, so that students are less likely to make the same errors for other file paths. This also makes the error message read more similarly to other error messages in Source (e.g. `Redeclaring name ${this.name}.` in `VariableRedeclaration`).

Ok, that's fair.

> The inconsistency is intended as the former is referring to the characters directly but the latter describes the characters first, hence the brackets were added to indicate that the term "consecutive slashes" and the characters `//` refer to the same thing. Perhaps changing it to `(e.g. '//')` would make it clearer?

My understanding is that "e.g." is semantically the same as "for example". Are you sure about this?

> Speaking of, `one of '_', '/', '.', '-',` can also be reworded to `the chars '_', '/', '.' or '-',`. Though this means there will be two "or"s in the sentence, it reduces the chances of misinterpreting the error message that the file path can only contain one of the special chars.

This is slightly problematic as it requires special formatting instead of just calling `join` on the valid non-alphanumeric characters. But I guess I can implement a formatting function if it will make the user experience better.

zhaojj2209: Yeah on second thoughts let's not add the "e.g.". The rewording of `one of '_', '/', '.', '-',` is also not really needed. The initial concern was the inconsistency between the wording with respect to `'_', '/', '.', '-',` and `//`, but if the errors are to be split into two then this becomes less of a concern.

zhaojj2209 (approved):
LGTM, thanks for the changes!

PR #1338, opened 2023-01-18
Modify Source Typed to use success typing

zhaojj2209:
Modifies the Source Typed variant such type errors are thrown only when the intersection of the expected and actual type is the empty set.


martin-henz (approved):
Nice. Reaching a stable point now.

Issue #1341, opened 2023-01-28
Stepper: step limit off by one


martin-henz:
https://sourceacademy.org/playground#chap=1&prgrm=PTAEGUEkGEAVQFLlARgHQCY0oFA4GYCuAdgMYAuAlgPbGj4CGF1ATpQwDYAUxAlKAG8coEaBYBTcoRZ06AXgWpQw0atEB+VCrWqAXKDoAqek3Kt23OgFpUvANw4AvnkbM2nLgFZ7QA

This program takes exactly 1000 steps, but with a step limit of 1000, it gives "maximal number of steps exceeded":
<img width="765" alt="Screen Shot 2023-01-28 at 10 32 54 AM" src="https://user-images.githubusercontent.com/30220768/215237689-1d3e906b-9e5f-4d8c-b43e-f61b40e42f14.png">
Setting the step limit to 1001 succeeds:
<img width="765" alt="Screen Shot 2023-01-28 at 10 33 12 AM" src="https://user-images.githubusercontent.com/30220768/215237724-5dc74375-eb59-43e6-bd35-bfccecaca5ea.png">

martin-henz:
Fixed by https://github.com/source-academy/js-slang/pull/1342

PR #1342, opened 2023-01-29
Fixed stepper bug

joel-foo:
Resolved this issue: https://github.com/source-academy/js-slang/issues/1341

### DIFF
--- a/src/stepper/stepper.ts
+++ b/src/stepper/stepper.ts
@@ -2981,7 +2981,7 @@ export function getEvaluationSteps(
     let i = -1
     let stillEval = false
     while ((reducedWithPath[0] as es.Program).body.length > 0) {
-      if (steps.length == limit) {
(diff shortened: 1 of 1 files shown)


martin-henz (changes requested):
The choice of variable name `stillEval` isn't right. It should be `limitExceeded`.

I was considering using the test
```
(reducedWithPath[0] as es.Program).body.length === 0)
```
in line 3002 and not introduce the boolean variable at all, but having the boolean variable `limitExceeded` makes things clearer.

martin-henz:
Can you add a test case for this?

joel-foo:
Sure, I will amend it accordingly.

martin-henz (approved):

martin-henz:
fixes https://github.com/source-academy/js-slang/issues/1341

PR #1343, opened 2023-01-29
Refactor Source runner to operate on an AST instead of code

ianyong:
### Description

This is necessary because after the pre-processing step where non-Source module imports & exports are transformed, we end up with a single AST containing the locations of the original line & column numbers (for user-written code) for use in error messages. Originally, I attempted to transpile the AST back into code (to be compatible with the existing Source runner), but then the location information is lost.

Another benefit of operating on the AST is that all of the existing components, such as the infinite loop detector, continue to work for multiple file programs!

Please see the individual commits for the context behind why each change was necessary.

Part of https://github.com/source-academy/frontend/issues/2176. Related to #1332.

### How to Test

Since this is merely a refactoring PR, there should be no changes in behaviour whatsoever.

### DIFF
--- a/src/infiniteLoops/__tests__/runtime.ts
+++ b/src/infiniteLoops/__tests__/runtime.ts
@@ -72,7 +72,7 @@ const testForInfiniteLoopWithCode = (code: string, previousPrograms: es.Program[
   const context = createContext(Chapter.SOURCE_4, Variant.DEFAULT)
   const program = parse(code, context)
   if (program === undefined) {
-    return
+    throw new Error('Unable to parse code.')
   }
   return testForInfiniteLoop(program, previousPrograms)
 }
--- a/src/infiniteLoops/__tests__/runtime.ts
+++ b/src/infiniteLoops/__tests__/runtime.ts
@@ -1,7 +1,11 @@
+import * as es from 'estree'
+
 import { runInContext } from '../..'
+import createContext from '../../createContext'
 import { mockContext } from '../../mocks/context'
 import * as moduleLoader from '../../modules/moduleLoader'
-import { Chapter } from '../../types'
+import { parse } from '../../parser/parser'
+import { Chapter, Variant } from '../../types'
 import { stripIndent } from '../../utils/formatters'
 import { getInfiniteLoopData, InfiniteLoopError, InfiniteLoopErrorType } from '../errors'
 import { testForInfiniteLoop } from '../runtime'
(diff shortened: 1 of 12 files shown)


ianyong:
Note that this refactor changes the interface of the `Context` type. It will require a corresponding PR on the frontend when bumping js-slang version.

src/infiniteLoops/__tests__/runtime.ts
```diff
@@ -64,13 +68,22 @@ test('works in runInContext when throwInfiniteLoops is false', async () => {
   expect(result?.[1]).toBe(false)
 })
 
+const testForInfiniteLoopWithCode = (code: string, previousPrograms: es.Program[]) => {
+  const context = createContext(Chapter.SOURCE_4, Variant.DEFAULT)
+  const program = parse(code, context)
+  if (program === undefined) {
+    return
```
zhaojj2209, line 75: Given that `testForInfiniteLoop` can return `undefined`, it would be better to throw an error here, else there's a possibility that testcases with `undefined` as the expected result will pass even if the code fails the parse.

zhaojj2209 (approved):

src/runner/sourceRunner.ts
```diff
@@ -49,22 +48,19 @@ const DEFAULT_SOURCE_OPTIONS: IOptions = {
   throwInfiniteLoops: true
 }
 
-let previousCode = ''
+let previousCode: {
+  files: Partial<Record<string, string>>
+  entrypointFilePath: string
+} | null = null
 let isPreviousCodeTimeoutError = false
```
shenyih0ng, line 55: i'm wondering if we can make do without `previousCode`? it seems that it is only being used to check if the current program == previous program to derive whether we need to increase the timeout. perhaps it will be sufficient to have only a boolean flag to track if there was a previous timeout? i'm not sure if i'm missing out any other use cases for this

this may be unrelated to the refactoring in this PR but i thought i would bring this up for discussion still since this may simplify `sourceRunner`

ianyong: If I'm not wrong, we only want to increase the timeout if there was a previous timeout and the code didn't change? The second condition seems to be why we store the previous code instead of just a boolean flag.

But yes, I agree. I think that whether the timeout should be increased or not should simply be a flag in the context. The responsibility of setting this flag should be moved to the frontend. It is much easier to determine whether the code has changed or not there (and also makes changing the mechanism in the future easier). An even better alternative might be to allow specifying the timeout on the context itself, so that the frontend can decide how much time a program should be given to evaluate, with js-slang only responsible for evaluating the program as per the parameters passed in. This makes the timeout behaviour more apparent to users of js-slang.

PR #1345, opened 2023-01-30
Add test case for fixing stepper limit off by one issue

Ella-e:
Add a test case for issue #1342 (which fixes issue #1341: Step limit off by one).


martin-henz (changes requested):
Can you add another test case where the limit is 999 so that we see an error: step limit exceeded? That way we would have covered off-by-one errors in both directions.

Ella-e:
> Can you add another test case where the limit is 999 so that we see an error: step limit exceeded? That way we would have covered off-by-one errors in both directions.

Sure, I will add it later today.

Ella-e:
> Can you add another test case where the limit is 999 so that we see an error: step limit exceeded? That way we would have covered off-by-one errors in both directions.

I just noticed that the stepper will automatically increase the odd number limits by one (as the last even number step is only for the output of "Evaluation complete" output), so limit=999 will also work on our fixed version of stepper. Should I instead use limit=998 in writing the test case?

martin-henz:
So we are still off-by-one, now in the other direction. If the limit is 999, we should not be able to get a total of 1000 steps. I think you found another bug.

Ella-e:
> So we are still off-by-one, now in the other direction. If the limit is 999, we should not be able to get a total of 1000 steps. I think you found another bug.
Screenshot 2023-01-30 at 8 10 35 PM
The line that casts the limit number is line 2963 in stepper.ts file. I think based on the comments below (even steps are program before reduction and odd steps are program after reduction), all programs should have even steps, so the automaticlly casting might be reasonable? 

The UI of source academy currently do the casting by not showing the odd number steps. As shown in the screenshot below, if I set the limit to 999, the limit displayed on the right will become 1000 after pressing run. 
Screenshot 2023-01-30 at 8 20 15 PM
This might be quite confusing for beginners of stepper.

martin-henz:
Yes, the automatic casting is reasonable. To remove the confusion on the user side, do you think we could change the input field such that after entry of a n odd value, the number switches to the next even number? That would require a change in the frontend.

Ella-e:
> Yes, the automatic casting is reasonable. To remove the confusion on the user side, do you think we could change the input field such that after entry of a n odd value, the number switches to the next even number? That would require a change in the frontend.

If we are to preserve the current mechanism of stepper, I think it would be better to add instructions on how our stepper is working and auto-switch the odd number limits on the frontend. On the other hand, if we want users to be able to see every step, whether it is before or after reduction, we may need to change our current implementation stepper. I am not sure which one makes more sense and is in compliance with the idea of our stepper.

martin-henz:
Nice discussion. Let's talk about this at the next meeting and decide.

martin-henz (approved):
Nice. As discussed.

PR #1350, opened 2023-02-05
Add source file path to error messages for multi-file programs (non-transpiled)

ianyong:
## Description

The Acorn parser [provides an option `sourceFile` which adds a `source` attribute onto the `loc` object of every AST node](https://github.com/acornjs/acorn/tree/master/acorn/#interface). When we evaluate a program that consists of more than one file, we enable this option on the Acorn parser. Then, whenever we print error messages, if the `source` attribute exists, we also print out the full path of the file in front of the error message.

Note that this currently does not work for transpiled programs because they do not respect the `loc` object of the AST. There is some conversion being done before transpilation. I will deal with adding the source file path to errors messages for transpiled multi-file programs in another PR.

## Source Typed

For Source Typed, the Babel parser is used instead. While it does have [an equivalent `sourceFilename` option](https://babeljs.io/docs/en/babel-parser#options), the attribute is named `filename` instead of `source`. This leads to type errors, which we are forced to suppress. Ideally, the parser should always output ESTree-compliant ASTs (which isn't the case now). If this is not possible to achieve, then perhaps we will have to modify the ESTree types to allow for the `filename` attribute on the `loc` object.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -64,6 +64,7 @@ let verboseErrors: boolean = false
 
 export function parseError(errors: SourceError[], verbose: boolean = verboseErrors): string {
   const errorMessagesArr = errors.map(error => {
+    const filePath = error.location?.source ? `[${error.location.source}] ` : ''
     const line = error.location ? error.location.start.line : '<unknown>'
     const column = error.location ? error.location.start.column : '<unknown>'
     const explanation = error.explain()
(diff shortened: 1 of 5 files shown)


zhaojj2209 (approved):
LGTM!

Babel parser is powerful enough to parse all type syntax, but it's not extensible to modifications (e.g. no option to detect trailing commas/missing semicolons). Therefore it's unlikely that we can do much refactoring to the parser, so any refactoring would have to be done to the ESTree types instead.

ianyong:
> Babel parser is powerful enough to parse all type syntax, but it's not extensible to modifications (e.g. no option to detect trailing commas/missing semicolons). Therefore it's unlikely that we can do much refactoring to the parser, so any refactoring would have to be done to the ESTree types instead.

Perhaps an acceptable workaround is to walk through the AST generated by Babel parser and replace all instances of `filename` with `source`?

PR #1352, opened 2023-02-06
refactor: organize parsers better

shenyih0ng:
## Description

This PR is an effort to improve the code structure of parsers in`js-slang`. Hopefully this will allow better extensibility and smoother integration for upcoming `.py` and `.scm` parsers.

Currently we are working with 3 parsers (source, source-typed, fullJS), but all of the logic is concentrated in `parser/parser.ts`. As the more lang feats are added, it is becoming increasingly hard to trace the control/data flow of parsing a program.

## Changes

- Separate each parser implementation into individual modules/directories
- Separate parser tasks into `parse` and `validate`
    - `parse`: returns a AST + generic js syntax check
    - `validate`: language specific checks (source 1-4 restrictions) 
- Refactor `source` validation implementation ([`SourceParser::validate`](https://github.com/source-academy/js-slang/blob/5afe0dc46c00d0e2918eaafad2b06450bc533239/src/parsers/source/index.ts#L64))
- Instead of returning `null` instead of `undefined` if parse/validate fails
    - `null` seems capture the notion no AST being output than `undefined` 
    - _this is a little opinionated so do let me know what are your thoughts_

## TODOs

- [x] Test with `frontend`

## Next steps

- Better generalization for parser error recovery
- Better handling of babel and acorn AST types (i dont have a concrete plan for this but the overlapping type names cause confusions sometimes)


Fidget-Spinner (approved):
I took a brief look at the changes (I didn't look at the typed parser because I'm not familiar with that part of the code). This LGTM.

There's already something this PR fixes for us:
One trouble the py team has encountered is that we need to add custom error handlers for the different error type that our parser raises. (Our parser doesn't raise `SyntaxError extends SourceError`). Now that each parse action is split up, the error handling is a lot simpler.

Fidget-Spinner:
Thanks for the refactor! This should make our lives easier :).

ianyong:
Oops, I was in the middle of reviewing when the PR got merged. Mostly looks good except for some instances of type casting, but these instances presumably already existed in the codebase - just shifted around.

> Instead of returning `null` instead of `undefined` if parse/validate fails
> 
>     * `null` seems capture the notion no AST being output than `undefined`
> 
>     * _this is a little opinionated so do let me know what are your thoughts_

I wholeheartedly agree with this change. It definitely captures the semantics better.

Thank you!

martin-henz:
oh, sorry! @shenyih0ng can take a look at the two issues.

ianyong:
> oh, sorry! @shenyih0ng can take a look at the two issues.

Oh no, there are no (non-nitpicky) issues. Just wanted to show my support for this particular change that @shenyih0ng made.

PR #1353, opened 2023-02-08
Rename 'filename' attribute to 'source' in Babel parser-generated AST

ianyong:
Context of this change from #1350:
> ## Source Typed
>
> For Source Typed, the Babel parser is used instead. While it does have [an equivalent `sourceFilename` option](https://babeljs.io/docs/en/babel-parser#options), the attribute is named `filename` instead of `source`. This leads to type errors, which we are forced to suppress. Ideally, the parser should always output ESTree-compliant ASTs (which isn't the case now). If this is not possible to achieve, then perhaps we will have to modify the ESTree types to allow for the `filename` attribute on the `loc` object.

As a workaround, we recursively traverse the AST generated by the Babel parser and replace all instances of `filename` with `source`. This results in the AST returned from the `parse` function being compliant with the ESTree types.

Related to #1350. Part of https://github.com/source-academy/frontend/issues/2176.


martin-henz (approved):
Looks good. As discussed.

PR #1354, opened 2023-02-11
Add source file path to error messages for multi-file programs (transpiled)

ianyong:
## Description

Updated the transpiler to make use of the source file path when Source errors are thrown.

Related to #1350. Part of https://github.com/source-academy/frontend/issues/2176.

### DIFF
--- a/src/runner/errors.ts
+++ b/src/runner/errors.ts
@@ -101,6 +101,7 @@ export async function toSourceError(error: Error, sourceMap?: RawSourceMap): Pro
 
   let { line, column } = errorLocation
   let identifier: string = 'UNKNOWN'
+  let source: string | null = null
 
   if (sourceMap && !(line === -1 || column === -1)) {
     // Get original lines, column and identifier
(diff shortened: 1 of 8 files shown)


src/runner/errors.ts
```diff
@@ -112,16 +113,17 @@ export async function toSourceError(error: Error, sourceMap?: RawSourceMap): Pro
     line = originalPosition.line ?? -1 // use -1 in place of null
     column = originalPosition.column ?? -1
     identifier = originalPosition.name ?? identifier
+    source = originalPosition.source ?? null
```
ianyong, line 116: Note that this will currently always return `source` as the value because it is hardcoded as the file name for the `SourceMapGenerator`:
https://github.com/source-academy/js-slang/blob/cb0b853caedfbde72a9674f1353af8cbff2bcfbd/src/transpiler/transpiler.ts#L619
https://github.com/source-academy/js-slang/blob/cb0b853caedfbde72a9674f1353af8cbff2bcfbd/src/transpiler/transpiler.ts#L650

However, if everything is working well, we should not encounter any native errors.

martin-henz (approved):
Looks good!

PR #1360, opened 2023-02-20
Add parse tree types to Source 4 Typed

zhaojj2209:
Changes:
- Added support for handling type aliases with circular dependencies (necessary to support parse tree types).
- Added support for parse tree types.
- Added specifications for Source 4 Typed.
- Made adjustments to Source Typed specs.


src/typeChecker/parseTreeTypes.prelude.ts
```diff
@@ -0,0 +1,77 @@
...
+type ReturnStatement = Pair<"return_statement", Pair<Expression, null>>;
+type WhileLoop = Pair<"while_loop", Pair<Expression, Pair<Block, null>>>;
+type ForLoop = Pair<"for_loop", Pair<Expression | VariableDeclaration, Pair<Expression, Pair<Expression, Pair<Block, null>>>>>;
+type BreakStatement = Pair<"break_statement", null>;
+type ContinueStatement = Pair<"continue_statement", null>;
+
+// Parameters
+type Parameters = List<Name>;
+
+// If-Statement
```
ianyong, line 34: Do these comments do anything?

src/typeChecker/typeErrorChecker.ts
```diff
@@ -1182,53 +1374,46 @@ function lookupTypeAndRemoveForAllAndPredicateTypes(name: string): Type | undefi
...
-      !typeNode.typeParameters ||
-      typeNode.typeParameters.params.length !== type.typeParams.length
-    ) {
-      context.errors.push(
-        new InvalidNumberOfTypeArgumentsForGenericTypeError(typeNode, name, type.typeParams.length)
-      )
+  // If type saved in type environment is not generic,
+  // given variable type should not have type arguments
+  if (aliasType.kind !== 'forall') {
+    if (varType.typeArgs && varType.typeArgs.length > 0) {
```
ianyong, line 1387: Would it be possible to make the check more specific, instead of relying on falsey values?

src/typeChecker/utils.ts
```diff
@@ -129,7 +129,7 @@ export function formatTypeString(type: Type, formatAsLiteral?: boolean): string
         ? `(${elementTypeString})[]`
         : `${elementTypeString}[]`
     case 'variable':
-      if (type.typeArgs) {
+      if (type.typeArgs && type.typeArgs.length > 0) {
```
ianyong, line 132: Same here.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -1280,20 +1465,20 @@ function substituteVariableTypes(type: Type, typeVar: Variable, typeToSub: Type)
...
+function mergeTypes(node: tsEs.Node, ...types: Type[]): Type {
   const mergedTypes: Type[] = []
   for (const currType of types) {
     if (isEqual(currType, tAny)) {
       return tAny
     }
     if (currType.kind === 'union') {
       for (const type of currType.types) {
-        if (!containsType(mergedTypes, type)) {
+        if (!containsType(node, mergedTypes, type, [], [], true)) {
```
ianyong, line 1476: Why do we skip type alias expansion here?

src/typeChecker/typeErrorChecker.ts
```diff
@@ -1109,34 +1249,86 @@ function getAnnotatedType(typeNode: tsEs.TSType, context: Context): Type {
...
-function getParamTypes(params: (tsEs.Identifier | tsEs.RestElement)[], context: Context): Type[] {
-  return params.map(param => getTypeAnnotationType(param.typeAnnotation, context))
+function getHeadType(node: tsEs.Node, type: Type): Type {
+  switch (type.kind) {
+    case 'pair':
+      return type.headType
+    case 'list':
+      return type.elementType
+    case 'union':
+      return tUnion(...type.types.map(type => getHeadType(node, type)))
```
ianyong, line 1296: `type.types` seems weird. `type.constituentTypes`?

src/typeChecker/typeErrorChecker.ts
```diff
@@ -908,55 +931,80 @@ function hasTypeMismatchErrors(actualType: Type, expectedType: Type): boolean {
...
+      }
+      // Skips expansion, treat as type mismatch
+      if (skipTypeAliasExpansion) {
+        return true
+      }
+      // Pair and List should not be added to visited type aliases
+      // as they can be used repeatedly without causing circular dependencies
+      if (expectedType.name !== 'Pair' && expectedType.name !== 'List') {
+        visitedTypeAliasesForExpectedType.push(expectedType)
+      }
```
ianyong, line 997: Does this mean that I can't declare my own `Triple` type and use it like `Triple<number, string, Triple<number, string, boolean>>`?

ianyong (commented):
Some questions. Only checked the code and not the specifications.

src/typeChecker/parseTreeTypes.prelude.ts
zhaojj2209: The comments used to denote the sections that correspond to the BNF form in the Source 4 specs.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: Thanks for pointing this out! Will need to tighten the circular dependency check to take into account the type arguments as well, after which your stated example should be possible.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: `mergeTypes` is called when removing duplicate types from union types, in which case it is sufficient to compare the variable types at name level without expanding them.

ianyong: Hmm, would it be good to leave a comment here explaining why (coming from a long-term maintainability perspective)?

src/typeChecker/typeErrorChecker.ts
```diff
@@ -799,34 +789,49 @@ function typeCheckAndReturnArrowFunctionType(
  * Recurses through the two given types and returns an array of tuples
  * that map type variable names to the type to substitute.
  */
-function getTypeVariableMappings(actualType: Type, expectedType: Type): [string, Type][] {
+function getTypeVariableMappings(
+  node: tsEs.Node,
+  actualType: Type,
+  expectedType: Type
+): [string, Type][] {
+  // If type variable mapping is found, terminate early
```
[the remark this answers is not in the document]
zhaojj2209: If the type variable mapping is found, we don't need to expand the actual type as it can be done later during type checking.

src/typeChecker/typeErrorChecker.ts
zhaojj2209: On second thoughts, will be keeping the attribute name as `types` to be consistent with the naming in the `TSUnionType` node.

ianyong (approved):

PR #1361, opened 2023-02-21
Add explicit control evaluator

dickongwd:
# Explicit Control Evaluator
This explicit-control evaluator aims to replace the interpreter at `src/interpreter` used for visualising environments. This goes hand in hand with the pull requests [(1)](https://github.com/source-academy/frontend/pull/2350) and [(2)](https://github.com/source-academy/frontend/pull/2351) in the [frontend repo](https://github.com/source-academy/frontend).

## Motivation
The current interpreter has several drawbacks:
- Tail recursive (or iterative) processes may result in recursive processes in the interpreter. With explicit control, we aim to follow closely to the user's implementation.
- The interpreter currently only supports visualising environments at an instance. Ultimately, this evaluator will enable the creation of a modified environment visualiser to allow the user to step through (and back) each line of code and see the environment (and agenda and stash, see below).

## Description
To implement an evaluator with explicit control, we use two stacks: the Agenda and the Stash. The agenda holds nodes (user code) or instructions (specific to the evaluator), while the stash holds values and various results.
The program string is initially pushed into the agenda and items from the agenda is read and executed one by one, until the agenda is empty. At the end of the program, the result can be found in the stash.

## Changes
- The source code for the new evaluator is housed under `src/ec-evaluator`.
- Tests have been included in `src/ec-evaluator/__tests__` to ensure that the evaluator matches the current implementation of Source as closely as possible.
- A new variant has been added as `EXPLICIT_CONTROL` in `src/types.ts`.
- Source runner code in `src/runner/sourceRunner.ts` and the repl `src/repl/repl.ts` have been modified to add a new option of running with the new variant.
- Tests under `src/__tests__/inspect.ts` and its corresponding snapshots have been updated to match the output of the evaluator when used together with breakpoints and the environment visualiser.

## How to test
1. Running it in the repl by passing it as a variant:
```
node dist/repl/repl.js -c 4 -v explicit-control
```
2. With the frontend: [link local js-slang to local frontend](https://github.com/source-academy/js-slang#using-your-js-slang-in-your-local-source-academy) and select the explicit control variant. Be sure to change your frontend [branch](https://github.com/source-academy/frontend/pull/2351).
3. Run the tests:
```
yarn test --watch
```
Select `p` to filter and use `src/ec-evaluator/__tests__/*` as the pattern to match.


src/runner/utils.ts
```diff
@@ -50,6 +50,25 @@ export function determineExecutionMethod(
...
   } else if (areBreakpointsSet()) {
     isNativeRunnable = false
+  } else if (theOptions.executionMethod === 'auto') {
+    if (context.executionMethod === 'auto') {
+      if (verboseErrors) {
+        isNativeRunnable = false
+      } else if (areBreakpointsSet()) {
+        isNativeRunnable = false
+      } else {
+        let hasDeuggerStatement = false
```
martin-henz, line 60: Typo: should be "hasDebuggerStatement" and not "hasDeuggerStatement"

martin-henz (changes requested):
Very nice! Just one typo.

src/runner/utils.ts
martin-henz: This is still "hasDeuggerStatement"

martin-henz (changes requested):

martin-henz (approved):

PR #1363, opened 2023-02-24
Repl module "default_js_slang" function support

larrywang0701:
My module "repl" have a function called "default_js_slang" that allows students to run code in js-slang side through my module. This function need support from js-slang side to get the evaluation function and context in "sourceRunner.ts".
https://github.com/source-academy/modules/pull/188


martin-henz (approved):
Looks good.

PR #1364, opened 2023-02-24
Fix failing test cases for Env Viz

dickongwd:
The recent push of the new evaluator has made the test suite for the environment visualiser to fail on the frontend (https://github.com/source-academy/frontend/pull/2351). This pull request aims to resolve the failing test cases.

## Changes
Some functions may be created but not have a corresponding binding in the environment frame (e.g. function expressions in array literals). The explicit control evaluator is modified such that a dummy variable will be created when that occurs. This follows the old interpreter in `src/interpreter`.


martin-henz (approved):
Looks good.

PR #1366, opened 2023-03-02
Modify Source Typed specs

zhaojj2209:
Made Source Typed specifications more precise and in line with the intended Source Typed behaviour.


martin-henz (approved):
That's an improvement. (Need to re-read the document.)

PR #1368, opened 2023-03-07
Fix empty environments in envViz

### LINKED ISSUE #1367: Explicit-control evaluator: empty frame created
In the environment model, we avoid creating empty frames. However, this example
https://share.sourceacademy.org/bvsmi
gives rise to an empty frame in the environment visualizer:
Screen Shot 2023-03-07 at 5 10 34 PM

vansh284:
Fixes #1367.
Empty frames were being created by the environment visualizer due to functions pointing to the empty frames.

## Changes
Empty environments are no longer created by the ec-evaluator. If a block has no declarations, the environment will not be pushed.

This change also requires updating some of the envVisualiser snapshots at js-slang and on the frontend (refer to [#2371](https://github.com/source-academy/frontend/pull/2371)) which previously created empty frames.

### DIFF
--- a/src/ec-evaluator/interpreter.ts
+++ b/src/ec-evaluator/interpreter.ts
@@ -44,7 +44,6 @@ import {
   currentEnvironment,
   declareFunctionsAndVariables,
   defineVariable,
-  getNonEmptyEnv,
   getVariable,
   handleRuntimeError,
   handleSequence,
@@ -186,8 +185,10 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
   Program: function (command: es.BlockStatement, context: Context, agenda: Agenda, stash: Stash) {
     context.numberOfOuterEnvironments += 1
     const environment = createBlockEnvironment(context, 'programEnvironment')
-    pushEnvironment(context, environment)
-    declareFunctionsAndVariables(context, command)
+    // Push the environment only if it is non empty.
+    if (declareFunctionsAndVariables(context, command, environment)) {
+      pushEnvironment(context, environment)
+    }
     agenda.push(...handleSequence(command.body))
   },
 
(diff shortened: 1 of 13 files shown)


src/ec-evaluator/utils.ts
```diff
@@ -199,6 +199,29 @@ export const createBlockEnvironment = (
   }
 }
 
+/**
```
martin-henz, line 202: Wouldn't it be simpler to avoid creating empty frames in the first place? An environment extension should just take the given environment if there are no bindings to be added.

martin-henz (changes requested):

src/ec-evaluator/interpreter.ts
```diff
@@ -185,8 +185,10 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
   Program: function (command: es.BlockStatement, context: Context, agenda: Agenda, stash: Stash) {
     context.numberOfOuterEnvironments += 1
     const environment = createBlockEnvironment(context, 'programEnvironment')
-    pushEnvironment(context, environment)
-    declareFunctionsAndVariables(context, command)
+    // Push the environment only if it is non empty.
+    if (declareFunctionsAndVariables(context, command, environment)) {
+      pushEnvironment(context, environment)
```
martin-henz, line 190: no
`    context.numberOfOuterEnvironments += 1`
here?

src/ec-evaluator/interpreter.ts
```diff
@@ -195,8 +197,10 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
     agenda.push(instr.envInstr(currentEnvironment(context)))
 
     const environment = createBlockEnvironment(context, 'blockEnvironment')
-    pushEnvironment(context, environment)
-    declareFunctionsAndVariables(context, command)
+    // Push the environment only if it is non empty.
+    if (declareFunctionsAndVariables(context, command, environment)) {
+      pushEnvironment(context, environment)
```
martin-henz, line 202: no
    context.numberOfOuterEnvironments += 1
here?

martin-henz (changes requested):

martin-henz (approved):
Great!

PR #1369, opened 2023-03-07
Limit type alias declarations and import statements in Source Typed to top level

zhaojj2209:
Adds a check that throws an error if type alias declarations are found in block statements and not programs (top level), to be in line with Source Typed specifications.

Also fixes Source Typed parser such that syntax errors related to import statements can be caught.


ianyong (approved):

martin-henz (approved):
Looks good!

Issue #1372, opened 2023-03-10
Make js-slang development possible on Windows


Fidget-Spinner:
For the unfortunate few of us who develop on Windows, it would be nice to support js-slang here.

Currently there are a few obstacles:

1. `.sh` scripts are part of setup. 
Solution: Either inform the user to install MinGW shell (which comes with Git in Windows by default), or convert them to NodeJS/Python.
2. `.sh` scrips are part of pre-commit and pre-push.
Same solution as above.

PR #1373, opened 2023-03-10
Add Windows support to autocomplete.sh

Fidget-Spinner:
Part of work on #1372. This makes the script multi-platform compatible.

This does not complete the issue.

### DIFF
--- a/scripts/autocomplete.sh
+++ b/scripts/autocomplete.sh
@@ -10,9 +10,9 @@ CONST_DECL = "const"
 FUNC_DECL = "func"
 
 
-base_dir = "docs/source"
+base_dir = f"docs{os.path.sep}source"
 src_filename = "global.html"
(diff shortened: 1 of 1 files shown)


scripts/autocomplete.sh
```diff
@@ -76,7 +76,7 @@ def process_function(namespace, element):
 
 def process_dir_globals(target):
     infile = os.path.join(base_dir, target, src_filename)
-    with open(infile) as f:
+    with open(infile, encoding="utf-8") as f:
```
Fidget-Spinner, line 79: On Windows, the `open` encoding depends on locale, which means opening the file is a lottery.

This should become default in Python 3.15 [1]. However, we have to force it for now.

[1] See: https://peps.python.org/pep-0686/

martin-henz (approved):
Looks good!

PR #1376, opened 2023-03-21
Add Full TypeScript Chapter

zhaojj2209:
Adds the Full TypeScript chapter. The `FullTSParser` makes use of the [@ts-morph/bootstrap](https://github.com/dsherret/ts-morph/tree/latest/packages/bootstrap) library to access the TypeScript compiler API, which is run on the code string to retrieve compile-time errors. The errors are presented in the form of a single colour-coded string; regex matching is then used to retrieve the line numbers and format each error into a Source error.

If the code passes the compile check, it will be parsed by the Babel Parser into a Babel ESTree-compliant AST, and AST manipulation is performed so that any TS nodes are removed. The resulting AST is then passed into the existing `fullJSRunner` for evaluation.

Support for Source built-ins and libraries is added by prepending the declarations to the program code. This comes with some caveats:
- The built-in functions cannot be typed, as the functions are only accessed at runtime. Therefore, each declaration is replaced with a stand-in constant with type `any` and value `1`.
- The prelude for source 4 has many lines, which significantly slows down the runtime of the program. Therefore, functions declared in preludes are replaced in the same way as built-ins, to shorten code length.

Any errors related to import declarations are also ignored, as the TS Compiler API does not have access to information on Source Modules.

**How to test:**
Clone the corresponding branch on frontend, link to local js-slang, then proceed to run any TS code.

### DIFF
--- a/src/createContext.ts
+++ b/src/createContext.ts
@@ -430,7 +430,7 @@ const createContext = <T>(
   externalContext?: T,
   externalBuiltIns: CustomBuiltIns = defaultBuiltIns
 ): Context => {
-  if (chapter === Chapter.FULL_JS) {
+  if (chapter === Chapter.FULL_JS || chapter === Chapter.FULL_TS) {
     // fullJS will include all builtins and preludes of source 4
     return {
       ...createContext(
@@ -440,7 +440,7 @@ const createContext = <T>(
         externalContext,
         externalBuiltIns
       ),
-      chapter: Chapter.FULL_JS
+      chapter
     } as Context
   }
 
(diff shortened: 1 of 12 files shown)


src/createContext.ts
```diff
@@ -440,7 +440,7 @@ const createContext = <T>(
         externalContext,
         externalBuiltIns
       ),
-      chapter: Chapter.FULL_JS
+      chapter
```
martin-henz, line 443: good catch

martin-henz (approved):
Looks good. Note that the prelude needs to be written in a particular way for this to work in the future. Can we document this somewhere, maybe in a README in the folder where the prelude resides?

PR #1378, opened 2023-03-26
Use UNKNOWN_LOCATION if node location does not exist

ianyong:
This change is necessary for multiple file programs to work correctly in the frontend. This is because the AST nodes which are added by the preprocessor do not have a corresponding location (due to not being parsed from code in the first place). One of the instances of `.loc!` is being called (no idea which because it's almost impossible to trace without proper error handling), resulting in the evaluation of multiple-file programs dying.

We really should ban the use of the non-null assertion operator (`!`). Its only purpose is to let lazy programmers write code that bypass the type checker.

Part of https://github.com/source-academy/frontend/issues/2176.

### DIFF
--- a/src/ec-evaluator/interpreter.ts
+++ b/src/ec-evaluator/interpreter.ts
@@ -9,7 +9,7 @@
 import * as es from 'estree'
 import { uniqueId } from 'lodash'
 
-import * as constants from '../constants'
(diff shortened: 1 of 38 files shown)


zhaojj2209 (approved):
LGTM

PR #1382, opened 2023-03-27
Remove syntax highlighting for 'as'

ianyong:
It turns out that `as` is not a keyword in JavaScript. The following is a valid program:
```js
const as = 5;
```

Related to #1381. Part of https://github.com/source-academy/frontend/issues/2176.


zhaojj2209:
Could you help to remove the syntax highlighting of `as` for Source Typed as well? Lines 580-583 of this same file.

PR #1388, opened 2023-03-30
Make stdlib available to modules

### LINKED ISSUE #1221: Importing an undefined symbol does not throw error
```.ts
import { undefined_symbol } from 'curve';
```
The above code in source should throw an error, but currently does not.

### LINKED ISSUE #1242: Module functions' implementations not hidden when displayed
Native Source functions have their internals displayed as `[implementation hidden]` when they get passed into `display()`, however, this hiding does not happen for imported module functions.

leeyi45:
# stdlib
This PR is mainly aimed at allowing module code to import `js-slang` code directly. Modules can now write code as follows:
```ts
import { pair } from 'js-slang/dist/stdlib/list';
import { runInContext } from 'js-slang';
```
# Import Support for EC Evaluator
The EC evaluator is now able to support imports, though execution cannot be suspended while import declarations are processed.

# Issues to address
- Fix #1221: Imports will now be checked for undefined symbols unless `checkImport` is given as false
- Fix #1242: Source modules will now be wrapped by a new builtin, `wrapSourceModule` that will hide the implementations of module functions

This PR will not be backward compatible.

### DIFF
--- a/src/__tests__/environment.ts
+++ b/src/__tests__/environment.ts
@@ -1,6 +1,6 @@
 import { Program } from 'estree'
 
-import { evaluate } from '../interpreter/interpreter'
+import { evaluateProgram as evaluate } from '../interpreter/interpreter'
 import { mockContext } from '../mocks/context'
 import { parse } from '../parser/parser'
 import { Chapter } from '../types'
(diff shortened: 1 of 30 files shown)


martin-henz (approved):
Nice! Looks fine.

PR #1402, opened 2023-04-08
Add scm-slang to js-slang

s-kybound:
This PR adds scm-slang as a Git submodule in js-slang. Disregarding the representation of data to users, the logic of scm-slang is fully functional, and basic preludes/builtins from the scheme-base library have been included into each relevant chapter. 

This PR implements the scm-slang parser, new Scheme chapters: 1 - 4, Full (this chapter does not add anything at the moment, but can be implemented in future) as well as preludes for the Scheme language. 

scm-slang has been linted to prevent confluct with js-slang linting rules.

Changes to code outside scm-slang have been limited to reduce the "invasiveness" of the new language, however future work must be done to "reformat" error messages and textual representations of returned values given to end users as identifiers that are invalud in JS have been encoded. Currently, users may see these encoded values instead of their expected values in error messages and return values.


src/repl/repl.ts
```diff
@@ -76,6 +76,12 @@ function validChapterVariant(chapter: any, variant: any) {
   for (const lang of sourceLanguages) {
     if (lang.chapter === chapter && lang.variant === variant) return true
   }
+  for (const lang of scmLanguages) {
+    if (lang.chapter === chapter && lang.variant === variant) return true
+  }
+  if (chapter < 0) {
+    return true
+  }
```
Fidget-Spinner, line 84: I left this for easier debugging purposes. I think you can leave it out and it should improve the test coverage slightly.

s-kybound: Will do, thanks!

s-kybound:
The final representation of errors and data types has also been changed now to reflect their proper values.

src/createContext.ts
```diff
@@ -473,7 +637,7 @@ const createContext = <T>(
   externalContext?: T,
   externalBuiltIns: CustomBuiltIns = defaultBuiltIns
 ): Context => {
-  if (chapter === Chapter.FULL_JS || chapter === Chapter.FULL_TS) {
+  if (chapter === Chapter.FULL_JS || chapter === Chapter.FULL_TS || chapter === Chapter.PYTHON_1) {
```
Fidget-Spinner, line 640: Sorry this was an old change that I removed in the py-slang branch but I forgot to ask you to update your branch. This breaks py-slang, could you review and approve https://github.com/source-academy/js-slang/pull/1405 which fixes it please?

s-kybound: Will do.

PR #1403, opened 2023-04-12
Add Interactive Environment Visualiser

dickongwd:
# Interactive Environment Visualiser

The new interactive environment visualiser is a revamp of the current environment visualiser. It uses an explicit control evaluator as described in #1361, and allows users to step through the program to see changes to the environment. There is a corresponding pull request on the frontend https://github.com/source-academy/frontend/pull/2441.

## Description

Currently, the environment visualiser runs when the program encounters a breakpoint. In the new visualiser, the user has to open the environment tab before running to activate it.

The functionality of breakpoints is changed. They no longer break programs but serve as a point to fast-forward (or backward) the environment visualiser.

When the program is run for the first time in the environment visualiser, the evaluator first runs the program all the way to the end and counts the number of steps. Breakpoints are also recorded as breakpoint steps.

## Changes
- The explicit control evaluator in `src/ec-evaluator` is modified to count the number of steps.
- New variables have been added into the context runtime to track the number of steps.
- Tests under `src/__tests__/inspect.ts` are skipped because of the new implementation of breakpoints.

## Future Work
1. Using a more efficient and smarter way of getting the environment state. The current implementation reruns the program whenever the slider is shifted.
2. Display the agenda and stash stacks together with the environment visualiser.
3. Refactor the codebase to remove suspended states caused by breakpoints. Because of the new environment visualiser, the program only records breakpoints when we encounter them and allows the program to finish even if breakpoints are present.

### DIFF
--- a/src/__tests__/inspect.ts
+++ b/src/__tests__/inspect.ts
@@ -13,7 +13,9 @@ function flattenEnvironments(result: Result): Environment[] {
     .map((env: Environment) => ({ ...env, tail: null }))
 }
 
-test('debugger; statement basic test', () => {
+// Test suite skipped since functionality of debugger statements
+// has been changed for environment visualiser.
+xtest('debugger; statement basic test', () => {
   const code1 = `
   let a = 2;
   debugger;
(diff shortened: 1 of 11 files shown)


martin-henz (approved):
Looks good!

PR #1404, opened 2023-04-13
Fix for scm-slang conditionals

s-kybound:
Resolves a bug which rendered scheme conditional syntax unevaluable

image_2023-04-14_00-47-56

This code does not compile, as scheme conditionals without a "catch-all" else clause may return undefined as a value. This commit exposes undefined, NaN and Infinity as builtin values so that they can be defined and used by both the parser and end users.


Fidget-Spinner (approved):
LGTM.

Issue #1407, opened 2023-04-15
Scheme: numbers not distinguishable from strings


martin-henz:
<img width="699" alt="Screen Shot 2023-04-15 at 3 22 37 PM" src="https://user-images.githubusercontent.com/30220768/232195449-dd81b650-601e-4c26-ad03-3c389a9708f1.png">

s-kybound:
Apologies - this is due to an oversight in the current implementation of the `display` for the scheme languages, which will always output a string at the moment. Otherwise, the language can properly distinguish between different types of data. I'll get to fixing it soon.

PR #1409, opened 2023-04-16
updated scheme builtins

s-kybound:
- Resolves https://github.com/source-academy/js-slang/issues/1407 - display now uses the original SA display

- Implemented the scheme-cxr library


martin-henz (approved):
Looks fine. Yes, the c(d|a)*r functions are an important part of the Scheme tradition. A must-have in our §2+ sublanguages.

PR #1410, opened 2023-04-17
Add documentation for AUXILIARY functions

ianyong:
AUXILIARY functions are like MISC functions, except they are not intended to be directly called by the user (although they can be). Currently, there are two such functions, `__access_export__` and `__access_named_export__`, which are used for local modules.


martin-henz (approved):
Looks good!

PR #1411, opened 2023-04-20
Fix file path in error message when module cannot be found

ianyong:
Should be passing in the current file path and not the entrypoint file path.

### DIFF
--- a/src/localImports/__tests__/preprocessor.ts
+++ b/src/localImports/__tests__/preprocessor.ts
@@ -104,7 +104,7 @@ describe('preprocessFileImports', () => {
     expect(stripLocationInfo(actualProgram)).toEqual(stripLocationInfo(expectedProgram))
   }
 
-  it('returns undefined if the entrypoint file does not exist', () => {
+  it('returns undefined & adds CannotFindModuleError to context if the entrypoint file does not exist', () => {
     const files: Record<string, string> = {
       '/a.js': '1 + 2;'
     }
@@ -115,7 +115,7 @@ describe('preprocessFileImports', () => {
     )
   })
 
-  it('returns undefined if an imported file does not exist', () => {
+  it('returns undefined & adds CannotFindModuleError to context if an imported file does not exist', () => {
     const files: Record<string, string> = {
       '/a.js': `import { x } from './non-existent-file.js';`
     }
--- a/src/localImports/__tests__/preprocessor.ts
+++ b/src/localImports/__tests__/preprocessor.ts
@@ -110,6 +110,20 @@ describe('preprocessFileImports', () => {
     }
     const actualProgram = preprocessFileImports(files, '/non-existent-file.js', actualContext)
     expect(actualProgram).toBeUndefined()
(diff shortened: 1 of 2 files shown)


leeyi45:
Works as intended, but I think it might be wise to add some extra elaborations pointing out that file extensions are required for local imports given that up to that source imports don't need them.

ianyong:
> Works as intended, but I think it might be wise to add some extra elaborations pointing out that file extensions are required for local imports given that up to that source imports don't need them.

File extensions are not required for local imports. What is required is a relative path.

leeyi45:
> > Works as intended, but I think it might be wise to add some extra elaborations pointing out that file extensions are required for local imports given that up to that source imports don't need them.
> 
> File extensions are not required for local imports. What is required is a relative path (and even this is an arbitrary restriction that only serves to distinguish Source module imports from local module imports).
image
The current implementation requires file extensions

ianyong:
> > > Works as intended, but I think it might be wise to add some extra elaborations pointing out that file extensions are required for local imports given that up to that source imports don't need them.
> > 
> > 
> > File extensions are not required for local imports. What is required is a relative path (and even this is an arbitrary restriction that only serves to distinguish Source module imports from local module imports).
> 
> image The current implementation requires file extensions

You did not remove the file extension from your file. See https://share.sourceacademy.org/1u27s.

leeyi45:
Ah I see. Coming from the javascript background confused me. If we're not following the ESM resolution rules then I guess there's no issues.

leeyi45:
I noticed that when resolving absolute file paths, the current URL segment is also included, i.e. when using the playground `'/main` gets resolved to `/playground/main` (in fact the relative paths do too). Is this intended behaviour? Should we be strict about the meaning of `/` as the root directory?

ianyong:
> I noticed that when resolving absolute file paths, the current URL segment is also included, i.e. when using the playground `'/main` gets resolved to `/playground/main` (in fact the relative paths do too). Is this intended behaviour? Should we be strict about the meaning of `/` as the root directory?

The current URL segment is not included - the URL paths and the file system are entirely separate components/concepts. `/playground` is the directory corresponding to the playground. This is because each workspace has its own directory as defined here:
https://github.com/source-academy/frontend/blob/35446e40c3658cb33f37cc48e3fd0c28682c91b1/src/pages/fileSystem/createInBrowserFileSystem.ts#L11-L70
(Code snippet doesn't show because it's from the frontend while this PR is for js-slang.)

If we want to always use `/` for all workspaces, then there can only be a single Source program for every workspace which is not what we want. An alternative would be to instantiate a file system for every workspace, but that might make the performance of the Source Academy worse on weaker devices.

Either way, this design decision resides entirely in the frontend and has nothing to do with js-slang. You could for example, change the code that I linked such that `/` is used without having to change js-slang at all.

ianyong:
@RichDom2185 Sorry for requesting your review out of the blue, but I would like to get this fix merged in before the PR becomes stale. Thank you!

zhaojj2209 (approved):

PR #1413, opened 2023-05-02
Update py-slang and scm-slang subrepos

Fidget-Spinner:
This includes a critical bugfix for py-slang which crashes the frontend at the moment. I fixed that in https://github.com/source-academy/py-slang/commit/4afe35bab9ff6f30764d6572cb6a87d3df1a5799

To see the crash happening, go to https://sourceacademy.org/playground , switch to Python, and type ``1 is not 2`` into the code editor and try run that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "js-slang",
-  "version": "1.0.20",
+  "version": "1.0.23",
   "license": "Apache-2.0",
   "description": "Javascript-based implementations of Source, written in Typescript",
   "keywords": [
--- a/src/localImports/__tests__/preprocessor.ts
+++ b/src/localImports/__tests__/preprocessor.ts
@@ -104,12 +104,26 @@ describe('preprocessFileImports', () => {
     expect(stripLocationInfo(actualProgram)).toEqual(stripLocationInfo(expectedProgram))
   }
 
-  it('returns undefined if the entrypoint file does not exist', () => {
+  it('returns undefined & adds CannotFindModuleError to context if the entrypoint file does not exist', () => {
     const files: Record<string, string> = {
       '/a.js': '1 + 2;'
     }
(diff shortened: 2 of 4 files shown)


RichDom2185 (changes requested):
I'm blocking this for two reasons:

* Despite being a fix for a critical py-slang bug:
  * Python is not used yet and still in its experimental phase in Source Academy
  * I strongly feel #1414 is more critical. We need to settle our NPM version control, and depending on the results of that discussion, there may or may not be a force-push necessary.
* This one I feel might be more of a nit, but let me know what you think: By saying "This operator is not supported", I'm assuming it means it's not supported by the specific py-slang chapter. But the error message states "not yet", not "not". _Is there a distinction between stuff that we have not had time to add support for, versus something we don't plan to support?_ My view is:
  * If we don't plan to support it, we say "not", not "not yet"
  * Even if we plan to support it for the specific chapter, we should still say "not", instead of "not yet" -- and just add the support for the operator (change the list of "supported operators" in the language) later. I think this makes the error much cleaner, as now the functioning and behaviour of py-slang is independent of its development.

Fidget-Spinner:
"Not yet" is correct. The problem is that we have not yet decided how the translated Python code in SICPy should look like. If we follow more idiomatic Python (which is what we're currently leaning towards), we would want to implement these operators, otherwise these operators will be replaced with builtins, just like in Source.

RichDom2185:
> "Not yet" is correct. The problem is that we have not yet decided how the translated Python code in SICPy should look like. If we follow more idiomatic Python (which is what we're currently leaning towards), we would want to implement these operators, otherwise these operators will be replaced with builtins, just like in Source.

I see, thanks for letting me know of the problem. Seems like a long-term one we can only solve once SICPy is finalised. In that case then we can continue to merge after #1414. Thanks!

RichDom2185 (approved):
LGTM, thanks for the patience!

PR #1420, opened 2023-05-18
Automate publishing of js-slang to NPM

### LINKED ISSUE #1414: Missing commits of updating of versions: 1.0.21, 1.0.22, 1.0.23
Seems like the past few deployments to NPM (`1.0.21`, `1.0.22`, `1.0.23`) were not accompanied with the usual "Bump version" commit message. This is not good (I actually think it's quite critical considering we published these releases) as the "release notes"/changes between these three versions are thus unknown, despite being supposedly "patch" releases.

I can foresee this as a huge risk if we continue to deploy manually, and if more people start using our package.

I think we should set up a deploy to NPM GitHub action using a trigger workflow dispatch available to admins only to publish the latest version to NPM and add the version bump to `package.json` automatically. Perhaps we can also automatically generate a corresponding GitHub release; seems like the last one was `v0.3.6· back in 2020.

RichDom2185:
Use a GitHub Actions workflow to publish new versions of js-slang to npmjs.

---

Fixes #1414.

Logic: Whenever a PR is closed and merged onto `master`, if the package version in `package.json` is _changed_\*, GitHub Actions will request to run a workflow under the protected environment (configured in repo settings) called "publish-npm". If the workflow is approved by the specified maintainers, a new release on npmjs will be created.

**\*** _"Changed" is defined as not matching the latest version on NPM_

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -1,0 +1,52 @@
+name: Publish to npm
+
+on:
+  pull_request:
+    branches:
+      - master
+    types: [closed]
+
+jobs:
+  verify:
+    name: Verify should publish
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify PR merged
+        env:
+          IS_MERGED: ${{ github.event.pull_request.merged }}
+        run: |
+          if [ "$IS_MERGED" != "true" ]; then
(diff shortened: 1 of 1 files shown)


shenyih0ng (approved):
thanks for working on this 🔥

> reminder to self to update the token in environment as it's currently still empty

PR #1421, opened 2023-05-23
Bump version to v1.0.24

RichDom2185:
Changelog: https://github.com/source-academy/js-slang/compare/d04dd404d0552b620d8c52eb5c4b066234deb21e...b152558ffffde15788b5e6b0a2fe2b757485448e

Also serves to test our new CI pipeline for deploying to NPM automatically.

PR #1422, opened 2023-05-25
bumping version

martin-henz:
trying out the new workflow


RichDom2185:
Hi Prof, unfortunately the workflow is still broken at the moment. We tried it out back in #1421 and it failed, I'll close this PR first in the meantime.

martin-henz:
OK

PR #1423, opened 2023-05-25
Fix workflow for publishing to NPM

RichDom2185:
* Change the workflow trigger from PR merge to `master` push
* Do not fail the workflow when GH Actions is not supposed to publish a new version; instead, just skip the publishing step after verification.


shenyih0ng:
found this workflow that we can use: https://github.com/JS-DevTools/npm-publish, it seems to do what we want to

Example workflow (publish package + tag commit):

```yaml
name: Publish to npm

on:
  push:
    branches: main

jobs:
  publish:
    runs-on: ubuntu-latest
    environment: publish-npm # Await approval from maintainers
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-node@v3
        with:
          node-version: "16"

      - name: Publish to NPM
        uses: JS-DevTools/npm-publish@v2
        with:
          token: ${{ secrets.NPM_TOKEN }}
          dry-run: true # = npm publish --dry-run

      - name: Tag Release  # tag commit with new version (corresponds to npm version) 
        run: |
          git tag ${{ steps.publish.outputs.version }}
          git push origin ${{github.run_id}}
```

martin-henz:
OK, let's try this.

PR #1425, opened 2023-05-28
Prevent bypassing default export & import ban

### LINKED ISSUE #1424: Ban on default exports & imports can be bypassed
Default exports & imports are banned in Source 1 thru 4 by disallowing `ExportDefaultDeclaration` & `ImportDefaultSpecifier` AST nodes respectively. This is because even though default exports & imports are implemented (and can be used in the library parser language), we do not wish to add unnecessary complexity to Source 1 thru 4.

For example, the following code results in the AST containing an `ExportDefaultDeclaration` node:
```js
export default function square(x) {
    return x * x;
}
```
The error message is `Line 1: Export default declarations are not allowed`.

While the following code results in the AST containing an `ImportDefaultSpecifier` node:
```js
import x from './a.js';
```
The error message is `Line 1: Import default specifiers are not allowed`.

We can however bypass these AST node checks by writing the same code in a slightly different way that results in an AST without `ExportDefaultDeclaration` & `ImportDefaultSpecifier` nodes:
```js
// /a.js
function square(x) {
    return x * x;
}

export { square as default };

// /b.js (entrypoint)
import { default as x } from './a.js';

x(5);
```

The code in `/a.js` is parsed into an `ExportNamedDeclaration` AST node while the code in `/b.js` is parsed into an `ImportSpecifier` AST node despite the program semantics being the same. Because [`default` is a keyword in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words), we treat the name `default` specially. Specifically, importing/exporting names as `default` simply makes use of the default import/export functionality:
https://github.com/source-academy/js-slang/blob/7cc42d051c9e4059d86275abb0961249ae679d98/src/stdlib/localImport.prelude.ts#L2-L30

ianyong:
Add new rules to prevent the default export & import ban from being bypassed.

Resolves #1424.

### DIFF
--- a/src/parser/__tests__/allowed-syntax.ts
+++ b/src/parser/__tests__/allowed-syntax.ts
@@ -315,6 +315,16 @@ test.each([
     export default x;
     x;
     `
+  ],
+
+  [
+    Chapter.LIBRARY_PARSER,
+    `
+    function square(x) {
+      return x * x;
+    }
+    export { square as default };
+    `
   ]
 ] as [Chapter, string][])(
   'Syntaxes are allowed in the chapter they are introduced %#',
--- a/src/parser/__tests__/allowed-syntax.ts
+++ b/src/parser/__tests__/allowed-syntax.ts
@@ -331,17 +331,28 @@ test.each([
     Chapter.LIBRARY_PARSER,
     `
     import { default as x } from './a.js';
-    `
+    `,
+    // Skip the success tests because the imported file does not exist.
+    true
   ]
-] as [Chapter, string][])(
+] as [Chapter, string, boolean | undefined][])(
   'Syntaxes are allowed in the chapter they are introduced %#',
-  (chapter: Chapter, snippet: string) => {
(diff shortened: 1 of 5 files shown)


src/parser/__tests__/allowed-syntax.ts
```diff
@@ -315,16 +315,44 @@ test.each([
...
+    `
+  ],
+
+  [
+    Chapter.LIBRARY_PARSER,
+    `
+    import { default as x } from './a.js';
+    `,
+    // Skip the success tests because the imported file does not exist.
+    true
```
ianyong, line 336: Not ideal but there is no way around the `CannotFindModuleError` unless the entire test abstraction here is rewritten. Either way, it is more important to test that the rule is active for Source 1 thru 4 than it is to test that the rule is inactive for the library parser.

zhaojj2209 (approved):

ianyong:
Deploy error should be fixed by #1423.

Issue #1348, opened 2023-02-01
Use asynchronous GET requests / Replace `xmlhttprequest-ts` with `fetch` API for module loading


Yongbeom-Kim:
Continued from #1340 

# Problem
- Currently, `js-slang` uses the `xmlhttprequest-ts` package to make GET requests to retrieve module data.
- The [package](https://www.npmjs.com/package/xmlhttprequest-ts) has not had any updates/releases in the past 5 years, and it seems safe to say that this package should be deprecated in favour of something like `fetch` instead, as suggested by a contributor(?) https://github.com/hmoog/xmlhttprequest-ts/issues/6
- It's also worthy of noting that synchronous/blocking XMLHttpRequests (as is being done in js-slang) are [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#synchronous_request)
- Well, it makes sense that the get request should be asynchronous/non-blocking?

# Proposal
- Refactor js-slang to load modules asynchronously (with async functions and promises)
- Replace all XMLHttpRequest GET requests with `fetch` requests
- Use top-level await to load modules (if needed)
  - I don't think this is a big issue, noting that top-level await is [supported on all browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#browser_compatibility)
  - This also requires bumping tsconfig.json to:
```
{
  "compilerOptions": {
    "module": "esnext" (or system),
    "target": "es2017",
}
```

leeyi45:
As part of this, I'm thinking we delegate loading of tabs to the frontend. Tabs are a frontend only concept, so I don't think it makes sense for js-slang to do the loading.

RichDom2185:
I would say this is an issue of high enough priority; I will pin it to the top of our issues list.

leeyi45:
This transition is actually almost complete. Most of the features surrounding module loading are asynchronous, namely
- Loading the module manifest
- Module Documentation
- Transpiler
- Stepper

The two remaining areas of concern are:
- The EC-Evaluator: Refer to (my comment on this)[https://github.com/source-academy/js-slang/pull/1428#issuecomment-1662237400]
- The regular interpreter, which is due to be removed as per #1476
Both of which still require the synchronous versions

PR #1428, opened 2023-06-22
Further Enhancements to the Module System

### LINKED ISSUE #1416: Incorrect Error for ExportNamedDeclarations with sources
In folder mode:
```
// /program.js
import { show, heart } from './other.js';
show(heart);

// /other.js
export { show, heart } from 'rune';
```
leads to the error: `Line 5: ReferenceError: Cannot access 'show' before initialization`. If this type of export is unsupported, then this error message is not indicative of that fact.

Code [here](https://share.sourceacademy.org/o9quh)

leeyi45:
# Notable Changes
## `export from` resolution
When using `export from` declarations, the system will now resolve the symbol through the directives to where the symbol was defined:
```
// a.js
export const a = "a";

// b.js
export { a as b } from './a.js';

// c.js
import { b as c } from './c.js'
// gets translated to
import { a as c } from './a.js'
```
This change may cause snapshots that use import directives to produce different import statements when preprocessed.
Should fix #1416 by introducing support for such export declarations

## Asynchronous Imports
Addresses #1348 partially by introducing a new set of functions that use `fetch` to retrieve module files. However, the interpreter does not support asynchronous operation, so the original synchronous module loading code has been left in.

## Import Preprocessor
The import preprocessor has been changed to treat local and source imports more uniformly:

- The preprocessor now supports resolving and checking for undefined symbols for local modules as well
- Though `export *` statements are not currently supported by other language features, the preprocessor has the ability to detect reexported symbols should such statements become available in the future
- Options for module resolution have been added, allowing specifications like `import a from './a'` to be resolved to `import a from './a/index.js'`, similar to how Node and browsers resolve modules. This should help work toward the js-slang repl being able to execute multiple files

## Assertions
Introduce a new`AssertionError` type that derives from `RuntimeSourceError` to gracefully handle unexpected conditions that arise during program execution, along side an `assert` function. Using a custom function allows us to modify the errors assertions throw, and removes the need for polyfills for the NodeJS assert module to work with the browser.

# Minor Changes
## Execution Options
- As part of an effort to make debugging js-slang easier, several options to display debugging information have been added to execution such as `logTranspilerOutput` have been added
- Options controlling how imports are resolved have also been added. For example, js-slang can be configured to avoid loading tabs, which would come in handy especially for execution contexts that can't display tabs

## Refactoring of AST utils
- Code relating to manipulating an AST have been collected under `src/utils/ast`. Where possible, type guards that were spread out across different parts of js-slang have been consolidated.
- `import * as es from './utils/ast/types'` can be used to replace `import * as es from 'estree'`. This will provide extra AST node types and reduce the amount of type checking needed when dealing with an AST. All types are compatible with `estree`.


ianyong:
> ## `export from` resolution
> 
> When using `export from` declarations, the system will now resolve the symbol through the directives to where the symbol was defined:
> 
> ```
> // a.js
> export const a = "a";
> 
> // b.js
> export { a as b } from './a.js';
> 
> // c.js
> import { b as c } from './c.js'
> // gets translated to
> import { a as c } from './a.js'
> ```
> 
> This change may cause snapshots that use import directives to produce different import statements when preprocessed. Should fix #1416 by introducing support for such export declarations

Is this something we want to support? I believe in prior discussions with Prof @martin-henz, we did not want to support re-exporting of names in Source (or at least not directly). What this means is that while the following program is not supported:
```js
// a.js
export const a = 123;

// b.js
export { a as b } from './a.js';

// c.js
import { b } from './b.js';
b;
```

The following program is allowed:
```js
// a.js
export const a = 123;

// b.js
import { a as b } from './a.js';
export { b };

// b.js (equivalent alternative)
import { a } from './a.js';
export { a as b };

// c.js
import { b } from './b.js';
b;
```

The main difference is that the name (either `a` or `b` depending on when the renaming is done) is available in the scope of `b.js` in the second program, but not the first program. In the second program, we are not directly re-exporting names from another module, but importing names into `b.js`'s environment before exporting them.

The reason why we might want such a behaviour is so as to be able to provide a simple mental model for students. While the solution you proposed works (at least in theory), I think it adds non-trivial complexity to the module import mental model because how an import statement is preprocessed now depends on the contents of other modules.

@martin-henz If we do decide to ban re-exports, I can work on it since this is something I should've done earlier but missed out on. >.<

package.json
```diff
@@ -33,7 +33,7 @@
   "dependencies": {
```
shenyih0ng, line 33: bump `@types/node` (dev dep) to `20.4.4` to solve CI error?

src/utils/ast/astUtils.ts
```diff
@@ -0,0 +1,106 @@
+import assert from '../assert'
```
ianyong, line 1: A bit weird to have a file `astUtils` in the path `src/utils/ast/`.

src/utils/ast/astUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+        return processors.withFunction(node.declaration as es.FunctionDeclarationWithId)
+      case 'ClassDeclaration':
+        return processors.withClass(node.declaration as es.ClassDeclarationWithId)
+    }
+  } else if (node.source) {
+    return processors.withSource(node as es.ExportNamedDeclarationWithSource)
+  } else {
+    return processors.localExports(node as es.ExportNamedLocalDeclaration)
+  }
+}
```
ianyong: How do you know that all of these type assertions are safe? One huge issue I faced while working on my FYP was the use of (incorrect) type assertions and non-null assertion operators everywhere that bypassed type safety. It meant that while the application could compile, I had a lot of trouble fixing the runtime errors that occurred due to incorrect types because I cannot possibly know where errors occur until I run the program.

From a maintainability point of view, this is a blocker for me.

src/utils/ast/types.ts
```diff
@@ -0,0 +1,73 @@
...
+    source: es.Literal
+  }
+>
+
+export type ModuleDeclarationWithSource =
+  | es.ImportDeclaration
+  | es.ExportAllDeclaration
+  | ExportNamedDeclarationWithSource
+
+export * from 'estree'
```
ianyong, line 73: These types are okay if they are being used for AST nodes that we create ourselves. However, it seems that there are a lot of type assertions being done on AST nodes that originate from the AST parser (see one of my comments above). What guarantees do you have that these types are correct? And if you do have such guarantees, would it be better to fix the types upstream (in `estree`) instead?

ianyong (commented):
I've not reviewed the entire PR because it's too big, even after spending a couple hours. Ideally, this should have been split into many smaller PRs since there are many different changes being made. Anyway, there are some very fundamental issues (mostly relating to the approach of some of the solutions, as well as regarding typing) that need to be addressed.

tsconfig.json
```diff
@@ -29,6 +29,9 @@
   "exclude": [
     "src/stdlib/metacircular-interpreter",
     "src/stdlib/**/*.js",
+    "src/**/__tests__/**",
+    "src/**/__mocks__/**",
+    "jest.*.*s",
```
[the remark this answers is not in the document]
leeyi45: The type checking will be done by jest when tests are run

If the tests are included in tsconfig.json, they get compiled to javascript with the rest of the js-slang code. This causes issues for the module mocks

package.json
leeyi45: On my system, updating `@types/node` produces this error when running `yarn test`:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/leeyi45/source/source_academy/js-slang/node_modules/wrap-ansi/index.js from /Users/leeyi45/source/source_academy/js-slang/node_modules/cliui/build/index.cjs not supported.
Instead change the require of index.js in /Users/leeyi45/source/source_academy/js-slang/node_modules/cliui/build/index.cjs to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/leeyi45/source/source_academy/js-slang/node_modules/cliui/build/index.cjs:293:14)
    at Object.<anonymous> (/Users/leeyi45/source/source_academy/js-slang/node_modules/yargs/build/index.cjs:1:60678)
    at Object.<anonymous> (/Users/leeyi45/source/source_academy/js-slang/node_modules/yargs/index.cjs:5:30)
    at _yargs (/Users/leeyi45/source/source_academy/js-slang/node_modules/jest-cli/build/run.js:30:39)
    at buildArgv (/Users/leeyi45/source/source_academy/js-slang/node_modules/jest-cli/build/run.js:143:26)
    at Object.run (/Users/leeyi45/source/source_academy/js-slang/node_modules/jest-cli/build/run.js:118:24)
    at Object.<anonymous> (/Users/leeyi45/source/source_academy/js-slang/node_modules/jest-cli/bin/jest.js:16:17)
    at Object.<anonymous> (/Users/leeyi45/source/source_academy/js-slang/node_modules/jest/bin/jest.js:12:3)
```

jest.setup.ts
```diff
@@ -0,0 +1,16 @@
...
+  memoize: jest.fn((x: any) => x),
+}))
+
+jest.mock('./src/modules/moduleLoaderAsync')
+jest.mock('./src/modules/moduleLoader')
+
+// @ts-ignore
+SourceMapConsumer.initialize({
+  'lib/mappings.wasm': 'https://unpkg.com/source-map@0.7.3/lib/mappings.wasm'
+})
```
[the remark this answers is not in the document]
leeyi45: When running tests on my system, without this code in `jest.setup.ts`  causes `   TypeError: Cannot read properties of undefined (reading 'then')` when running certain tests

leeyi45:
> > ## `export from` resolution
> > When using `export from` declarations, the system will now resolve the symbol through the directives to where the symbol was defined:
> > ```
> > // a.js
> > export const a = "a";
> > 
> > // b.js
> > export { a as b } from './a.js';
> > 
> > // c.js
> > import { b as c } from './c.js'
> > // gets translated to
> > import { a as c } from './a.js'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This change may cause snapshots that use import directives to produce different import statements when preprocessed. Should fix #1416 by introducing support for such export declarations
> 
> Is this something we want to support? I believe in prior discussions with Prof @martin-henz, we did not want to support re-exporting of names in Source (or at least not directly). What this means is that while the following program is not supported:
> 
> ```js
> // a.js
> export const a = 123;
> 
> // b.js
> export { a as b } from './a.js';
> 
> // c.js
> import { b } from './b.js';
> b;
> ```
> 
> The following program is allowed:
> 
> ```js
> // a.js
> export const a = 123;
> 
> // b.js
> import { a as b } from './a.js';
> export { b };
> 
> // b.js (equivalent alternative)
> import { a } from './a.js';
> export { a as b };
> 
> // c.js
> import { b } from './b.js';
> b;
> ```
> 
> The main difference is that the name (either `a` or `b` depending on when the renaming is done) is available in the scope of `b.js` in the second program, but not the first program. In the second program, we are not directly re-exporting names from another module, but importing names into `b.js`'s environment before exporting them.
> 
> The reason why we might want such a behaviour is so as to be able to provide a simple mental model for students. While the solution you proposed works (at least in theory), I think it adds non-trivial complexity to the module import mental model because how an import statement is preprocessed now depends on the contents of other modules.
> 
> @martin-henz If we do decide to ban re-exports, I can work on it since this is something I should've done earlier but missed out on. >.<

I wrote this code before @martin-henz indicated that such exports should be banned in Source. This would require some work on my part to rewrite some tests that make use of this syntax

tsconfig.json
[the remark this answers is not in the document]
ianyong: Ah okay, thanks! If `ts-jest` does perform type checking, then there's no issue here.

> I'm not well versed enough with tsconfig to know if there is a way to exclude the test files from the build output, or how to make global module mocks work with the test files included in the build output

A common practice would be to have different tsconfig files. For example, we could have a `tsconfig.json` that we use for type checking (`tsc --noEmit`) and a `tsconfig.build.json` that extends `tsconfig.json` but excludes the test files that we use for building (`tsc --project tsconfig.build.json`). Currently it doesn't seem like we need to take on this additional complexity though.

jest.setup.ts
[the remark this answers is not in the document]
ianyong: Hmm, I can't replicate this locally. Logically, this shouldn't be needed in a Node.js environment.

martin-henz:
> > > ## `export from` resolution
> > > When using `export from` declarations, the system will now resolve the symbol through the directives to where the symbol was defined:
> > > ```
> > > // a.js
> > > export const a = "a";
> > > 
> > > // b.js
> > > export { a as b } from './a.js';
> > > 
> > > // c.js
> > > import { b as c } from './c.js'
> > > // gets translated to
> > > import { a as c } from './a.js'
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This change may cause snapshots that use import directives to produce different import statements when preprocessed. Should fix #1416 by introducing support for such export declarations
> > 
> > 
> > Is this something we want to support? I believe in prior discussions with Prof @martin-henz, we did not want to support re-exporting of names in Source (or at least not directly). What this means is that while the following program is not supported:
> > ```js
> > // a.js
> > export const a = 123;
> > 
> > // b.js
> > export { a as b } from './a.js';
> > 
> > // c.js
> > import { b } from './b.js';
> > b;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The following program is allowed:
> > ```js
> > // a.js
> > export const a = 123;
> > 
> > // b.js
> > import { a as b } from './a.js';
> > export { b };
> > 
> > // b.js (equivalent alternative)
> > import { a } from './a.js';
> > export { a as b };
> > 
> > // c.js
> > import { b } from './b.js';
> > b;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The main difference is that the name (either `a` or `b` depending on when the renaming is done) is available in the scope of `b.js` in the second program, but not the first program. In the second program, we are not directly re-exporting names from another module, but importing names into `b.js`'s environment before exporting them.
> > The reason why we might want such a behaviour is so as to be able to provide a simple mental model for students. While the solution you proposed works (at least in theory), I think it adds non-trivial complexity to the module import mental model because how an import statement is preprocessed now depends on the contents of other modules.
> > @martin-henz If we do decide to ban re-exports, I can work on it since this is something I should've done earlier but missed out on. >.<
> 
> I wrote this code before @martin-henz indicated that such exports should be banned in Source. This would require some work on my part to rewrite some tests that make use of this syntax

Yes, I think a restriction along these lines will keep a simple mental model and implementation.

src/utils/ast/astUtils.ts
leeyi45: I actually encountered the same problems you did, which is why I created these functions. The different variations of each type of export node weren't being properly handled, so to keep to DRY and make the handing more robust I added these functions.

For the two you've highlighted here, if you follow the code flow, they account for every single variant (accounted for by Typescript) of both the `ExportDefaultDeclaration` and `ExportNamedDeclaration`.

jest.setup.ts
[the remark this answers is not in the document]
leeyi45: Agreed, I just don't really know where the issue is, or how to resolve it at the moment

ianyong:
> * Options for module resolution have been added, allowing specifications like `import a from './a'` to be resolved to `import a from './a/index.js'`, similar to how Node and browsers resolve modules. This should help work toward the js-slang repl being able to execute multiple files

Actually, how does adding support for index files help work towards multiple file execution? My understanding is that this is entirely a Node.js thing, browsers do not resolve modules like this and this is not part of the ECMAScript standard. Even other JavaScript runtimes like Deno [do not support index files](https://deno.land/manual@v1.35.2/references/contributing/style_guide#do-not-use-the-filename-indextsindexjs). For what it's worth, the creator of Node.js, Ryan Dahl also feels that [index files are a mistake](https://youtu.be/M3BM9TB-8yA?t=880) due to needlessly complicating the module loading system.

Unless the goal is to be able to run code that also runs in a Node.js environment, I don't see why we should support this. You do not need this for multiple file execution.

src/utils/ast/astUtils.ts
ianyong: For the 2 type assertions above, could you make use of [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) which contain only the relevant code that lets us know that the type assertion is safe. There are 2 reasons for this:
1. We avoid using type assertions with the long term goal of eventually banning them via a lint rule (unless suppressed along with an explanation).
1. By encapsulating all of the logic that helps us check if a type conversion is valid within a function, it is easier to verify, and we can use the type predicate in other places of the codebase.

The rest of the type assertions within the code snippet (within `processExportNamedDeclaration`) look unsafe to me though.

leeyi45:
I have since taken several of @ianyong's suggestions into account and modified this PR appropriately, the last remaining issue is to solve the error that comes up when running `jest`. Then I will be able to move forward with unit testing.

leeyi45:
As a comment for the EC-Evaluator team @dickongwd @vansh284, would there be an issue if the Program command evaluator were written as part of the `ECEvaluate` function? Programs should not contain other programs as children if I understand correctly, so there's no need to write it as a separate evaluator:

```ts
function ECEvaluate(...) {
  for (const node of program.body) {
    // Evaluate imports first
  }
  for (const node of program.body) {
    // Run command evaluators
  }  
}
```
This makes it easier to pass options to the the code that handles module imports, and also removes the need to make every evaluator async.

vansh284:
> As a comment for the EC-Evaluator team @dickongwd @vansh284, would there be an issue if the Program command evaluator were written as part of the `ECEvaluate` function? Programs should not contain other programs as children if I understand correctly, so there's no need to write it as a separate evaluator:
> 
> ```ts
> function ECEvaluate(...) {
>   for (const node of program.body) {
>     // Evaluate imports first
>   }
>   for (const node of program.body) {
>     // Run command evaluators
>   }  
> }
> ```
> 
> This makes it easier to pass options to the the code that handles module imports, and also removes the need to make every evaluator async.

You are right in saying that Programs wouldn't contain other programs as children, so the final evaluation would still be the same.

However, (I'm not sure if you are aware) the 'agenda' in the ec-evaluator is visualised in the frontend as part of the Env Visualizer. The command evaluators act on each of the 'agenda' items, so if we have no command evaluator for the Program, it cannot be pushed on the 'agenda'.

So if instead of having a Program command evaluator, we directly evaluate the program at the start, the students wouldn't be able to see the first step of the visualization (which is the Program item on the agenda) as it would be skipped.

Maybe you can let us know exactly what your goal is, so we can think of a way to make it work with the ec-evaluator.

leeyi45:
> > As a comment for the EC-Evaluator team @dickongwd @vansh284, would there be an issue if the Program command evaluator were written as part of the `ECEvaluate` function? Programs should not contain other programs as children if I understand correctly, so there's no need to write it as a separate evaluator:
> > ```ts
> > function ECEvaluate(...) {
> >   for (const node of program.body) {
> >     // Evaluate imports first
> >   }
> >   for (const node of program.body) {
> >     // Run command evaluators
> >   }  
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This makes it easier to pass options to the the code that handles module imports, and also removes the need to make every evaluator async.
> 
> You are right in saying that Programs wouldn't contain other programs as children, so the final evaluation would still be the same.
> 
> However, (I'm not sure if you are aware) the 'agenda' in the ec-evaluator is visualised in the frontend as part of the Env Visualizer. The command evaluators act on each of the 'agenda' items, so if we have no command evaluator for the Program, it cannot be pushed on the 'agenda'.
> 
> So if instead of having a Program command evaluator, we directly evaluate the program at the start, the students wouldn't be able to see the first step of the visualization (which is the Program item on the agenda) as it would be skipped.
> 
> Maybe you can let us know exactly what your goal is, so we can think of a way to make it work with the ec-evaluator.

Perhaps I could refer to the implementation within the stepper:
```ts
async function evaluateImports(
  program: es.Program,
  context: Context,
  { loadTabs, wrapModules }: ImportTransformOptions
) {
  let importNodes: es.ImportDeclaration[]
  ;[importNodes, program.body] = partition(program.body, importDeclarationFilter)

  try {
    const importNodeMap = arrayMapFrom(
      importNodes.map(node => {
        const moduleName = node.source.value
        assert(
          typeof moduleName === 'string',
          `ImportDeclarations should have string sources, got ${moduleName}`
        )
        return [moduleName, node]
      })
    )

    const environment = currentEnvironment(context)
    await importNodeMap.forEachAsync(async (moduleName, nodes) => {
      await initModuleContextAsync(moduleName, context, loadTabs, nodes[0])
      const functions = await loadModuleBundleAsync(moduleName, context, wrapModules, nodes[0])

      for (const node of nodes) {
        for (const spec of node.specifiers) {
          let importedName: string
          switch (spec.type) {
            case 'ImportSpecifier': {
              importedName = spec.imported.name
              break
            }
            case 'ImportDefaultSpecifier': {
              importedName = 'default'
              break
            }
            case 'ImportNamespaceSpecifier': {
              throw new Error('Namespace imports are not supported!')
            }
          }

          declareIdentifier(context, spec.local.name, node, environment)
          defineVariable(context, spec.local.name, functions[importedName], true, node)
        }
      }
    })
  } catch (error) {
    // console.log(error)
    handleRuntimeError(context, error)
  }
}
```
I am able to access the `loadTabs` and `wrapModules` options. Notice that `evaluateImports` is also an async function. For the EC-Evaluator to be able to make use of async module loading code, the code that evaluates `Program` nodes needs to also be async.

leeyi45:
Due to the circumstances, I was unable to work on this. Given the large number of changes that have been merged since I will reopen this with a fresh branch

PR #1431, opened 2023-06-26
fix: missing functions in Source Native variants

### LINKED ISSUE #1429: Source X Native: Some functions are not available
Screen Shot 2023-06-23 at 7 12 51 PM

where X is 2-4. (Source 1 Native shouldn't have reverse)

shenyih0ng:
Fixes #1429 

## Description

Currently, when a user references certain functions in Source Native variants a `<identifier> not undeclared` error will be raised. These functions are mostly from the `list` and `stream` libraries.

The error occurs due to a bug in the runtime context of Source Native variants where functions from libraries are not properly keep tracked of. More specifically, the context only kept track of builtin functions and not those that injected via library preludes. This results in a failed check for undefined variables during the transpilation stage which raised the `<identifier> not undeclared` error back to the frontend.

## The fix

- Keep track of prelude identifier during first run of a Source Native context
- Add previously declared identifier to the context after each run

### DIFF
--- a/src/runner/__tests__/runners.ts
+++ b/src/runner/__tests__/runners.ts
@@ -179,6 +179,37 @@ describe('Additional JavaScript features are not available in Source Native', ()
   )
 })
 
+describe('Functions in Source libraries (e.g. list, streams) are available in Source Native', () => {
+  test('List functions are present in Source Native', async () => {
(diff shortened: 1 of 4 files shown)


martin-henz (approved):
Looks good!

PR #1436, opened 2023-07-11
Add simplifications to explicit control evaluator

dickongwd:
# Simplifications of the explicit control evaluator

## Description
This pull request presents simplifications to the explicit control evaluator.

## Changes
- Top-level `push undefined if needed` instructions are not pushed anymore, but `undefined` is pushed immediately to the stash. Addresses https://github.com/source-academy/frontend/issues/2536
- `ENVIRONMENT` instruction will not be created by block statements if there is another `ENVIRONMENT` instruction already on the agenda. There might be more cases where the `ENVIRONMENT` instruction is unnecessary. Addresses https://github.com/source-academy/frontend/issues/2538
- Expression statements will be fast forward (so the next step would not just remove `;`). Programs with only one statement will also be fast forwarded. Addresses https://github.com/source-academy/frontend/issues/2535
- Issues involving inconsistent steps fixed. Addresses https://github.com/source-academy/frontend/issues/2539
- Redundant `ENVIRONMENT` instructions removed for nullary functions and function bodies with no declarations. Addresses https://github.com/source-academy/frontend/issues/2542
- Interpreter now only adds `return undefined` to function bodies when needed.

### DIFF
--- a/src/ec-evaluator/__tests__/ec-evaluator-errors.ts
+++ b/src/ec-evaluator/__tests__/ec-evaluator-errors.ts
@@ -151,7 +151,7 @@ test('Infinite recursion with a block bodied function', () => {
     function i(n) {
       return n === 0 ? 0 : 1 + i(n-1);
     }
-    i(25000);
+    i(30000);
   `,
     optionEC4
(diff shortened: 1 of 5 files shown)


martin-henz (approved):
Looks good!

PR #1447, opened 2023-07-19
Fix env viz - program environment not reflecting values that are imported from modules

### LINKED ISSUE #1434: Env Viz: Program environment does not reflect values that are imported from modules
Screen Shot 2023-07-05 at 3 58 50 PM

Here the program environment should reflect the names show, heart, and stack, in addition to stack_them.

The name “stack” should be allocated in the program environment. Currently, the function “stack_them” magically knows how to apply the imported “stack” function, which is not good.

tsinyee:
Fixes #1434

Import modules are now handled whenever an 'ImportDeclaration' node is encountered rather than handling all of them together before the other nodes


martin-henz (approved):
Looks good!

PR #1451, opened 2023-07-21
Fix env viz import issues

### LINKED ISSUE #1448: Env Viz: Import statements handled inconsistently
Example:
https://share.sourceacademy.org/bqb59

Screen Shot 2023-07-19 at 6 44 23 PM

The step above makes sense: import declarations are hoisted to the beginning.

Screen Shot 2023-07-19 at 6 46 22 PM

Why is there a POP after the first import, but not after the second?

Screen Shot 2023-07-19 at 6 46 33 PM

This step and the next steps are not leading to any change.

Screen Shot 2023-07-19 at 6 46 39 PM

Screen Shot 2023-07-19 at 6 46 46 PM

But now suddenly:

Screen Shot 2023-07-19 at 6 46 56 PM

Stepping back from the last step shows states that are inconsistent with the forward steps.

### LINKED ISSUE #1450: Env Viz: order of declaration in Program environment
Screen Shot 2023-07-19 at 6 59 53 PM

Here the names x and y are displayed right away in the Program environment, but the imported names are gradually added. I think more consistent would be if we show all toplevel names right away in the program environment. For this, we need to make sure that the function that scans out the names declared in a block will catch the functions in the import declarations.

tsinyee:
ImportDeclarations are now handled in the Program evaluator in cmdEvaluators so that all toplevel names are shown right away in the program environment - Fixes #1450
ImportDeclarations are now non value producing - Fixes #1448 (POP issue)
Screenshot 2023-07-21 at 7 46 53 PM

Added a condition in runECEMachine to check if the command is an ImportDeclaration, then the current step is not incremented - Fixes #1448  (redundant steps issue but this feels quite hackish? )
Screenshot 2023-07-21 at 7 47 13 PM

### DIFF
--- a/src/ec-evaluator/interpreter.ts
+++ b/src/ec-evaluator/interpreter.ts
@@ -218,7 +218,6 @@ function runECEMachine(context: Context, agenda: Agenda, stash: Stash, isPrelude
 
   let command = agenda.peek()
 
-
   // First node will be a Program
   context.runtime.nodes.unshift(command as es.Program)
 
@@ -297,7 +296,6 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
       pushEnvironment(context, environment)
       evaluateImports(command as unknown as es.Program, context, true, true)
       declareFunctionsAndVariables(context, command, environment)
-
     }
 
     if (command.body.length == 1) {
--- a/src/ec-evaluator/interpreter.ts
+++ b/src/ec-evaluator/interpreter.ts
@@ -222,10 +222,6 @@ function runECEMachine(context: Context, agenda: Agenda, stash: Stash, isPrelude
   context.runtime.nodes.unshift(command as es.Program)
 
   while (command) {
-    let isImportDeclaration = false
-    if (isNode(command) && command.type === 'ImportDeclaration') {
-      isImportDeclaration = true
-    }
     // Return to capture a snapshot of the agenda and stash after the target step count is reached
     if (!isPrelude && steps === context.runtime.envSteps) {
       return stash.peek()
@@ -262,10 +258,8 @@ function runECEMachine(context: Context, agenda: Agenda, stash: Stash, isPrelude
       stash.push(undefined)
     }
     command = agenda.peek()
-    // If command is an ImportDeclaration, don't add unnecessary step
-    if (!isImportDeclaration) {
-      steps += 1
-    }
+
+    steps += 1
   }
 
   if (!isPrelude) {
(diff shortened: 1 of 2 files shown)


src/ec-evaluator/interpreter.ts
```diff
@@ -249,9 +261,11 @@ function runECEMachine(context: Context, agenda: Agenda, stash: Stash, isPrelude
     if (agenda.isEmpty() && stash.isEmpty()) {
       stash.push(undefined)
     }
-
     command = agenda.peek()
-    steps += 1
+    // If command is an ImportDeclaration, don't add unnecessary step
+    if (!isImportDeclaration) {
```
vansh284, line 266: I agree that this seems a bit hacky, and also feels a bit like its going against the concept of the agenda. Another problem with this is that at step 2 (in the screenshot) the user thinks that the import statement is being evaluated, but in step 3, another command is evaluated instead.

vansh284: Suggestion: In 'Program' when the block is decomposed into individual statements and pushed onto the agenda, don't push any import statements since the imports are already evaluated in 'Program'. (I would think modifying the 'handleSequence' function appropriately would work best). Then we wouldn't need this hack since our evaluator won't encounter any import statements.

vansh284: Rest looks good to me!

vansh284 (changes requested):

src/ec-evaluator/interpreter.ts
tsinyee: > Suggestion: In 'Program' when the block is decomposed into individual statements and pushed onto the agenda, don't push any import statements since the imports are already evaluated in 'Program'. (I would think modifying the 'handleSequence' function appropriately would work best). Then we wouldn't need this hack since our evaluator won't encounter any import statements.

this means that the import statements won't be in the agenda at all?

vansh284: Yep! The user can see the import statements initially in the program block, but since the import statements are evaluated in the program block itself it doesn't seem to make sense to continue to push the import statements on the agenda since they do nothing and only add to confusion. 
Do you agree prof? @martin-henz

martin-henz: Yes, I agree. The handling of the program needs to create the program environment which will include the imported names and the newly declared names. The imported names can be immediately assigned to their implementation without the need of executing a step in the machine.

tsinyee: okay, i have made the necessary changes to the handleSequence function so the import statements are only in the program block but not on the agenda

Screenshot 2023-07-27 at 2 39 03 PM

Screenshot 2023-07-27 at 2 39 43 PM

vansh284 (approved):
LGTM!

PR #1453, opened 2023-07-24
Enable source module imports in stepper

Ziwen510:
## Description

This pull request integrates source module imports within the stepper. The module objects manifest in their REPL string representation in the stepper. Any generated functions in the middle of the process are displayed as `[Function]`.

The test cases are commented out before mock module functions become available.

### DIFF
--- a/src/runner/sourceRunner.ts
+++ b/src/runner/sourceRunner.ts
@@ -86,6 +86,9 @@ function runSubstitution(
   options: IOptions
 ): Promise<Result> {
   const steps = getEvaluationSteps(program, context, options.stepLimit)
+  if (context.errors.length > 0) {
+    return resolvedErrorPromise
+  }
   const redexedSteps: IStepperPropContents[] = []
   for (const step of steps) {
     const redex = getRedex(step[0], step[1])


.husky/pre-push
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn jsdoc && yarn autocomplete && yarn format:ci && yarn eslint && yarn test
+# yarn jsdoc && yarn autocomplete && yarn format:ci && yarn eslint && yarn test
```
ianyong, line 4: This shouldn't be commented out.

src/utils/astCreator.ts
```diff
@@ -91,10 +91,11 @@ export const blockArrowFunction = (
 export const functionExpression = (
   params: es.Pattern[],
   body: es.Statement[] | es.BlockStatement,
-  loc?: es.SourceLocation | null
+  loc?: es.SourceLocation | null,
+  id?: es.Identifier
 ): es.FunctionExpression => ({
   type: 'FunctionExpression',
-  id: null,
+  id: id || null,
```
ianyong, line 98:
```suggestion
  id: id ?? null,
```

Minor nit: The [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) better captures the semantics here instead of `||` which is checking if `id` is a [falsy value](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). For objects, this distinction typically doesn't matter, but it does for primitive types. For example, 0 or the empty string could be valid values, but are falsy (and thus treated as if they were invalid). Still, it's good to always be as intentional as possible and the semantics of `??` is preferable here.

src/stepper/stepper.ts
```diff
@@ -2455,9 +2479,168 @@ function treeifyMain(target: substituterNodes): substituterNodes {
...
+function jsTreeifyMain(target: substituterNodes, visited: Set<substituterNodes>): substituterNodes {
+  // recurse down the program like substitute
+  // if see a function at expression position,
+  //   visited before recursing to this target: replace with the name
+  //   else: replace with a CallExpression "(() => FunctionExpression)()" where the FunctionExpression's body is treeified
+  let verboseCount = 0
+  const treeifiers = {
+    // Identifier: return
+    ExpressionStatement: (target: es.ExpressionStatement): es.ExpressionStatement => {
+      return ast.expressionStatement(treeify(target.expression) as es.Expression)
```
ianyong, line 2491: As much as possible, let's try not to use type assertions. They defeat the point of having a type system by silencing the compiler on type errors. Type assertions only affect the static semantics of a program but not the dynamic semantics, meaning that even though the TypeScript compiler doesn't complain, your program could run into runtime type errors. Instead, we should make use of [type narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html).

I know the current codebase isn't in a very good state (there are type assertions and non-null assertion operators everywhere), but let's try not to add onto the mess.

ianyong (commented):
This PR is missing a description and test cases, so I haven't fully reviewed it because I lack context of what is going on here.

I have some code quality comments from a quick look through, with the use of type assertions being one that might take a bit of effort to resolve.

Ziwen510:
### Description

This pull request integrate source module imports within the stepper. The module objects manifest in their REPL string representation in the stepper. Any generated functions in the middle of the process are displayed as `[Function]`.

The test cases are commented out for now before mock module functions become available.

src/stepper/stepper.ts
Ziwen510: This does require some effort. I open an issue for this so that it can hopefully be fully resolved. [issue #1457 ](https://github.com/source-academy/js-slang/issues/1457)

martin-henz:
As discussed, let's handle the refactoring later, see #1457.

martin-henz (approved):
Looks good.

PR #1458, opened 2023-07-30
Fix: not displaying runtime errors in stepper

Ziwen510:
## Description
Address the issue where the stepper is not generating any runtime errors.


martin-henz (approved):
Looks good!

PR #1460, opened 2023-08-07
Fix stepper: infinite loop in findMain

Ziwen510:
This pull request resolves the issue of breakdowns while executing substitutions in the stepper.


martin-henz (approved):
Looks good, as discussed.

Issue #1399, opened 2023-04-04
Add support for module types in Source Typed


zhaojj2209:
Currently, all names imported from modules default to the `any` type (see below code snippet, specifically line 563):
https://github.com/source-academy/js-slang/blob/0dd4d8c67a8ec132682d10343f2526a74f5f8ce5/src/typeChecker/typeErrorChecker.ts#L543-L566

In order to support typing of modules, we would need to design a system that allows for the importing of module types.

Possible implementation: every module should include a type declaration file containing a mapping of names to types. When encountering import statements, the type error checker can then retrieve the types from the type declaration file and conduct type checking using those types. If the type declaration file does not exist/any errors are encountered, the type error checker can default back to using the `any` type for all names in the module.

PR #1467, opened 2023-08-12
Add proof-of-concept for handling module types in Source Typed

zhaojj2209:
Part of #1399

This PR adds a basic implementation of module type handling in Source Typed, using the `rune` module as an example. Source Typed is now able to properly handle types for runes, as well as throw errors when attempting to import a name that does not exist in the module.

Summary of implementation:
- A map of names available in the runes module to their respective type information is created (see `runeTypes.ts`). The map also contains an extra `prelude` attribute, which contains the types shared by variables and functions in the module (in this case, `Rune` and `AnimatedRune`). The type information in the map are in the form of Source Typed code strings (constant/function/type alias declarations) which can be evaluated by the Source Typed type checker. To hide the actual implementation of the module, the code in the map are placeholder code that only serves to save the correct types in the type environment when the code is evaluated. The types for the variables/functions are retrieved from the [rune module documentation](https://source-academy.github.io/modules/documentation/modules/rune.html).
- When encountering import statements that import from the `rune` module, the Source Typed type checker will fetch the prelude code + code for the names imported for that module, and evaluate the code as a program string. Following evaluation, the correct types for the module names will be saved in the type environment for use in the rest of the program.

Sample screenshots:
Screenshot 2023-08-12 at 14 37 58
Screenshot 2023-08-12 at 14 36 56

To test this PR, link local js-slang to local frontend and write runes module code in any of the Source Typed chapters. Some unit tests have also been written.

To expand this proof-of-concept to an implementation that is extensible to all modules, the following needs to be done:
- The map of module names to types (`runeTypes.ts`) should be shifted to the modules repo. All future module type maps should be written in the modules repo instead and made available for fetching (e.g. as JSON files).
- The `TODO: Add logic for fetching module type declarations map from modules repo` code should be replaced with logic that fetches the module maps from the module repo as and when needed.

When writing module type maps, note the following:
- The map must contain a `prelude` attribute that contains code for the types shared by variables and functions in the module. If there are no shared types, an empty string should be provided.
- Code in the map should be placeholder code that hides the actual module implementation and only provides the correct types needed.
- The module map must contain all names in the module, else the `NameNotFoundInModule` error will be thrown for names that are not in the module map.
- Types in the map must be consistent with the [modules documentation](https://source-academy.github.io/modules/documentation/modules.html).

Limitations in the current approach:
- Since the placeholder value for runes is the string `'Rune'`, type errors will not be thrown if attempting to assign type `string` to the rune. This is an unfortunate result of hiding the module implementation using Source primitive values. One possible solution would be to add special placeholder type for modules, however this will likely require more extensive changes to Source Typed, and is currently out of scope for this PR.

### DIFF
--- a/src/errors/typeErrors.ts
+++ b/src/errors/typeErrors.ts
@@ -727,3 +727,26 @@ export class DuplicateTypeAliasError implements SourceError {
     return this.explain()
   }
 }
+
+export class NameNotFoundInModuleError implements SourceError {
+  public type = ErrorType.TYPE
+  public severity = ErrorSeverity.ERROR
+
+  constructor(
+    public node: tsEs.ImportDeclaration,
+    public moduleName: string,
+    public name: string
+  ) {}
+
+  get location() {
+    return this.node.loc ?? UNKNOWN_LOCATION
+  }
+
+  public explain() {
+    return `Module '${this.moduleName}' has no exported member '${this.name}'.`
+  }
+
+  public elaborate() {
+    return this.explain()
+  }
+}
--- a/src/typeChecker/__tests__/source1Typed.test.ts
+++ b/src/typeChecker/__tests__/source1Typed.test.ts
@@ -1016,8 +1016,8 @@ describe('import statements', () => {
     )
   })
 
-  it('defaults to any for all imports', () => {
-    const code = `import { show, heart } from 'rune';
+  it('throws error if module does not exist', () => {
+    const code = `import { show, heart } from 'doesnotexist';
       show(heart);
       heart(show);
       const x1: string = heart;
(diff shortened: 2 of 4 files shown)


martin-henz:
Yes as a proof-of-concept this is interesting. I don't think we should start hard-wiring module types into js-slang as in https://github.com/source-academy/js-slang/pull/1467/files#diff-2e2aa2db6619eec1be92cd4c227f51bafd0c3e8a0ba28a58c105e5e08944ec5d

Instead, the module should provide the types to be used in Source §x Typed.

We need to take a closer look at this next year.

martin-henz:
Closing this issue in favor of https://github.com/source-academy/modules/issues/217

PR #1469, opened 2023-08-21
Fix stepper: no precheck for undefined variables

Ziwen510:
Add a checker for undefined variables in the stepper before the interpreter is run.

### DIFF
--- a/src/stepper/converter.ts
+++ b/src/stepper/converter.ts
@@ -91,7 +91,7 @@ function evaluateFunctionObject(node: substituterNodes, context: Context) {
 }
 
 export function objectToString(value: any): string {
-  if ('toReplString' in value) {
+  if (!(value === null) && 'toReplString' in value) {
     return value.toReplString()
   }
   return '[Object]'
--- a/src/stepper/converter.ts
+++ b/src/stepper/converter.ts
@@ -91,7 +91,7 @@ function evaluateFunctionObject(node: substituterNodes, context: Context) {
 }
 
 export function objectToString(value: any): string {
-  if (!(value === null) && 'toReplString' in value) {
+  if (value !== null && 'toReplString' in value) {
     return value.toReplString()
   }
   return '[Object]'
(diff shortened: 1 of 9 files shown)


src/stepper/converter.ts
```diff
@@ -91,7 +91,7 @@ function evaluateFunctionObject(node: substituterNodes, context: Context) {
 }
 
 export function objectToString(value: any): string {
-  if ('toReplString' in value) {
+  if (!(value === null) && 'toReplString' in value) {
```
RichDom2185, line 94: Can we use `value !== null` instead for simplicity?

src/stepper/stepper.ts
```diff
@@ -2501,7 +2508,7 @@ function jsTreeifyMain(
     },
 
     Literal: (target: es.Literal): es.Literal => {
-      if (typeof target.value === 'object') {
+      if (typeof target.value === 'object' && !(target.value === null)) {
```
RichDom2185, line 2511: Likewise, can we use `target.value !== null` here?

RichDom2185 (commented):
Thanks for working on this so quickly! Is it possible to add more test cases in order to keep the test coverage up? What was the reason that some of the original test cases were deleted?

martin-henz (approved):
Looks good!

Ziwen510:
The three deleted test cases will actually cause undefined variable error in our normal transpiler run, this pr makes them raise errors in stepper (which is originaly runnable). I have added one new test case.

src/stepper/converter.ts
Ziwen510: Done

src/stepper/stepper.ts
Ziwen510: Done

RichDom2185:
> The three deleted test cases will actually cause undefined variable error in our normal transpiler run, this pr makes them raise errors in stepper (which is originaly runnable). I have added one new test case.

Hmm, perhaps we can leave this as an issue then, cause it makes sense to test each individual behaviour separately.

RichDom2185 (approved):
LGTM, thanks!

RichDom2185:
> > The three deleted test cases will actually cause undefined variable error in our normal transpiler run, this pr makes them raise errors in stepper (which is originaly runnable). I have added one new test case.
> 
> Hmm, perhaps we can leave this as an issue then, cause it makes sense to test each individual behaviour separately.

Referenced in #1475.

PR #1470, opened 2023-08-21
Use POSIX paths on Windows systems

ianyong:
## Context

> The default operation of the `node:path` module varies based on the operating system on which a Node.js application is running. Specifically, when running on a Windows operating system, the `node:path` module will assume that Windows-style paths are being used.
>
> -- https://nodejs.org/api/path.html#windows-vs-posix

While the above behaviour is useful when it comes to manipulating files on the host operating system (OS)'s filesystem, we want the preprocessor to make use of POSIX file paths because BrowserFS in the frontend is a POSIX filesystem.

Note that it is technically not wrong to make use of Windows file paths in the preprocessor, but this would introduce additional complexity into the tests (because they would need to handle file paths in both systems) as well as other parts of the preprocessor (because the Windows drive letter contains characters which are not valid JavaScript names). In fact, if js-slang were to support running multiple-file programs on the host OS in the future, this behaviour might even be necessary. However until that is the case, we should apply the [YAGNI principle](https://martinfowler.com/bliki/Yagni.html) to keep things simple.

## Changes

* Make use of the POSIX specific implementations of the `path` methods regardless of OS by importing from [`node:path/posix`](https://nodejs.org/api/path.html#pathposix)

## How to Test

There should be no change in behaviour on POSIX OSes. I will need someone using Windows to verify that the preprocessor tests all pass.

### DIFF
--- a/src/localImports/preprocessor.ts
+++ b/src/localImports/preprocessor.ts
@@ -1,5 +1,5 @@
 import es from 'estree'
-import * as path from 'path'
+import * as posixPath from 'path/posix'
 
 import { CannotFindModuleError, CircularImportError } from '../errors/localImportErrors'
 import { parse } from '../parser/parser'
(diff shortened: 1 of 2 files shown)


RichDom2185 (commented):
I was just setting up `js-slang` on Windows and it seems like both the master branch and this branch are failing `yarn test`

@leeyi45 do you know if this is expected? Generally, I use WSL when developing on my Windows machine, instead of the Windows file system directly.

ianyong:
> I was just setting up `js-slang` on Windows and it seems like both the master branch and this branch are failing `yarn test`
> 
> @leeyi45 do you know if this is expected? Generally, I use WSL when developing on my Windows machine, instead of the Windows file system directly.

The acceptance criterion for this PR is that the preprocessor tests no longer fail. Any other tests failing is out of scope.

Also, if we intend to run js-slang on Windows, it might be good to add an OS [matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy) so that regressions that are specific to Windows can be caught early.

leeyi45 (approved):
LGTM. Let's keep the matrix strategy idea in mind, but since I'm the only one trying to develop on Windows at the moment its not too high a priority

PR #1471, opened 2023-08-22
Use Asynchronous Code for Loading Modules

leeyi45:
This PR is mainly aimed at addressing #1348 . I have only been able to write an async implementation for the transpiler and stepper. Both the interpreter and ec-evaluator will need further work and discussion for such implementations.

This PR also incorporates the changes from #1470 .

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[*]
+end_of_line = lf
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,6 +9,7 @@
     "plugin:@typescript-eslint/recommended-requiring-type-checking",
     "prettier"
   ],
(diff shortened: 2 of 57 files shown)


martin-henz:
> This PR is mainly aimed at addressing #1348 . I have only been able to write an async implementation for the transpiler and stepper. Both the interpreter and ec-evaluator will need further work and discussion for such implementations.
> 
> This PR also incorporates the changes from #1470 .

I think the interpreter is no longer in use, see https://github.com/source-academy/js-slang/issues/1348

So what would remain to be refactored is the module loading of ec-evaluator.

Yongbeom-Kim (approved):
LGTM... but I'm not sure if I'm familiar enough with the js-slang repo that I didn't miss anything

leeyi45:
I'll do another once over myself, but if all's good I'll do the merge

PR #1472, opened 2023-08-23
Fix evaluation error when no evaluation code is provided

sayomaki:
## Description

This PR aims to resolve a bug where if there is no code to be evaluated in the playground stepper, an error will be thrown in the page (see attached image below:)

Screenshot 2023-08-23 at 11-49-10 Source Academy

The bug can be replicated by simply executing empty code in the playground stepper, or when modules are imported but no code has been provided, e.g.
```
import { heart } from "rune";
```

The first commit attempts to resolve the problem, but no useful information is provided back to the user (looks as if nothing has occurred) (maybe this could be detected and implemented via the frontend?), and thus the second commit will return an evaluation response with a message to signal that nothing has been evaluated (see images below):

<details>
<summary>New behaviour when nothing is evaluated (click to expand)</summary>

Screenshot 2023-08-23 at 11-47-49 Source Academy

Screenshot 2023-08-23 at 11-47-37 Source Academy
</details>

If the second commit/user interaction feedback is not required (e.g. implemented in frontend), feel free to let me know and I will remove the commit from this PR.

## Testing
These changes have been tested locally, and no longer produces the error as shown above in the first image.

### DIFF
--- a/src/stepper/__tests__/__snapshots__/stepper.ts.snap
+++ b/src/stepper/__tests__/__snapshots__/stepper.ts.snap
@@ -2095,6 +2095,8 @@ false ? 1 : 100 * factorial(100 - 1);
 "
 `;
 
+exports[`Evaluation of empty code and imports Evaluate empty program 1`] = `""`;
+
 exports[`Infinite recursion 1`] = `
 "function f() {
   return f();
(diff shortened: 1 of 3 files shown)


RichDom2185 (approved):
LGTM, thanks a lot! Prof @martin-henz, shall we bump js-slang and quickly deploy the new changes after this?

RichDom2185:
One minor comment, would it be possible to add test cases to ensure this behaviour?

Issue #1476, opened 2023-09-03
Remove src/interpreter/interpreter.ts


martin-henz:
I think src/ec-evaluator is replacing src/interpreter/interpreter.ts, so the latter is dead code.

Background: The interpreter used to be the workhorse implementation in SA, prior to the transpiler. After the transpiler was introduced, it became the implementation behind the environment visualizer. Now the environment visualizer uses ec-evaluator, which means that the interpreter is unemployed.

I suggest we should remove src/interpreter/interpreter.ts from js-slang, and leave some breadcrumbs in the wiki. It is bound to decay over time and cause extra work.

See also https://github.com/source-academy/js-slang/issues/1695

RichDom2185:
I think we should proceed with the removal. The only references to `src/interpreter/interpreter.ts` are as follows:

* https://github.com/source-academy/js-slang/blob/9b489441ac30fab86b5e72cdbd63bb46fc47dfe6/src/interpreter/closure.ts#L15-L20

* https://github.com/source-academy/js-slang/blob/9b489441ac30fab86b5e72cdbd63bb46fc47dfe6/src/runner/sourceRunner.ts#L115-L127

* https://github.com/source-academy/js-slang/blob/9b489441ac30fab86b5e72cdbd63bb46fc47dfe6/src/mocks/context.ts#L5

  Perhaps we can just replace this with
  ```tsx
  import { createBlockEnvironment } from '../ec-evaluator/utils'
  ```

How are would it be to completely remove it?

martin-henz:
I think we should completely remove it. It's a confusing legacy component at this point. Where are these "mocks" used?

RichDom2185:
> I think we should completely remove it. It's a confusing legacy component at this point. Where are these "mocks" used?

The mocks in the last bullet point are used in test files. I'll make a PR with the necessary changes.

leeyi45:
Resolved by #1745

PR #1484, opened 2023-09-10
Fix incorrect `function.name` for aliased imports

RichDom2185:
Uses `Object.defineProperty` to set the correct `function.name` value for aliased imports in:

* ec-evaluator
* transpiler

Did not do so for `interpreter/interpreter.ts` due to the deprecation note in #1476.

### How to test

Consider the following program:

```js
import { play as asdf } from 'sound';

asdf("test");
```

Previously, this will result in the following error:

```
Error: play is expecting sound, but encountered test
```

Now, it will result in:

```
Error: asdf is expecting sound, but encountered test
```

---

Tested with:

* Source §1/2/3/4
* Source §1/2/3/4 Native
* Source §1/2/3/4 Typed
* Source §1/2 Lazy
* Source §4 Explicit-Control
* Full JS

### DIFF
--- a/src/ec-evaluator/interpreter.ts
+++ b/src/ec-evaluator/interpreter.ts
@@ -195,7 +195,9 @@ function evaluateImports(
           }
 
           declareIdentifier(context, spec.local.name, node, environment)
-          defineVariable(context, spec.local.name, functions[spec.imported.name], true, node)
+          const importedObj = functions[spec.imported.name]
+          Object.defineProperty(importedObj, 'name', { value: spec.local.name })
+          defineVariable(context, spec.local.name, importedObj, true, node)
         }
       }
     })
--- a/src/runner/sourceRunner.ts
+++ b/src/runner/sourceRunner.ts
@@ -284,7 +284,7 @@ export async function sourceRunner(
     return runNative(program, context, theOptions)
   }
 
-  return runInterpreter(program!, context, theOptions)
+  return runInterpreter(program, context, theOptions)
 }
 
 export async function sourceFilesRunner(
--- a/src/transpiler/transpiler.ts
+++ b/src/transpiler/transpiler.ts
@@ -109,21 +109,41 @@ export async function transformImportDeclarations(
       }
 
       const declNodes = nodes.flatMap(({ specifiers }) =>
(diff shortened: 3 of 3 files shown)


src/infiniteLoops/runtime.ts
```diff
@@ -268,6 +268,7 @@ function prepareBuiltins(oldBuiltins: Map<string, any>) {
     }
   }
   newBuiltins.set('undefined', undefined)
+  newBuiltins.set('Object', Object)
```
RichDom2185, line 271: Not sure if this is the right move; what are the implications of this? Perhaps someone could verify/advise.

src/infiniteLoops/instrument.ts
```diff
@@ -587,7 +587,9 @@ async function handleImports(programs: es.Program[]): Promise<[string, string[]]
         }
       )
       program.body = (importsToAdd as es.Program['body']).concat(otherNodes)
-      const importedNames = importsToAdd.flatMap(node =>
+      const importedNames = (
+        importsToAdd.filter(node => node.type === 'VariableDeclaration') as es.VariableDeclaration[]
+      ).flatMap(node =>
```
RichDom2185, line 592: I'm not sure if there's a less hacky way to handle this...

RichDom2185 (commented):
I just have some doubts on the code quality as follows:

src/infiniteLoops/runtime.ts
martin-henz: I see that you are updating the name property in the function object. But here, are you trying to add the name "Object" to the names of builtins that are visible from Source programs?

leeyi45:
If multiple files alias the same import the name information is lost

```ts
// a.js
import { show as s } from 'rune';
export function a(x) {
  return s(x);
}

// b.js
import { show as b } from 'rune';
import { a } from './a.js';

b(0);
a(0);

```

would theoretically cause the second call to `a` to display the function's name as `'b'` (or vice versa). However, right now we only use the `Function.name` syntax when throwing errors, so this isn't an issue at the present moment

leeyi45:
The function's name also doesn't change in the documentation pop-ups

PR #1489, opened 2023-09-17
minor improvements in env viz

martin-henz:
* avoid pushing marker if agenda is empty
* remove PUSH_UNDEFINED_IF_NEEDED
* distinguish value-producing loop bodies; insert UNDEFINED only for non-value-producing loop bodies
* avoid pushing environment if agenda is empty


vansh284 (approved):
LGTM! Don't see any problems and the changes will make the tool more intuitive for students to use.

PR #1491, opened 2023-09-29
minor change in while loops to be consistent with for loops

martin-henz:


RichDom2185 (commented):
Can we bump the version in `package.json` for this repo as well? (To 1.0.36)

Issue #1493, opened 2023-10-17
CSE Machine: "Redeclaring name" error


martin-henz:
This program runs correctly in the default runtime, but gives an error when run in CSE machine:
```
function f() {
    function g(t) {
        return 0;
    }
    return g;
}
const h = f();
h(100);
```
(https://share.sourceacademy.nus.edu.sg/i8q0o)

The error: ```Line 7: Redeclaring name h.```

martin-henz:
If function f has at least one parameter, the CSE machine will work correctly.

PR #1494, opened 2023-10-17
Fix nullary functions on the CSE machine

dickongwd:
This pull request closes https://github.com/source-academy/js-slang/issues/1493.

Made a simple change to remove the check for function arguments when pushing environment instructions during application calls. The machine now always restores the original environment after function applications.

Also added 2 tests to the test suite to test if nullary functions properly restore their environments. These 2 code snippets fail on the current live version of Source Academy:

```
function f() {
    function g(t) {
        return 0;
    }
    return g;
}
const h = f();
h(100);
```

```
function f() {
  const a = 1;
  return a;
}
const a = f();
a;
```

### DIFF
--- a/src/ec-evaluator/__tests__/__snapshots__/ec-evaluator.ts.snap
+++ b/src/ec-evaluator/__tests__/__snapshots__/ec-evaluator.ts.snap
@@ -45,6 +45,44 @@ foo();",
 }
 `;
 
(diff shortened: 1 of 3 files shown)


RichDom2185 (approved):
LGTM, can we update the test snapshots?

PR #1507, opened 2023-11-09
Import Analyzer and Augmentations

leeyi45:
All import declarations are checked for undefined imports, including local modules, so importing a symbol that isn't exported by a module will throw an error. Furthermore, the preprocessor and transpiler have been augmented to be able handle all 3 types of import specifiers. Though namespace imports may not be allowed in Source, the functionality has been incorporated so that the full JS and TS runners may make use of them.

Currently `js-slang` assumes that multiple files means that the source file path should be included with the AST. I've provided an option under `IOptions` to add the file path information to the AST. This is so that even without knowing beforehand the number of files required, file path information can be added.

As per @ianyong's suggestions, I've kept the number of type assertions and scale of these changes to a minimum.

I'm not 100% sold on some of the names I've given things (e.g. `linker.ts`), so any suggestions regarding those would be appreciated.


leeyi45:
I've noticed some serious bugs not being caught by tests, let me work on those before reopening this

leeyi45:
This really should be merged soon....

src/modules/preprocessor/filePaths.ts
```diff
@@ -1,8 +1,6 @@
-import {
-  ConsecutiveSlashesInFilePathError,
-  IllegalCharInFilePathError,
-  InvalidFilePathError
-} from '../errors/localImportErrors'
+import { ConsecutiveSlashesInFilePathError } from '../errors'
+import { IllegalCharInFilePathError } from '../errors'
+import { InvalidFilePathError } from '../errors'
```
RichDom2185, line 3: Are we not combining all three imports into one?

src/parser/__tests__/disallowed-syntax.ts
```diff
@@ -1206,25 +1208,25 @@ test('no assigning to reserved keywords - verbose', () => {
...
+//     "enable verbose";
+//     import * as x from "one_module";
+//     `,
+//     { chapter: Chapter.SOURCE_4 }
+//   ).toMatchInlineSnapshot(`
+//             "Line 2, Column 7: Namespace imports are not allowed
+//             You are trying to use Namespace imports, which is not allowed (yet).
+//             "
+//           `)
+// })
```
RichDom2185, line 1232: Why is this commented out?

tsconfig.prod.json
```diff
@@ -0,0 +1,18 @@
...
+  "exclude": [
+    "src/stdlib/metacircular-interpreter",
+    "src/stdlib/**/*.js",
+    "node_modules",
+    "dist",
+    "sicp_publish",
+    "src/**/__tests__/**",
+    "src/**/__mocks__/**",
+  ]
+}
```
RichDom2185, line 18: Can we add a newline at the end of this file?

RichDom2185 (commented):
Did a quick, non-thorough skim over the files and generally looks good, thanks!

Can I check what's the motivation for allowing namespace imports in Source?

leeyi45:
There's not really a main motivation: I just thought that since namespace imports are supported in fullJS I might as well implement it along the way

package.json
```diff
@@ -78,7 +79,7 @@
     "escodegen": "^2.0.0",
     "eslint": "^8.57.0",
     "eslint-config-prettier": "^9.1.0",
-    "eslint-plugin-simple-import-sort": "^12.0.0",
+    "eslint-plugin-import": "^2.29.1",
```
RichDom2185, line 82: What was the rationale for changing to this over the previous plugin?

src/name-extractor/__tests__/autocomplete.ts
RichDom2185, line 1: Should we use the `stripIndent` utility for the test cases here?

src/name-extractor/__tests__/autocomplete.ts
```diff
@@ -1,547 +1,524 @@
...
+      { name: 'baz', meta: DeclarationKind.KIND_LET, score: 1 },
+      { name: 'bar', meta: DeclarationKind.KIND_LET, score: 0 }
+    ]
+  ],
+  [
+    'Test accessing for-loop parameter in for-loop body',
     '\
+      for (let foo=10;) {\n\
+        f\n\
+      }\
```
RichDom2185, line 195: Can we replace this (and all the other occurrences) with a template literal instead?

RichDom2185 (commented):
I'm almost done with the review (just left with a couple more files after this one) – but just some minor comments for now:

src/modules/preprocessor/analyzer.ts
```diff
@@ -0,0 +1,190 @@
...
+              continue
+            }
+
+            const importedName = getImportedName(spec)
+
+            if (!dstModuleDocs.has(importedName)) {
+              if (importedName === 'default') throw new UndefinedDefaultImportError(dstModule, spec)
+              throw new UndefinedImportError(importedName, dstModule, spec)
+            }
+          }
```
joeng03, line 147: The outer `if` could be refactored to a guard clause, thus reducing a layer of code.

joeng03 (commented):
Great work overall, just a few nits. 

Could you guide me through the execution flow which enables this test case:
```
        '/a.js': `import * as a from 'one_module';`,
        '/b.js': `import * as a from 'one_module';`
```
to throw `DuplicateImportNameError`? It seems to me that although namespaceSpecifiers.length > 0, regularSpecifiers.length is 0, therefore this predicate
```
        inamespaceSpecifiers.length > 0 && regularSpecifiers.length > 0) ||
        importedNames.size > 1
```
will be false? I may be missing something here

package.json
leeyi45: The previous plugin doesn't handle import duplicates, the new ones does

src/modules/preprocessor/analyzer.ts
```diff
@@ -0,0 +1,190 @@
...
+      const importedNames = new Set(regularSpecifiers.map(getImportedName))
+
+      if (
+        (namespaceSpecifiers.length > 0 && regularSpecifiers.length > 0) ||
+        importedNames.size > 1
+      ) {
+        // This means that there is more than one unique export being given the same
+        // local name
+        const specs = [...regularSpecifiers, ...namespaceSpecifiers]
+        throw new DuplicateImportNameError(localName, specs)
```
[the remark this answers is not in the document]
leeyi45: What do you mean?

leeyi45:
> Great work overall, just a few nits.
> 
> Could you guide me through the execution flow which enables this test case:
> 
> ```
>         '/a.js': `import * as a from 'one_module';`,
>         '/b.js': `import * as a from 'one_module';`
> ```
> 
> to throw `DuplicateImportNameError`? It seems to me that although namespaceSpecifiers.length > 0, regularSpecifiers.length is 0, therefore this predicate
> 
> ```
>         inamespaceSpecifiers.length > 0 && regularSpecifiers.length > 0) ||
>         importedNames.size > 1
> ```
> 
> will be false? I think I'm missing something here

This case shouldn't trigger a DuplicateImportError

src/modules/preprocessor/analyzer.ts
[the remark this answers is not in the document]
RichDom2185: @joeng03 do you mean removing the spread operator? Even if we use `Array.concat` it still returns a new copy basically equivalent to the more readable spread syntax.

RichDom2185 (approved):
LGTM, thanks a lot! Let's merge this before more conflicts arise; the above comments are minor nits and shouldn't be a blocker.

PR #1517, opened 2024-01-27
CSE Machine: implement variable declaration check

NhatMinh0208:
# Description:
In the stepper and transpiler, the function `checkForUndefinedVariables` ensures that all variables used in the program are declared before the program is run. The CSE machine does not include this check, meaning that some programs containing undeclared variables run normally if such variables are never accessed. This PR implements a version of  `checkForUndefinedVariables` for the CSE machine, which the stepper can also utilize.

# Changes:

- [X] implement  `checkProgramForUndefinedVariables` for the stepper and CSE machine


martin-henz (approved):
Looks good to me!

PR #1518, opened 2024-01-31
CSE Machine: Fix array expression unpack order

NhatMinh0208:
# Description

This PR reverses the order in which values in an array expression are evaluated so that it is consistent with the Source language and JS specification. 

# Changes

- [X] Fix array expression unpack order


martin-henz (approved):
Great. Well done.

PR #1521, opened 2024-02-02
CSE Machine: `hasXStatement` fixes

NhatMinh0208:
# Description:

The functions `hasContinueStatement`, `hasReturnStatement` and `hasBreakStatement` are used to detect `continue`, `return` and `break` statements. These functions expect a block statement and recurse into `if` statements, meaning they fail on `else if` constructs since these constructs nest an `if` statement inside an `if` statement.

This PR adds variants of the above functions ( `hasContinueStatementIf`, `hasReturnStatementIf` and `hasBreakStatementIf`) so that `else if` constructs can now be properly recursed into.

# Changes:

- [X] Add `hasContinueStatementIf`, `hasReturnStatementIf` and `hasBreakStatementIf`
- [X] Fix `hasContinueStatement`, `hasReturnStatement` and `hasBreakStatement` to work with `else if` constructs

### DIFF
--- a/src/cse-machine/utils.ts
+++ b/src/cse-machine/utils.ts
@@ -508,9 +508,9 @@ export const hasReturnStatementIf = (statement: es.IfStatement): boolean => {
   // Parser enforces that if/else have braces (block statement)
   hasReturn = hasReturn || hasReturnStatement(statement.consequent as es.BlockStatement)
   if (statement.alternate) {
-    if (statement.alternate.type === 'IfStatement') {
+    if (isIfStatement(statement.alternate)) {
       hasReturn = hasReturn || hasReturnStatementIf(statement.alternate as es.IfStatement)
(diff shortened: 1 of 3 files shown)


src/cse-machine/utils.ts
```diff
@@ -503,56 +503,107 @@ export const checkStackOverFlow = (context: Context, control: Control) => {
   }
 }
 
-/**
- * Checks whether a function body returns in every possible branch.
- * Returns true if every branch has a return statement, else returns false.
- * @param body The function body to be checked
- */
```
RichDom2185, line 510: Why was this documentation removed?

NhatMinh0208: Sorry for the error, I've mistakenly implemented a wrong version of `hasReturnStatement`. I've just fixed the function and put the documentation back.

martin-henz (approved):
Looks good now.

PR #1522, opened 2024-02-02
1395 remove hints type inference

ZweZeya:
Removed getTypeInformation from src/index.js and typeCheck from src/typeChecker/typeChecker.js which were used in the frontend repository to render type hints.


RichDom2185 (commented):
Seems like you forgot to remove the tests, resulting in the CI failure?

martin-henz (approved):
Looks good!

PR #1538, opened 2024-02-15
CSE Machine: Implement two-step block statement evaluation

### LINKED ISSUE #1537: CSE Machine: Evaluation of block statements should be split into two steps
Currently, when evaluating block statements, the CSE machine creates the environment for the block _and_ unpacks the block statement in the same step, as seen when running [this](https://share.sourceacademy.org/dicgu) program.

This is confusing to students. We want to split evaluation of block statements into two steps:

1. Create the environment and the necessary instructions.
2. Unpacks the block statement and place its content into the Control stack.

NhatMinh0208:
# Description

This PR implements two-step block statement evaluation for the CSE machine.

This is done by introducing a new subtype of `BlockStatement` called `RawBlockStatement` to indicate a block statement that has completed environment creation, and only needs to get its body unpacked into the control stack.

Resolves #1537.

# Changes
- [X] add `RawBlockStatement`
- [X] change block handling logic to push `RawBlockStatement` instead of unpacking directly


RichDom2185 (approved):
LGTM, thanks!

PR #1541, opened 2024-02-16
1540 stepper last line of reductible code is not evaluated

### LINKED ISSUE #1540: Stepper: Last line of reductible code is not evaluated
The last line of a program is not evaluated in the stepper. 

Bug can be reproduced by doing a variable/function declaration.

hanscau:
Fix the issue of the stepper not evaluating the last line of statement. 

Changes:
1. Push "start of evaluation" into steps prior to beginning of reducible loop
2. Changed reducible condition to be based on isReducible rather than the length of program remaining
3. Added test to verify that one-liners are now evaluated

Resolves #1540


.github/workflows/nodejs.yml
RichDom2185, line 1: Please revert this change, the errors are universal and not caused by an incorrect workflow file (at least, does not seem to be). If it's temporary and for the purpose to see which other tests are failing then it's fine, but let's undo this change before the final version of this PR.

src/stepper/__tests__/stepper.ts
```diff
@@ -1424,7 +1455,7 @@ describe(`Evaluation of empty code and imports`, () => {
   test('Evaluate empty program', async () => {
     const code = ``
     const steps = await testEvalSteps(code, mockContext())
-    expect(steps.map(x => codify(x[0])).join('\n')).toMatchSnapshot()
+    // expect(steps.map(x => codify(x[0])).join('\n')).toMatchSnapshot()
```
RichDom2185, line 1458: Why was this commented out?

src/stepper/stepper.ts
```diff
@@ -60,6 +60,19 @@ function isIrreducible(node: substituterNodes, context: Context) {
...
+  if (progs.body.length === 0) return false
+  if (progs.body.length > 1) return true
+
+  const [lastStatement] = progs.body
+
+  if (lastStatement.type === 'ExpressionStatement') {
+    return !isIrreducible(lastStatement.expression, context)
+  } else {
+    return true
+  }
```
RichDom2185, line 73: Let's use early return to avoid "arrowhead"-style code:

```suggestion
  if (lastStatement.type !== 'ExpressionStatement') {
    return true
  }
  return !isIrreducible(lastStatement.expression, context)
```

RichDom2185: By the way, why does it return a negation of the reducibility of the last expression?

RichDom2185 (commented):
Hi, thanks for working on this! I just have some thoughts below.

Looking forward to your replies!

src/stepper/stepper.ts
hanscau: it should return if the expression is reducible, but since the function that exists is in the negation form, it is double negated to achieve isReducibleExpression

src/stepper/__tests__/stepper.ts
hanscau: Now the stepper evaluates even an empty program (previously it will return "Nothing is evaluated") hence the snapshot is different. Will update it

src/stepper/stepper.ts
RichDom2185: Oh, right, my bad, misread the function name, thanks!

martin-henz (approved):
The PR is good to go. The notion of reducibility will change in the future, but that's independent. Let's merge this first.

PR #1542, opened 2024-02-17
CSE Machine: Fix environment reset instruction insertion

### LINKED ISSUE #1506: CSE machine: function application in const as last statement
This program fails to run in the CSE machine:
```
function goo() {
    const g = () => 1;
    return g;
}
const p = goo()();
//999; //uncomment for this to work in CSE
```
See https://share.sourceacademy.nus.edu.sg/90gak

### LINKED ISSUE #1520: CSE Machine: Redeclaring name bug
While working on #1503 , I encountered a very weird bug
https://share.sourceacademy.nus.edu.sg/m9578
The result of the program should be undefined but instead it throws an error: "Redeclaring name b"

Image

NhatMinh0208:
# Description

This PR fixes the condition for inserting an environment reset instruction so that they are still inserted when the control stack has no nodes.

Resolves #1506, resolves #1520.

# Changes
- [X] fix condition for inserting environment reset instruction

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -28,7 +28,7 @@ jobs:
         exit 0
     - name: Install dependencies (apt)
       run: |
-        sudo apt-get update;
+        sudo apt-get update
         sudo apt-get install -y --no-install-recommends libxi-dev libgl1-mesa-dev
     - name: Setup Node
       uses: actions/setup-node@v4
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -28,6 +28,7 @@ jobs:
         exit 0
     - name: Install dependencies (apt)
       run: |
+        sudo apt-get update;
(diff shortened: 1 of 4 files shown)


.github/workflows/nodejs.yml
```diff
@@ -28,6 +28,7 @@ jobs:
         exit 0
     - name: Install dependencies (apt)
       run: |
+        sudo apt-get update;
```
RichDom2185, line 31: This semicolon is not needed as `|` already means preserve newlines (see <https://yaml-multiline.info/>).

```suggestion
        sudo apt-get update
```

Thanks for helping to fix the CI!

RichDom2185 (commented):
Changes look good, is this PR ready? If yes, let's mark it as ready and merge it quickly so that the workflows are fixed.

NhatMinh0208:
> Changes look good, is this PR ready? If yes, let's mark it as ready and merge it quickly so that the workflows are fixed.

Yeah it's ready now, lets get it merged

RichDom2185 (approved):
LGTM, thanks!

PR #1543, opened 2024-02-18
Remove interpreter

### LINKED ISSUE #1476: Remove src/interpreter/interpreter.ts

RichDom2185:
Closes #1476 

Todo:
- [ ] Fix tests


CZX123:
Just a heads up that #1645 will likely affect this PR

RichDom2185:
> Just a heads up that #1645 will likely affect this PR

No worries, it's already quite conflicted anyway, though thanks for the heads up!

RichDom2185:
Superseded by #1745

PR #1544, opened 2024-02-18
Fix CI workflows

RichDom2185:
Ensures that CI dependencies can be installed, fixing workflow errors once sources get outdated.


sayomaki (approved):
LGTM, good to be merged!

PR #1546, opened 2024-02-18
Bump minor versions of dependencies

RichDom2185:
Bumps minor versions of dependencies.

Also replaces lodash's individual types with types for the whole package (since we have the whole package as a dependency).


sayomaki (approved):
LGTM, thanks for the changes!

PR #1547, opened 2024-02-19
Implement continuations through the extension of the CSE-Machine

s-kybound:
This PR implements first-class continuations for the CSE machine by adding two new instructions, GENERATE_CONT and RESUME_CONT, which generate a continuation and consume continuations respectively.

A continuation is a structure capturing a saved state of the program, storing copies of the control, stash as well as a pointer to the environment at the point of time the continuation is captured. When it is called, the program control, stash and environment pointer are forcibly set to that of the continuation, effectively reverting the program back to the point the continuation is generated. Reifying continuations allow us to pass these structures as parameters, assign them and call them at any point of time, perhaps several times, allowing for highly complex program flows.

This is required to implement Scheme through the scm-slang project, but in theory simply by adding the built-in `call_with_current_continuation`/`call/cc` to a Source Language's builtins, and by forcing that language to use the explicit control evaluator exclusively, continuations should become first-class for that language.

Changes:
- The most significant changes have been made to the cse-machine folder
1. Continuations have been implemented as functions with additional data in the `continuations.ts` file
2. Interpreter: 2 Instructions, GENERATE_CONT, RESUME_CONT, have been created
3. Interpreter: cmdEvaluator for APPLICATION has been modified to check for call/cc function and continuation. Additionally, the function has been written with guard clauses.
- Scheme now exclusively use the CSE for evaluation
- `call/cc` added to Scheme's builtins


martin-henz (approved):
Nice and clean implementation.

PR #1556, opened 2024-02-20
Typed Source: Disallow the use of predeclared constants as literal types

### LINKED ISSUE #1553: Typed Source: Predeclared constants are also types
image

In lines 1 and 2, the predeclared constants `NaN` and `math_PI` can be used as types without error. But those constants are values and it should not be possible to use them as types.

Zenkoh1:
This removes the predeclared constants literal types.

Closes #1553


RichDom2185 (approved):
LGTM, thanks!

PR #1563, opened 2024-03-01
1562 stepper incorrect evaluation sequence in program

hanscau:
Added checks for reducible second statements in accordance to the stepper specs.

Reduce second statements if it is reducible, keeping the first statement on the buffer. 

Fix code highlighting due to shifting of statements

Added tests and updated old tests to reflect new changes.

### DIFF
--- a/src/stepper/stepper.ts
+++ b/src/stepper/stepper.ts
@@ -1796,8 +1796,6 @@ function reduceMain(
           isIrreducible(firstStatement.expression, context)
         ) {
           const [secondStatement] = otherStatements
-          // if the first statement is irreducible, reduce second statement if it's reducible
-          // else reduce the first statement
           if (
             secondStatement !== undefined &&
             secondStatement.type == 'ExpressionStatement' &&
--- a/src/stepper/stepper.ts
+++ b/src/stepper/stepper.ts
@@ -2010,7 +2010,7 @@ function reduceMain(
           ) {
             paths[0].push('body[0]')
             paths.push([])
-            let stmt = ast.blockStatement(otherStatements as es.Statement[])
+            const stmt = ast.blockStatement(otherStatements as es.Statement[])
             return [stmt, context, paths, explain(node)]
           } else {
             // Reduce the second statement and preserve the first statement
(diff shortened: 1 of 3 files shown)


src/stepper/stepper.ts
```diff
@@ -1795,15 +1795,42 @@ function reduceMain(
...
-          // let stmt
-          // if (otherStatements.length > 0) {
-          paths[0].push('body[0]')
-          paths.push([])
-          const stmt = ast.program(otherStatements as es.Statement[])
-          // } else {
-          //   stmt = ast.expressionStatement(firstStatement.expression)
-          // }
-          return [stmt, context, paths, explain(node)]
+          const [secondStatement] = otherStatements
```
martin-henz, line 1798: should this be
```
const [secondStatement, ...remainingStatements] = otherStatements
```
The pattern matching makes it look like the array has only one element at this point, but it could have more, right?

Aparently, in JS, this program succeeds:
```
const [a] = [1, 2]
display(a) // displays 1
```

src/stepper/stepper.ts
```diff
@@ -1972,15 +1999,45 @@ function reduceMain(
           firstStatement.type === 'ExpressionStatement' &&
           isIrreducible(firstStatement.expression, context)
         ) {
-          let stmt
-          if (otherStatements.length > 0) {
+          const [secondStatement] = otherStatements
```
martin-henz, line 2002: Is this similar to the issue above?

src/stepper/stepper.ts
```diff
@@ -2148,15 +2205,45 @@ function reduceMain(
...
```
martin-henz, line 2208: ditto

martin-henz (changes requested):

src/stepper/stepper.ts
hanscau: remainingStatement won't be utilised, so I did not include it in. Are we okay with that?

RichDom2185: Please refer to my comment offline. In short, I prefer the original implementation without the extra assignment, but for clarity might be good to add a comment explaining that the remaining statements are not used.

martin-henz (approved):
All good now.

PR #1564, opened 2024-03-02
Refactor Runners Part 1

leeyi45:
As discussed, I've been trying to standardize how the different Source runners are called. This PR is the first set of changes. `js-slang` will be able to read and run Source programs straight from the desktop.

If merged, this will supersede #1507


martin-henz:
Currently this PR doesn't pass linting, probably in src/vm/svml-machine.ts
See actions above. Maybe @leeyi45 can take another look?

leeyi45:
When I first wrote this I made several incorrect assumptions, will need to relook at this in the future

Issue #539, opened 2020-04-14
Stepper: Evaluation of toplevel statement sequences


martin-henz:
See https://github.com/source-academy/js-slang/issues/537

martin-henz:
See here for current status: https://github.com/source-academy/js-slang/issues/537#issuecomment-662235762

parnikkapore:
(reopening this to wait for the implementation - shouldn't have tagged the issues!)

martin-henz:
To clarify why this issue is still open: All evaluation paths handle 
```
3; {} // should produce 3
```
correctly, except the stepper. It is also fixed in the stepper specs, but not yet in the stepper implementation.

hanscau:
Solved by #1565

PR #1565, opened 2024-03-02
Stepper evaluation of toplevel statement sequences

hanscau:
Bug fix for issue #539 

Reduces empty block statements in the program, block statements and block expressions into epsilon. 

Added tests for all 3 contexts too


martin-henz (approved):
Looks good!

PR #1567, opened 2024-03-02
fix: reduce firstStatement if secondStatement is a return statement

hanscau:
Stepper was crashing because it was trying to reduce a return statement due to the first statement being a value. Fixed the evaluation sequence in accordance to the specs. 

image

Reduce value if the next statement is a return statement


martin-henz (approved):
Great!

PR #1569, opened 2024-03-03
CSE Machine: Push body directly for programs without top-level declarations

NhatMinh0208:
# Description:

This PR changes how `Program` nodes are handled.

Previously, the CSE machine would put `Programs` through the same two-step process as `BlockStatements`:
- create environment and repackage into `RawBlockStatement`
- unpack `RawBlockStatement` and push individual body nodes

This is done regardless of whether the `Program` has top-level declarations or not. This is now fixed so that `Programs` that don't have top-level declarations skip the first step and get its body pushed into the `Control` stack immediately.

# Changes:

- [X] Push body directly for programs without top-level declarations

### DIFF
--- a/src/cse-machine/interpreter.ts
+++ b/src/cse-machine/interpreter.ts
@@ -372,15 +372,21 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
       const next = command.body[0]
       cmdEvaluators[next.type](next, context, control, stash, isPrelude)
     } else {
-      // Push raw block statement
-      const rawCopy: RawBlockStatement = {
-        type: 'BlockStatement',
-        range: command.range,
-        loc: command.loc,
-        body: command.body,
-        isRawBlock: 'true'
+      // If program has top-level declarations, push raw block statement as intermediate step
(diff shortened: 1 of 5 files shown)


RichDom2185 (approved):
LGTM, thanks!

martin-henz:
This looks hacky to me. See also: https://github.com/source-academy/frontend/pull/2830#issuecomment-1975381888

We should just add a new kind of node "Sequence" to ASTs. Then any block that does not have a declaration will become a sequence, as far as the CSE machine is concerned. (We should do this in the stepper, too.)

We do this already in our parse function in Source §4, see js-slang/src/stdlib/parser.ts.

martin-henz (changes requested):
This looks hacky to me. See also: https://github.com/source-academy/frontend/pull/2830#issuecomment-1975381888

We should just add a new kind of node "Sequence" to ASTs. Then any block that does not have a declaration will become a sequence, as far as the CSE machine is concerned. (We should do this in the stepper, too.)

We do this already in our parse function in Source §4, see js-slang/src/stdlib/parser.ts.

PR #1570, opened 2024-03-04
Add cslang support

benmurphyy:
Add support for cslang by adding "FULL_C" programming language enum.Tied with the following source academy frontend PR: https://github.com/source-academy/frontend/pull/2831


martin-henz (approved):

RichDom2185 (commented):
Why are we changing `Chapter` from an enum to a const enum?

martin-henz:
> Why are we changing `Chapter` from an enum to a const enum?

Good catch. (Just looked up "const enum".) @benmurphyy the language chapters are not performance-critical. We should keep `Chapter` as a "normal" enum.

RichDom2185:
> > Why are we changing `Chapter` from an enum to a const enum?
> 
> Good catch. (Just looked up "const enum".) @benmurphyy the language chapters are not performance-critical. We should keep `Chapter` as a "normal" enum.

Yup, and there are some downsides to using const enum, to the point that the TS language guide suggests banning it in linters as a possibility: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

PR #1571, opened 2024-03-04
Typed Source: Fix error where template literals cannot be used

### LINKED ISSUE #1530: Typed Source: Template literals don't work
Template literals ("multiline strings") give out an error, but strangely only when trying to assign to a symbol (note that the error appears in line 5 – removing it will allow line 3 to run as expected).

image
image

Zenkoh1:
Add the case for nodes that are template literals

Closes #1530


martin-henz (approved):
Great. As discussed. The comment says it all.

PR #1572, opened 2024-03-05
Minimal change to expose steppable functionality from runCSEMachine()

JoelChanZhiYang:
Previously, the runCSEMachine function runs until the Control is empty or until the step limit is reached. 

It would be useful to extract each step in a generator function and export this functionality. This allows modules to run their own internal "steppable" CSE machine. 

The functionality of the runCSEMachine function is unchanged. 

```
// Example use case for generateCSEMachineStateStream

const code = context.unTypecheckedCode[0];
const context = createContext(...)

const program = parse(code, context);

context.runtime.isRunning = true;
context.runtime.control = new Control(program);
context.runtime.stash = new Stash();

const options = ... // Create options

const iterator = generateCSEMachineStateStream(
      context,
      context.runtime.control,
      context.runtime.stash,
      options.envSteps,
      options.stepLimit,
      options.isPrelude,
)

// Use a for loop to access the state at each step. For more control, use iterator.next()
for (const {stash, control, steps} of iterator) {
    // Do something with the state
}

```

### DIFF
--- a/src/cse-machine/interpreter.ts
+++ b/src/cse-machine/interpreter.ts
@@ -278,6 +278,31 @@ function runCSEMachine(
   stepLimit: number,
   isPrelude: boolean = false
 ) {
+  const eceState = generateCSEMachineStateStream(
+    context,
+    control,
+    stash,
+    envSteps,
+    stepLimit,
+    isPrelude
+  )
+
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Looks good!

PR #1577, opened 2024-03-07
Remove special cases for environment instruction generator

### LINKED ISSUE #1576: Inner Scope Access issues in REPL programs when using explicit control variants
The following code snippets added into the REPL line by line when using explicit control will behave unexpectedly.

```js
// using js-slang
(a => 1)(7);

a; // will access 7 even though a is in an inner scope
```

```js
// using js-slang
{ const b = 1; }

b; // will access 1 even though b is in an inner scope
```

```scheme
;; using scm-slang
(define a 1)

((lambda (a) a) '(1 2 3))

a ;; will access the value of list (1 2 3) and not 1
```

s-kybound:
The CSE machine currently does not inject an environment instruction into the control when a function application or a block statement is the last item in a program. While this works for a single program, it breaks the REPL, where each submitted code snippet is treated as a singular program. The best way to deal with this issue is to force an environment instruction for any environment-producing statements at the end of the program to allow for newer REPL snippets to read from the correct environment.

Resolves #1576.

### DIFF
--- a/src/cse-machine/interpreter.ts
+++ b/src/cse-machine/interpreter.ts
@@ -393,15 +393,8 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
     }
     // Normal block statement: do environment setup
     // To restore environment after block ends
-    // If there is an env instruction on top of the stack, or if there are no declarations, or there is no next control item
-    // we do not need to push another one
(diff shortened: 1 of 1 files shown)


martin-henz (approved):
Looks fine.

PR #1578, opened 2024-03-11
Revert const enum change in cslang

benmurphyy:
Revert const enum change from previous PR.


RichDom2185 (changes requested):
Sorry @benmurphyy, could you rebase this? I think Prof @martin-henz's merge undid your revert.

RichDom2185 (approved):
LGTM, thanks!

PR #1582, opened 2024-03-16
CSE Machine: Changes to step count

NhatMinh0208:
# Description
This PR makes some changes to how step count is updated, so that `context.runtime.envStepsTotal` represents the number of steps completed by the machine before the program finishes/runs into an error.

This is needed to implement source-academy/frontend#2847.
# Changes
- [X] `context.runtime.envStepsTotal` is updated during evaluation
- [X] Step count now starts from 0 and represents the number of steps completed


martin-henz (approved):
Looks good!

PR #1584, opened 2024-03-18
Update parser to new scm-slang parser

s-kybound:
This update completely converts the scm-slang parser used in js-slang to use the newer version.

This improved parser supports:

- Full coverage of location data - no more missing location data errors
-  Variadic functions
- Redefinitions of the same variable within the same scope are now possible
- Scope repair - begin no longer has its own scope - closer to Scheme behaviour
- Optimizations - unnecessary begin and sequence IIFE expressions can now be "absorbed" by enclosing functions and scopes, allowing for closer 1-1 behaviour between program and stepper/CSE machine
- Scheme pair representation is now 1-1 with Source Academy's representation
- Vectors are now supported
- Scheme numeric tower has been implemented
- Parser and Lexer errors have been improved to help in usefulness
- Datum comments with '#;' are implemented for Quality of Life
- Promise representation has been improved to follow Scheme - promises can be forced once, and subsequently store the value emitted

Most importantly - the parser now uses an intermediate AST representation, which has allowed us to perform above actions such as optimizations and scope repair. Most importantly, it is the first step for conversion between a full LISP AST and the IR AST used, which will greatly help in the future when implementing macro systems for the scheme implementation.

On js-slang's side, continuation representation now accepts a variadic amount of arguments, which will, in the future, allow us to implement scheme procedures that produce more than 1 value.


martin-henz (approved):
Looks good. Nice comments. Great to see the CSE machine in action as a Source Academy runner, including REPL.

PR #1586, opened 2024-03-19
1534 stepper description error for blockexpression

FYL2003:
Fix typo in function explain, from "Empty block statement evaluated" to "Empty block expression evaluated"


martin-henz (approved):
OK

PR #1591, opened 2024-03-20
Created pylib.ts, added functions as built-ins to Python supporting BigInt typing.

JJtan2002:
This pull request adds functions to support binary operators and Math functions on Integers in Python.
pylib.ts contains definitions of functions for binary operators and Math.
These functions are declared as built-in functions for Python 1 in createContext.ts.


martin-henz (approved):
Looks good!

PR #1592, opened 2024-03-21
Bump ESLint dependencies

RichDom2185:
Supersedes #1524.
Supersedes #1526.

New breaking changes are disabled temporarily. Ideally we should follow the ESLint recommendations in order to ease our transition to TS v5, but we don't want to introduce unnecessary merge conflicts at the moment due to the rapid development pace.

This PR just bumps ESLint and its related dependencies to prepare for it.


martin-henz (approved):
looks good

PR #1596, opened 2024-03-21
CSE Machine: Give prelude environments a separate name

### LINKED ISSUE #1595: CSE Machine: Visualization shows unreferenced globals
When running a program, the CSE machine visualization shows unreferenced globals instead of only referenced ones:

### Investigation

It appears the bug was caused by #1584; more specifically, [these changes](https://github.com/source-academy/js-slang/pull/1584/files?diff=split&w=0#diff-8a646412f7b595e811f6f11dd99a63b4f969d9e5558b8839e1346f4955ff83f4) to `interpreter.js` which causes some globals to be referenced by the program environment.

NhatMinh0208:
# Description:

This PR changes program environment creation so that prelude environments are given a separate name (`prelude`) instead of the standard `programEnvironment`.

Fixes #1595.

# Changes:

- [X] Give prelude environments the name `prelude`


src/cse-machine/utils.ts
```diff
@@ -312,6 +312,26 @@ export const createBlockEnvironment = (
...
+        id: uniqueId()
+      }
+    : {
+        name: 'programEnvironment',
+        tail: currentEnvironment(context),
+        head,
+        id: uniqueId()
+      }
+}
+
```
RichDom2185, line 334: Can we just combine the two and only change the name property to prevent duplication (since only the name property differs)?


```suggestion
  return {
        name: isPrelude ? 'prelude' : 'programEnvironment',
        tail: currentEnvironment(context),
        head,
        id: uniqueId()
      }
}

```

RichDom2185 (commented):
Looks good, thanks for working on this so quickly! I just have one minor comment:

src/cse-machine/utils.ts
NhatMinh0208: Yes that would be better.

RichDom2185 (approved):
LGTM, thanks!

PR #1597, opened 2024-03-22
1108 stepper programs that end with a declaration dont show correct result

FYL2003:
This change solves the issue of not returning undefined when encounter empty program and not returning undefined when declaration appears.


martin-henz (approved):
All good!

PR #1600, opened 2024-03-22
CSE Machine: improve how objects are handled in environments, change id behavior

### LINKED ISSUE #1585: CSE Machine: Improve dummy bindings and add new mainReference property to arrays and functions
Currently, arrays are not drawn inside the environment heap until they are assigned to a variable. The moment an object is added to the stash, it should immediately be drawn inside the environment at the correct environment frame.

Functions (or more specifically, closures) do actually get drawn, but only in [some conditions](https://github.com/source-academy/js-slang/blob/master/src/cse-machine/interpreter.ts#L724). They make use of the fact the raw numbers cannot be used as variable names, so any binding with a key being a number string is considered a "dummy binding" when drawn.

An example of dummy bindings by running the below code with the CSE machine:
```js
const a = [x => x];
a[0] = 1;
```
Also note how the dummy binding reappears in the last step, as dummy bindings never get removed in js-slang.

There is also a need for a `mainReference` property on array and closure objects, that points to the main reference for an array or function, which is either an `Environment` or an `Array` and its correct index. We could create a new Type `ArrayPosition`:
```ts
interface ArrayPosition {
    array: any[];
    index: number;
}
```
to represent the array and its index, and the `mainReference` would then have a type of `Environment | ArrayPosition`.

There are two approaches to how we can handle dummy bindings and `mainReference`:
### Approach 1
- Keep the current behavior of the CSE Machine mostly same, with the only change being that objects in the stash are also drawn in the environment first, using dummy bindings
- Since arrays and functions get removed entirely if variables get re-assigned and there are no more references, dummy bindings will also need to be removed the moment the object gets assigned to an actual variable or inside an array.
- The `mainReference` property on arrays and functions will also need to change regularly as variables get re-assigned.
    - If a dummy value gets assigned to a variable inside another environment frame/array position, then the `mainReference` will change to that environment frame/array position.
    - If a variable containing an object gets re-assigned to something else, and the object does not get deleted because it still have other references, the object's `mainReference` will change to one of the other references if the other references comes from different environment frame/array position. Deciding on which one to choose will be questionable, since we do not track the order in which references get added, so the only way would be to traverse the environment and find the first reference for the object.
- This will allow [issue 2700](https://github.com/source-academy/frontend/issues/2700) to be solved by providing a primary reference for which objects should be drawn at.
### Approach 2
- We do not delete any dummy bindings, and the `mainReference` will only have a type of `Environment`, since all objects are initially created inside an environment, and only gets assigned to an array afterwards
    - In fact, we could rename this property to `environment` since closures already have this property, and add this property to arrays as well
    - For dummy bindings, instead of assigning all objects to a numeric key property inside the frame, we could have a separate array property in `Environment` called `objects` that stores the objects instead, since the key in each dummy binding is useless
- When we draw the actual values on the frontend, we do a deep search for all values inside the environment frame to find the appropriate spot the draw the object.
    - This is only for UI purposes, so code examples like this:
    ```js
    const a = [0, x => x, 2];
    const b = pair(3, pair(4, null));
    ```
    will have the objects drawn from the array instead of being drawn next to the environment frame.
    - Since objects never get deleted, we know that the order of the bindings is the order in which the objects were created, so we can make a reasonable guess to which variable or array index to draw the object at.
    - However, in complex situations where on object has multiple references in multiple arrays in the same frame at different locations, it is not possible to know which array reference should we draw the object at, so we can only really guess the first one.
- This will allow [issue 2705](https://github.com/source-academy/frontend/issues/2705) to be solved.
- This will allow [issue 2700](https://github.com/source-academy/frontend/issues/2700) to be solved partially, for simple cases where there are not too many references for an object. However, objects can still jump within the environment frame for complex cases, but it is guaranteed that the object will stay in the frame it was created in.

### LINKED ISSUE #1589: CSE Machine: Keep id of environments, arrays & closures constant across different steps
The `Environment` and `Closure` objects currently have an `id` property that is not very useful. As the user run the CSE Machine and goes to the next steps, the `id` will always change between frames, since js-slang using the `uniqueId()` function from lodash to set the `id`.

It would be nice if `id` stays the same for the same environments, arrays and closures across different steps of a program. This would mean that arrays should also have an `id` property. Having the `id` stay the same could pave the way for better performance on the frontend (through re-use and mutation of objects that have the same id), and allow for more complex CSE machine animations inside the environment regarding movement of multiple different objects.

An idea would be to have a class variable that resets to `0` when the program first starts. When any environment, array and closure is first created, we assign that variable to the `id`, and increment it. Since the program execution from the beginning is consistent across different steps, the `id` values should also remain consistent for the same environments, arrays and closures.

CZX123:
### Description
This PR revamps how objects (i.e. arrays and closures) are handled in each environment, by giving each object an `environment` property and giving environments a `heap` property to store all objects. When any object is created, it is immediately added to the heap of the current environment. Fixes #1585, and allows frontend issues [\#2700](https://github.com/source-academy/frontend/issues/2700) and [\#2705](https://github.com/source-academy/frontend/issues/2705) to be solved.

This PR also adds the `id` property to arrays, and changes the id behavior of environments, arrays and closures. Instead of calling `uniqueId()` from lodash, a new property is created in the runtime context called `objectCount`, and a new `uniqueId` function takes this value and increments it. Fixes #1589, and could improve frontend issue [\#2701](https://github.com/source-academy/frontend/issues/2701) by allowing for better animations involving different objects in the CSE machine.

#### Additional Clarifications
- A heap can only store arrays and closures, and that includes arrays created from in-built functions like `list`, as well as closures created in the prelude like `equal`. In-built functions are **not** stored in the heap.
- The heap only works when run with the CSE Machine, the environments that is outputted by the interpreter will still have heaps (for type compatibility), but the heap storages will always be empty, as the interpreter does not require the heap for visualization just yet.
- Id behavior is changed for both the interpreter and CSE Machine, and lodash's `uniqueId` function is no longer used.
   - Since id behavior is now consistent, test cases which expect any string as an id, will no longer have to do so.

### Changes
- [X] Added new type `Array`, which is an array that also has special properties `id` and `environment`
- [X] Added new type `HeapObject`, an alias for `Array | Closure`
- [X] Added new class `Heap`, which contains a `Set` of `HeapObject`s.
- [X] Added new property `heap` inside `Environment`
- [X] Added new property `objectCount` inside `Context.runtime`
- [X] Added new `uniqueId` function that uses the value from `context.runtime.objectCount`
- [X] Added new test cases to test the heap and unique ids of each object
- [x] Changed existing test cases which expect any id string, to now match specific id strings


martin-henz (approved):
Neat and clear implementation!

PR #1601, opened 2024-03-22
1588 accidental unpacking during evaluating the if statement

FYL2003:
Before this change, the if statement inside a block statement is not value producing, details in the following video
https://github.com/source-academy/js-slang/assets/97390773/d1300c04-df1f-4f9c-8a50-9b0dcabe3d69

After the change, the if statement inside a block statement is value producing and will not accidentally unpack itself.

https://github.com/source-academy/js-slang/assets/97390773/c39527cb-06f1-4e5a-ae9e-bce52a26577e


src/createContext.ts
```diff
@@ -11,9 +11,9 @@ import { list_to_vector } from './stdlib/list'
 import { listPrelude } from './stdlib/list.prelude'
 import { localImportPrelude } from './stdlib/localImport.prelude'
 import * as misc from './stdlib/misc'
-import * as pylib from './stdlib/pylib'
 import { nonDetPrelude } from './stdlib/non-det.prelude'
 import * as parser from './stdlib/parser'
+import * as pylib from './stdlib/pylib'
```
martin-henz, line 16: Is this a mistake? I don't see why you would add pylib.

This might be the reason why CI fails.

martin-henz (changes requested):

src/createContext.ts
FYL2003: I did not do anything with this file, maybe my last git pull add it accidentally. So I need to remove this line and submit again?

martin-henz: Yes, please remove the line from your PR.

RichDom2185: It's not a mistake, pylib wasn't "added", the import line was just rearranged (see the deleted line above the added line) - probably due to linter. Just rearrange it back, please. Thanks!

martin-henz (approved):
All good now.

Issue #1551, opened 2024-02-20
Stepper: does not show "name ... not declared"


martin-henz:
The stepper runs validation, but not display the error. Example:
```
some_undeclared_name;
```
should show an error in REPL in all runners (transpiler, stepper, CSE machine): name ... not declared

Instead, the error is caught in stepper.ts (line 3211).

PR #1608, opened 2024-03-25
Stepper: Catch and show "name not declared" error 

hanscau:
Fix for #1551 

Catch RuntimeSourceError and display it on the stepper

Add appropriate test for "name not declared"


src/stepper/__tests__/stepper.ts
```diff
@@ -9,6 +9,10 @@ function getLastStepAsString(steps: [substituterNodes, string[][], string][]): s
   return codify(steps[steps.length - 1][0]).trim()
 }
 
+function getExplanation(steps: [substituterNodes, string[][], string][]): string {
+  return steps.map(x => x[2]).join('\n')
```
RichDom2185, line 13: What is the `[2]` here? Can we add a comment?

RichDom2185 (commented):
Just one minor readability/code clarity nit below:

RichDom2185 (approved):
LGTM, thanks!

src/stepper/__tests__/stepper.ts
hanscau: Left a comment for now, in the future all the indexes of steps should be accessed through a constant. Or maybe steps should just be an object array instead for better clarity...

Issue #1495, opened 2023-10-17
Native: Imported modules do not carry over in different evaluation steps


sayomaki:
When importing module functions and not using it directly (in the current program), this will lead to them being undeclared, i.e. using it in another program sharing the same context will lead to an error with name being undeclared.

However, this is not an issue/the case with the default source versions.

This causes issues when trying to execute module functions under REPL in different cells, or in the case of assessments, any module functions imported in the prepend chunk will not be declared and imported properly in the student's running program. (see examples below)

### Affected Chapters & Variants
- Source 1 to 4, Native variant

### Affected Modules
- All modules

#### 1. Playground
Run these 2 chunks of code in separate cells (in the REPL), and you will see that the only the first cell works correctly.
```js
import { noise_sound } from "sound";
noise_sound(2);
```
---
```js
noise_sound(2); // doesn't work
```

<img width="330" alt="Screenshot 2023-10-17 at 6 05 21 PM" src="https://github.com/source-academy/js-slang/assets/26355099/7cee5378-83e1-4b72-b341-144f1912a7d2">

#### 2. Assessment
Create any assessment with prepend importing a module, then set the interpreter variant to native.
You will see that the imported module functions do not work in the assessment.

### Workarounds
- For now, when using Source 1-4 Native, you will have to ensure the imports share the same program block as the actual code
- As for assessments, it will need to be manually imported by the students or given as part of the placeholder template.

martin-henz:
This issue persists as of today.

sayomaki:
The issue where imports do not carry over in assessment prelude and from the main code editor (on the left in playground) has been resolved in #1625, however importing new names in REPL issue still persists. 

The issue title and post has been updated accordingly to reflect the updated issue.

PR #1625, opened 2024-03-30
Source Native: Fix imports not working from prelude

### LINKED ISSUE #1500: Source 3 native lost imported functions when switching context from prepend to student's code
https://github.com/source-academy/frontend/issues/2419

In Source 3 Native, the import executed in the prepend for an assessment can NOT keep in student's code. You have to re-import the functions again in student's code or use another constant to hold the imported value in the prepend.

sayomaki:
As mentioned in issues #1500 and #1495, when using Source (any chapter) Native variant, importing module functions and constants in prelude (for assessments) or when using those names in the REPL (after running the main code, shown below) fails to work correctly.

It appears that the exported functions do not get picked up by `getGloballyDeclaredIdentifiers` (in the `transpileToFullJS` function) and thus get lost when running in the native variant. This causes the transpiler unable to recognise that the symbol has been imported and hence throws an error that the name is undeclared.

During the debugging process for this, there are two more bugs to related to the native variant when using the REPL (more details will in the respective issues that will be created), but other than that, the code seems to work and tested to be working locally in both the example (as shown below) and in assessments with prelude.

Resolves #1500.

**Current Behaviour (in REPL):**
Screenshot 2024-03-31 at 03-19-37 Source Academy

**Expected/Fixed Behaviour (in REPL):**
Screenshot 2024-03-31 at 03-42-41 Source Academy

### DIFF
--- a/src/transpiler/transpiler.ts
+++ b/src/transpiler/transpiler.ts
@@ -588,6 +588,7 @@ async function transpileToFullJS(
     globalIds.native
   )
 
+  program.body = (importNodes as es.Program['body']).concat(otherNodes)
   getFunctionDeclarationNamesInProgram(program).forEach(id =>
     context.nativeStorage.previousProgramsIdentifiers.add(id)
   )


leeyi45 (approved):
LGTM!

PR #1627, opened 2024-04-01
1554 stepper does not get stuck when it should instead it crashes

### LINKED ISSUE #1587: Stepper: ReferenceError is not shown, instead it get stuck
image

Stuck, instead it should show the error:

image

hanscau:
Handles errors thrown by builtin functions 

Fix various minor issues with errors that are thrown

Fixes #1587


src/stepper/stepper.ts
```diff
@@ -1741,22 +1741,31 @@ function reduceMain(
...
-            return [builtin[(callee as es.Identifier).name](...args), context, paths, explain(node)]
+          } catch (error) {
+            throw new errors.BuiltInFunctionError(callee, error.message)
           }
-          return [
-            builtin.evaluateModuleFunction((callee as es.Identifier).name, context, ...args),
-            context,
-            paths,
-            explain(node)
-          ]
```
hanscau, line 1759: Basically wrapped the part of the code that deals with builtin function with a trycatch block and wrap the internal error with a RuntimeSourceError

martin-henz (approved):
Looks great!

PR #1628, opened 2024-04-01
1622 stepper does not evaluate arguments when it should

hanscau:
Fix the issue by rearranging the reduction steps of function call

Refactor code for easier readability 

Added comments

Added tests


src/stepper/stepper.ts
```diff
@@ -1683,6 +1683,29 @@ function reduceMain(
...
+          ]
+        } else if (
+          currentArg.type === 'Identifier' &&
+          !(currentArg.name in context.runtime.environments[0].head)
+        ) {
+          throw new errors.UndefinedVariable(currentArg.name, currentArg)
+        }
+      }
+
+      //Reduce callee until it is irreducible
```
martin-henz, line 1708: Should we not reduce the callee, before the arguments?
Example:
```
((x => x)(x => x))(1 + 2);
```
currently evaluates the callee first, and then 1 + 2. That is the correct order.

src/stepper/stepper.ts
```diff
@@ -1683,6 +1683,29 @@ function reduceMain(
...
+            ast.callExpression(callee as es.Expression, reducedArgs as es.Expression[], node.loc),
+            cont,
+            path,
+            str
+          ]
+        } else if (
+          currentArg.type === 'Identifier' &&
+          !(currentArg.name in context.runtime.environments[0].head)
+        ) {
+          throw new errors.UndefinedVariable(currentArg.name, currentArg)
```
martin-henz, line 1704: Should this be UnassignedVariable? We shouldn't ahve UndefinedVariable anywhere, because we use the validator.

martin-henz (changes requested):

src/stepper/stepper.ts
hanscau: Yeap, it is redundant. Will remove it

martin-henz (approved):
Thanks for fixing the issues.

PR #1637, opened 2024-04-03
Remove Synchronous Module Loading

### LINKED ISSUE #1348: Use asynchronous GET requests / Replace `xmlhttprequest-ts` with `fetch` API for module loading

leeyi45:
This PR will officially deprecate and remove any synchronous module loading code. All module code is completely asynchronous.

The main idea is to leverage the native `import()` functions that browsers provide. However this is not available in Jest or NodeJS, so it is necessary to detect the environment and switches to using `require` when not running in the browser.

And even then, though Jest will have no problems with it, some testing has shown that NodeJS will refuse to load JSON files over the web (i.e trying to import `http://localhost:8022/modules.json` throws `MODULE_NOT_FOUND`), so JSON files are still loaded using `fetch`.

This does introduce complexity into how our tests are written (and perhaps in some other areas that I cannot foresee), so maybe it would be best to simply use the `fetch` then `eval` method, but why not use `import()` where we can?

This is fairly low priority, so it doesn't really need to merged ahead of anything else

Will close #1348


RichDom2185 (approved):
LGTM, thanks a lot!

PR #1639, opened 2024-04-03
CSE Machine: Fix condition for pushing environment instruction during function application

### LINKED ISSUE #1635: CSE machine: unnecessary env on the bottom of control
We should avoid pushing env instructions when control is empty:
Screenshot 2024-04-02 at 6 15 23 PM

NhatMinh0208:
### Description

For some reason, the code handling function application has been pushing environment instructions even when the control stack is empty.

This PR rectifies that.

Fixes #1635.

### Changes

- [X] environment instruction is no longer pushed when control stack is empty


martin-henz (approved):
LGTM

PR #1645, opened 2024-04-06
CSE Machine fixes for variadic functions, closures & heap

CZX123:
### Description
This PR fixes many issues related to the new heap addition.

### Changes
- Created duplicate `Closure` class for use in the CSE Machine
- Made `closureToJS` use CSE Machine instead of the old interpreter
  - This will allow the built-in function `apply_in_underlying_javascript` to correctly display the environment and ensure objects get added to the heap correctly
- Added new `declaredName` property in `Closure`, which stores the name of the constant variable the closure is assigned to
  - This is used for `apply_in_underlying_javascript` to create environment frames with better names
  - `functionName` will now always be in the form of `x => ...`
- Correctly added the argument array of a variadic closure to the heap
- Added `environment` property to functions returned from calling `stream`
- Added `move` method in `Heap` for use in the frontend
- Moved `Stack` and `IStack` to a separate file to avoid circular dependency issues


src/cse-machine/closure.ts
martin-henz, line 1: Great to see "closure" in this folder instead of using the legacy interpreter implementation.

martin-henz (approved):
Good work. Quite a few cleanups and improvements.

PR #1649, opened 2024-04-08
Amended stepper doc to new spec

hanscau:
Updated Stepper Doc to reflect the current implementation of stepper which now returns the return expression in a block expression instead of reducing it first.


hanscau:
image
 
Editted entry for easier review

martin-henz (approved):
Looks good!

Issue #1124, opened 2021-09-13
Bug in display_list


chownces:
`display_list` does not show the correct list notation, when we apply `display_list` to the suggested solution for the `subsets` question in Studio 6

`subsets(list(1,2,3));  // omitted subsets implementation`

![image](https://user-images.githubusercontent.com/53928333/133044157-503700af-0a01-432e-a248-1ef981b55f2d.png)

Cloud7050:
Oh nice, an issue has already been opened. I was confused by this, and messed around with my answer, the answer that was gone through, and bunches of test code to trying to figure it out. It seems to have to do with some odd mix of `append()`, `map()` and `x` being abstracted in this case. Small example to illustrate/replicate issue:

```js
const x = list(
    list(2),
    list(2, 3)
);
display_list(append(
    x,
    map(l => pair(1, l), x)
));
display_list(append(
    list(
        list(2),
        list(2, 3)
    ),
    map(l => pair(1, l), x)
));
```

parnikkapore:
Triaged this for fun for a bit and reduced the small example further I can't get it any smaller while still reproducing the issue:

```js
const x = list(
    list(2, 3)
);

display_list(append(
    x,
    map(l => pair(1, l), x)
)); // list(list(2, 3), [1, list(2, 3)]) - Wrong

display_list(append(
    list(list(2, 3)),
    map(l => pair(1, l), x)
)); // list(list(2, 3), list(1, 2, 3)) - Correct
```

martin-henz:
This bug is still there. It was rediscovered by a CS1101S student today.

RichDom2185:
Bug also present for the following sample code (as reported by Grace):

```js
const xs = list(2, 3);
display_list(list(pair(1, xs), pair(4, xs)));
```

PR #1651, opened 2024-04-08
Separate program values from their representations

s-kybound:
This PR extends the Result type with an optional Representation field, as long as a Result Value is present. This is done to allow alternate languages (aside from JS) to provide their own data representations for values whilst preserving the actual data passed. 

For example, arrays in JavaScript are represented as `[1, 2, 3, 4]`. The corresponding vector (array equivalent in scheme) should be represented as `#(1 2 3 4)`. This PR allows the new representation to be stored separately from the value returned (which is still a JS array).

After a program is run, the result will be mapped by the mapResult() function, which checks the language used in the context to run the appropriate result mapper. Ideally, languages will add their own logic into the mapResult() function (located in src/alt-langs/mapper.ts).

This PR is also non-invasive. If an alternate representation is not needed (the case for Source/JS/TS) or not implemented, js-slang defaults to using the original representation for the value.

(as a side note, this PR also tags all scheme pairs so that they are distinguishable from scheme vectors (despite both being represented using arrays) - a possible enhancement that can be brought to js-slang too?)


src/repl/repl.ts
```diff
@@ -35,7 +36,9 @@ function startRepl(
           eval: (cmd, unusedContext, unusedFilename, callback) => {
             runInContext(cmd, context, options).then(obj => {
               if (obj.status === 'finished' || obj.status === 'suspended-non-det') {
-                callback(null, obj.value)
+                // if the result is a representation, display the representation.
+                // else, default to standard value representation.
+                callback(null, obj.representation ? obj.representation : obj.value)
```
RichDom2185, line 41: Should we check for `!==` undefined here instead of truthy/falsy?

src/repl/repl.ts
```diff
@@ -97,7 +102,7 @@ function main() {
         'default'
       ],
       ['h', 'help', 'display this help'],
-      ['e', 'eval', "don't show REPL, only display output of evaluation"]
+      ['e', 'eval', "don't show REPL, only dispResultlay output of evaluation"]
```
RichDom2185, line 105: Typo

src/types.ts
```diff
@@ -275,6 +276,10 @@ export interface Finished {
   status: 'finished'
   context: Context
   value: Value
+  representation?: Representation // if the returned value needs a unique representation,
+  // (for example if the language used is not JS),
+  // the display of the result will use the representation
+  // field instead
```
RichDom2185, line 282: I'm thinking if we can even use this to handle #1124

src/repl/repl.ts
s-kybound: Thanks!

src/repl/repl.ts
s-kybound: Will do!

src/types.ts
s-kybound: Sorry, could you elaborate? 😅 Representations currently only concern the final return value of a program - for now, for functions like display() or display_list(), the output remains unaffected

RichDom2185 (approved):
LGTM, thanks!

martin-henz (approved):
Neat integration of Scheme into js-slang.

PR #1653, opened 2024-04-08
1612 stepper accidental unpacking during evaluating the if statement in block expression

FYL2003:
respective test cases got updated


martin-henz (approved):
I think this is the right approach.

PR #1656, opened 2024-04-09
Use fetch for docs loading exclusively

leeyi45:
Import attributes don't really have as much support across browsers as we'd like. This PR removes the `import()` loader and simply uses `fetch()` for loading module documentation


martin-henz (approved):
LGTM.

PR #1657, opened 2024-04-09
CSE: Fix buggy stream function checking

CZX123:
Sometimes, built-in functions were erroneously marked as a stream function, and had the `environment` property added to them.

### Changes
- Added `isStreamFn` to properly check for a non-closure function that returns a stream


martin-henz (approved):
LGTM

PR #1658, opened 2024-04-09
Update docs deployment workflow runner version

sayomaki:
Currently, we are using `peaceiris/actions-gh-ages@v3` to deploy the documentation for js-slang, however this workflow action uses Node 16 under the hood, which is already EOL.

A new version update for the workflow action has been released, and uses Node 20 instead. As such, this PR bumps the version of the workflow action used to the latest version.

More information about the newer workflow action here:
https://github.com/peaceiris/actions-gh-pages/blob/v4.0.0/CHANGELOG.md


RichDom2185 (approved):

martin-henz (approved):
LGTM. Let's try this.

PR #1673, opened 2024-04-14
Restore scm-slang stdlib

s-kybound:
This pull request expands scm-slang to use preludes for list procedures such as map, filter and fold. This allows students to reason about such functions, and more importantly, allows procedures which make use of continuations to work as expected with such higher order functions.

Additionally, it expands the math procedures available to scm-slang, allowing for procedures such as sin, cos, tan, exp, log, and the like, which work both on normal numbers (similar to normal JavaScript) but also on complex numbers. With this, our R7RS implementation can now properly operate on complex numbers.

Much of the original standard library of scm-slang, available in the original scm-slang implementation, has also been restored, bringing scm-slang back to "full power".


RichDom2185 (approved):
LGTM, thanks!

PR #1677, opened 2024-04-15
Add Math constants to Python

JJtan2002:


RichDom2185 (approved):

RichDom2185:
LGTM, thanks!

martin-henz (changes requested):
Let's match Python's math library. The constant log10e is not included there. Let's exactly match Python's math library, just like what we do for JavaScript.
```
(base) henz@adminsocs-MacBook-Pro-2 js-slang % python             
Python 3.10.12 (main, Aug  8 2023, 12:44:36) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> from math import (pi)
>>> pi
3.141592653589793
>>> from math import (log10e)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'log10e' from 'math' (/Users/henz/.pyenv/versions/3.10.12/lib/python3.10/lib-dynload/math.cpython-310-darwin.so)
>>> 
```

docs/specs/source_python_math.tex
```diff
@@ -9,7 +9,9 @@ \subsection*{MATH Library}
 \texttt{Math} library, see\\
 \href{https://www.ecma-international.org/ecma-262/9.0/index.html\#sec-math-object}{\color{DarkBlue}ECMAScript Specification, Section 20.2}. Examples:
```
martin-henz, line 10: Refer to Python documentation here, not to ECMAScript. Make sure that all Python math functions and only the Python math functions are available as math_...

martin-henz (changes requested):

docs/specs/source_python_math.tex
```diff
@@ -5,9 +5,9 @@ \subsection*{MATH Library}
 \begin{itemize}
 \item \href{https://sourceacademy.org/sicpjs/1.1.4\#p8}{\lstinline{math_}$\textit{name}$},
 where $\textit{name}$ is any name specified in the
-JavaScript
+Python
 \texttt{Math} library, see\\
-\href{https://www.ecma-international.org/ecma-262/9.0/index.html\#sec-math-object}{\color{DarkBlue}ECMAScript Specification, Section 20.2}. Examples:
+\href{https://docs.python.org/3.7/library/math.html}{\color{DarkBlue}Python 3.7.17 Documentation}. Examples:
```
martin-henz, line 10: Don't you want to go for the latest stable release? https://docs.python.org/3/library/math.html

Are you implementing all functions in the Python math library? How about math.isnan, or math.factorial. If any functions are missing, please create an issue where you list what needs to be done to make the implementation consistent with the specs.

martin-henz (changes requested):

martin-henz (approved):
LGTM

PR #1687, opened 2024-04-22
CSE Machine: Avoid pushing unnecessary env instructions

### LINKED ISSUE #1682: CSE: Avoid unnecessary env instructions
Consider:
Screenshot 2024-04-17 at 1 08 56 PM

The env instructions are not needed because there are only * instructions on control. More generally, an env instruction can be avoided if there are no statements or expressions on the control from the top of the control stack until the next env instruction. "Statements or expressions" includes instructions that have statements or expressions inside, such as branch instructions.

We should also avoid environment instructions when the environment does not change: application of nullary function that is declared in the current environment.
Screenshot 2024-04-23 at 7 56 43 AM
Screenshot 2024-04-23 at 7 56 58 AM

The pedagogical strategy is to introduce "complications" such as env instructions only when needed. Simple programs have simple explanations and complex programs may require more complex explanations.

DiligentPenguinn:
### Description
This PR implements simple logic to avoid pushing `envInstr` when not needed. Specifically, `envInstr` can be avoided if there are no environment-dependent commands on the control. A command is environment-dependent if its evaluation depends on the context given by the current environment. 
Implements and resolves #1682

### Type of change
- [ ] This change requires a documentation update

### How to test
The [following programs](https://share.sourceacademy.org/b37xg) should run with no environment instructions being pushed

### DIFF
--- a/src/cse-machine/__tests__/__snapshots__/cse-machine-runtime-context.ts.snap
+++ b/src/cse-machine/__tests__/__snapshots__/cse-machine-runtime-context.ts.snap
@@ -1,5 +1,2805 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Avoid unnescessary environment instruction 1 1`] = `
+Control {
+  "storage": Array [
+    Object {
+      "instrType": "BinaryOperation",
+      "srcNode": Node {
+        "end": 53,
+        "left": Node {
+          "arguments": Array [
+            Node {
+              "end": 48,
+              "left": Node {
+                "end": 46,
+                "loc": SourceLocation {
+                  "end": Position {
+                    "column": 7,
+                    "line": 4,
+                  },
+                  "start": Position {
+                    "column": 6,
+                    "line": 4,
+                  },
+                },
+                "name": "n",
+                "start": 45,
+                "type": "Identifier",
+              },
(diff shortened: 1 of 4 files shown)


martin-henz (approved):
This is a safe strategy to avoid unnecessary env instructions. It could be further generalized, but this is good enough for now. The advantage is that it is easy to do: doesn't complicate the machine very much.

DiligentPenguinn:
@martin-henz yes, I was thinking that further generalization would add extra complication to what is a pedagogical strategy. If this is desired then I can further work on it in the future to be more exhaustive.

martin-henz:
The `canAvoidInstr` function is extremely inefficient and slows down the deployment each time by almost one hour! @RichDom2185 and @sayomaki commented:

the code is not very efficient at all, it seems; we can just early return instead of iterating through the entire control every single step of the cse machine

I'm pretty sure this alone accounts for most of the 40minutes

same thing for the switch case logic, lots of unnecessary computation

@sayomaki :
consider using javascript's Array.some() or Array.every() instead

@RichDom2185 :
yes that also works, no need for a for loop, though I think the logic can be improved further

because the same computation is being repeated at every step

it should have been memoised

just create a new field in the control item and calculate it once only instead of calculating n times for each item where n is the number of control items

Issue #1698, opened 2024-04-27
CSE Machine: `canAvoidInstr` is too slow


martin-henz:
The `canAvoidInstr` function is extremely inefficient and slows down the deployment each time by almost one hour! 

@RichDom2185 and @sayomaki commented:

the code is not very efficient at all, it seems; we can just early return instead of iterating through the entire control every single step of the cse machine

I'm pretty sure this alone accounts for most of the 40minutes

same thing for the switch case logic, lots of unnecessary computation

@sayomaki :
consider using javascript's Array.some() or Array.every() instead

@RichDom2185 :
yes that also works, no need for a for loop, though I think the logic can be improved further

because the same computation is being repeated at every step

it should have been memoised

just create a new field in the control item and calculate it once only instead of calculating n times for each item where n is the number of control items

PR #1699, opened 2024-04-27
Temporary fix to speed up `canAvoidInstr`

DiligentPenguinn:
Following the fact that the check `canAvoidInstr` implemented in previous PR #1687 is too slow, this PR implements a temporary fix by caching the result of `isEnvDependent` as a field for each `ControlItem`

### DIFF
--- a/src/cse-machine/interpreter.ts
+++ b/src/cse-machine/interpreter.ts
@@ -442,9 +442,11 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
     // Push ENVIRONMENT instruction if needed - if next control stack item
     // exists and is not an environment instruction, OR the control only contains
     // environment indepedent item
-    if (next 
-      && !(isInstr(next) && next.instrType === InstrType.ENVIRONMENT)
-    && !canAvoidEnvInstr(control)) {
+    if (
+      next &&
+      !(isInstr(next) && next.instrType === InstrType.ENVIRONMENT) &&
+      !canAvoidEnvInstr(control)
+    ) {
       control.push(instr.envInstr(currentEnvironment(context), command))
     }
 
@@ -955,10 +957,12 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
 
       // Push ENVIRONMENT instruction if needed - if next control stack item
       // exists and is not an environment instruction, OR the control only contains
-    // environment indepedent item
-      if (next 
-        && !(isInstr(next) && next.instrType === InstrType.ENVIRONMENT)
-      && !canAvoidEnvInstr(control)) {
+      // environment indepedent item
+      if (
+        next &&
+        !(isInstr(next) && next.instrType === InstrType.ENVIRONMENT) &&
+        !canAvoidEnvInstr(control)
+      ) {
         control.push(instr.envInstr(currentEnvironment(context), command.srcNode))
       }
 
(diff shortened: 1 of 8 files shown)


RichDom2185 (commented):
Looks good for a hotfix, thanks!

Though I'd remove `isEnvDependent` entirely and move the logic to the node creator function instead. What do you think? Do you foresee any more changes to this logic? You mentioned this is just a basic check (for the demo the other day) right? What future more "advanced" checks might there be?

Also, can I check what's left of this PR before it can be marked as ready?

Thanks!

DiligentPenguinn:
@RichDom2185 Yes, I think that moving the logic to the node creator function is what should ultimately be done, but I won't be able to work on it very soon (due to exams). Until then I think it is still considerably slow with this fix though so I'm trying other ways to speed it up a bit more first. If you think it's fine for now then I'll mark it as ready

DiligentPenguinn:
@RichDom2185 I just pushed approach 2, this seems to fix the speed issue, and I think it's closer to what would be done when we move the logic to the node creator function eventually.

src/cse-machine/interpreter.ts
```diff
@@ -98,15 +98,34 @@ type CmdEvaluator = (
...
   public constructor(program?: es.Program | StatementSequence) {
     super()
-
+    this.numEnvDependentItems = 0
     // Load program into control stack
     program ? this.push(program) : null
   }
 
+  public canAvoidEnvInstr(): boolean {
+    return this.numEnvDependentItems == 0
```
RichDom2185, line 110: Let's use strict equality here

```suggestion
    return this.numEnvDependentItems === 0
```

src/cse-machine/interpreter.ts
```diff
@@ -98,15 +98,34 @@ type CmdEvaluator = (
...
+  }
+
+  public pop(): ControlItem | undefined {
+    const item = super.pop()
+    if (item != undefined && isEnvDependent(item)) {
```
RichDom2185, line 115: Ditto.

```suggestion
    if (item !== undefined && isEnvDependent(item)) {
```

src/cse-machine/interpreter.ts
```diff
@@ -98,15 +98,34 @@ type CmdEvaluator = (
...
+      this.numEnvDependentItems--
+    }
+    return item
+  }
+
   public push(...items: ControlItem[]): void {
     const itemsNew: ControlItem[] = Control.simplifyBlocksWithoutDeclarations(...items)
+    // testing
```
RichDom2185, line 123: Can we remove this comment?

src/cse-machine/utils.ts
RichDom2185, line 1: I'm good with this as long as it is a temporary hotfix. Please create an issue to migrate this to the node creator in the future.

Thanks!

RichDom2185 (commented):
Just some minor nits, otherwise all good, thanks a lot!

src/cse-machine/interpreter.ts
RichDom2185: Resolved in eca7bbf

src/cse-machine/interpreter.ts
RichDom2185: Resolved in eca7bbf

src/cse-machine/interpreter.ts
RichDom2185: Resolved in eca7bbf

src/cse-machine/utils.ts
RichDom2185: See #1702.

RichDom2185:
Part of #1698 (does not close it fully due to the comment above).

PR #1701, opened 2024-04-29
fixing unicode errors in build: more fixes needed

martin-henz:


RichDom2185 (changes requested):
We seem to be missing the leading backslashes of `\Theta` and `\Omega`.

Can we also add a check to the CI workflow?

martin-henz:
> We seem to be missing the leading backslashes of `\Theta` and `\Omega`.
> 
> Can we also add a check to the CI workflow?



> We seem to be missing the leading backslashes of `\Theta` and `\Omega`.
> 
> Can we also add a check to the CI workflow?

The backslashes wouldn't lead to a good output. We could escape these things by saying ^$\Theta$^ to get a nice PDF.

But then the json output for autocomplete would look really ugly (unless we do some more magic). Also jsdoc would look bad.

So at the moment, docs/lib/list.js is used for:
* PDF
* jsdocs
* autocomplete
and we just tolerate a "Theta" appearing everywhere, to keep things simple.

RichDom2185:
> We seem to be missing the leading backslashes of `\Theta` and `\Omega`.
> 
> Can we also add a check to the CI workflow?

Hi Prof @martin-henz, I've fixed these, though I can't check as I don't have LaTeX installed locally. Could you verify that it's correct now?

RichDom2185:
> So at the moment, docs/lib/list.js is used for:
> 
> * PDF
> * jsdocs
> * autocomplete
> 
> and we just tolerate a "Theta" appearing everywhere, to keep things simple.

Replacing `Θ` with `&#1012` works nicely in JSDoc. How about the others?

martin-henz:
> > So at the moment, docs/lib/list.js is used for:
> > 
> > * PDF
> > * jsdocs
> > * autocomplete
> > 
> > and we just tolerate a "Theta" appearing everywhere, to keep things simple.
> 
> Replacing `Θ` with `&#1012` works nicely in JSDoc. How about the others?

The file list.js is currently attached verbatim (using lstlisting in LaTeX) to source_2.pdf.

That means it will look ugly (illegible).

On the longer run, we should find a way to attach list.js in a nicer way to source_2.pdf.

I suggest we go ahead with just "Theta" (and make an issue) for now.

RichDom2185:
> I suggest we go ahead with just "Theta" (and make an issue) for now.

Made the issue in #1705

RichDom2185 (approved):

PR #1704, opened 2024-05-06
Miscellaneous repository updates

sayomaki:
- [x] ~~Add documentation compile check to CI, as mentioned in #1701~~ Completed in #1701
- [x] Fix pull request template not working (add template similar the format in the frontend repo)
- [x] Reduce the pre-push hook run time (removed building and testing of code, as we have the CI to do that now) and instead only check format & linting.
- [x] Remove pre-commit hook as it is handled by the pre-push hook already. For users who are looking to keep their code well-formatted during coding, the use of IDE provided features will suffice.


.github/workflows/nodejs.yml
```diff
@@ -39,6 +39,7 @@ jobs:
     - run: yarn build
     - run: yarn format:ci
     - run: yarn eslint
+    - run: yarn run jsdoc prepare
```
RichDom2185, line 42: This will always fail as the dependencies are not installed. I've fixed this in #1701 already, can just remove?

RichDom2185 (changes requested):

RichDom2185 (approved):
LGTM, thanks!

PR #1706, opened 2024-05-15
CSE Machine: Avoid pushing unnecessary env instructions (exhaustive version)

DiligentPenguinn:
### Description
This PR aims to implement an exhaustive version of PR #1699 and #1687 .
A `ControlItem` is environment-dependent if its evaluation depends on the current environment, OR its evaluation introduces environment-dependent items onto the control.
At each step, if the current control is free of environment-dependent items then we can avoid pushing environment instruction.
### Type of change

- [ ] Documentation update

### Some of my concerns
1) How exhaustive should the logic be? I think it’s enough to cover all node types and instructions allowed to be used in Source. Any remaining node types can default to being environment-independent

2) I initially wanted to place the logic in the AST file along with statementSeqTransform but I realized the logic needed to work for control items which can be either nodes or instructions. Instructions are only defined within the CSE Machine context so currently I place the logic there. 

3) I modified the test cases to use `generateCSEMachineStateStream` which allows me to check the number of environment-dependent items at each step, but now the test runs synchronously (because `generateCSEMachineStateStream` runs synchronously). Would this be a problem?


src/cse-machine/interpreter.ts
```diff
@@ -107,12 +107,17 @@ export class Control extends Stack<ControlItem> {
   }
 
   public canAvoidEnvInstr(): boolean {
-    return this.numEnvDependentItems === 0
+    return this.numEnvDependentItems == 0
```
RichDom2185, line 110: Why was this changed from `===` to `==`?

src/cse-machine/interpreter.ts
```diff
@@ -107,12 +107,17 @@ export class Control extends Stack<ControlItem> {
...
+
+  // For testing purposes
+  public getNumEnvDependentItems(): number {
+    return this.numEnvDependentItems
   }
 
   public pop(): ControlItem | undefined {
     const item = super.pop()
-    if (item !== undefined && isEnvDependent(item)) {
+    if (item != undefined && isEnvDependent(item)) {
```
RichDom2185, line 120: Was this change intentional? Note that `!= undefined` matches both (non-`null` and non-`undefined`) but `!== undefined` matches non-`undefined` only, i.e.

```ts
null == undefined; // true
null === undefined; // false
```

RichDom2185 (commented):
Hi, did a quick look and just some questions below

src/cse-machine/interpreter.ts
DiligentPenguinn: That was unintentional, I'll change it back

Issue #1697, opened 2024-04-27
Remove `src/vm/svml-machine.ts`


martin-henz:
Once https://github.com/source-academy/sicp/issues/991 is done, we can retire the SVML machine in `src/vm`. The machine is poorly written (mostly by yours truly).

We shall keep the compiler, which is needed for robotics.

leeyi45:
Resolved by #1745

Issue #1696, opened 2024-04-27
Remove `src/lazy`


martin-henz:
Once https://github.com/source-academy/sicp/issues/989 is resolved, we can retire the languages Source §x Lazy including their documentation.

leeyi45:
Resolved by #1743

Issue #1695, opened 2024-04-27
Remove `src/interpreter/interpreter-non-det.ts` (currently not working anyway)


martin-henz:
Remove all of `src/interpreter` once https://github.com/source-academy/sicp/issues/990 is resolved.

Currently, Source §3 Non-Det is not working. Older version works. E.g. the first example in SICP JS 4.3
https://sourceacademy.nus.edu.sg/sicpjs/4.3.1
works without error (but with spacing issues). (currently deployment of SA@NUS of today)

The newest version in SA.org gives a runtime error:
https://share.sourceacademy.org/s9irv
gives:

main.13f449e4.js:2 RangeError: Maximum call stack size exceeded
This is regardless whether the amb operator is used.

display(1);
gives the same error. Source §3 Non-Det does not work at all, currently.

RichDom2185:
Is this a duplicate of #1476?

martin-henz:
> Is this a duplicate of #1476?

No, it's in addition to #1476. I've clarified the title of this issue.

leeyi45:
Resolved by #1725

Issue #1617, opened 2024-03-29
Source §3 Non-Det: Always crashes


martin-henz:
Source §3 Non-Det currently does not work at all: It always crashes:

<img width="551" alt="Screenshot 2024-03-29 at 8 22 32 AM" src="https://github.com/source-academy/js-slang/assets/30220768/c60d88b2-a5ac-45f3-981b-6f5e568c7770">

CZX123:
This is related to https://github.com/source-academy/frontend/issues/2883, the errors thrown are similar.

leeyi45:
Resolved by #1725

PR #1707, opened 2024-05-22
Testing Refactor

leeyi45:
# TL; DR
- Remove the interpreter (alongside non-det)
- Remove the lazy variant
- Remove `svml-machine.ts`
- Remove remnants of the old type checker
- Make `js-slang` tests more robust and simpler

## What's going on:
Because our testing utilities relied heavily on the interpreter, removing it means touching almost every single part of `js-slang`, hence the ridiculous number of lines changed. I've tried to simplify the testing utilities and refactor repetitive code where necessary without breaking anything I hope. For example, I noticed that the stepper's tests were being run as async tests when `getEvaluationSteps` is no longer asynchronous. Hopefully with these changes it will make such things easier to catch.

I've also configured `tsconfig.prod.json` to exclude `src/utils/testing`. Previously this code was being shipped with the rest of `js-slang`, but given that we only use it for unit testing I don't think it should be included when `js-slang` is being built.

Given that @martin-henz has indicated that other parts of `js-slang` are marked for removal, I decided to remove them here to avoid having to refactor the tests for those parts of `js-slang`.

Please refer to the TODO comments to see what issues need to be addressed before this PR can be merged.

Will resolve these issues:
- #1476,
- #1543
- #1617
- #1695
- #1696
- #1697

Some issues that need to be revisited:
- #701
- #1616


leeyi45:
This remains a draft because there are some critical questions that need to be answered:

1. Since `svml-machine.ts` is removed, we need to specify the behaviour for when `runInContext` is called with `Variant.CONCURRENT`. The variant has been left in since there are parts of the `vm' that still rely on it.
2. String representations of functions in the `cse-machine` always default to arrow functions. Is this desired behaviour? Because currently the native runner and the cse-machine have different behaviour.
3. I am wholly unqualified to update `src/__tests__/environment.ts`. So the snapshot is failing here but I haven't just gone ahead and changed it.

leeyi45:
Minor comments: The different runner types have different error messages (and types) for the same problem. For example, `stream_length(integers_from(0));` throws a `maximum stack size exceeded` with the `cse-machine` (and I think the stepper behaves similarly), while the native runner (which uses the infinite loop detector) throws an InfiniteLoopError with the message `Line 1: The error may have arisen from forcing the infinite stream: function integers_from.`

I think we need to unify these error messages, but that I think that will have to wait until after this PR is done.

src/__tests__/environment.ts
```diff
@@ -18,11 +18,15 @@ test('Function params and body identifiers are in different environment', () =>
...
-  }
+  evaluate(parsed as any as Program, context, {
+    envSteps: 13,
+    stepLimit: 1000,
+    isPrelude: false
+  })
+  // const stepsToComment = 13 // manually counted magic number
+  // for (let i = 0; i < stepsToComment; i += 1) {
+  //   it.next()
+  // }
```
RichDom2185, line 29: Let's remove the commented code.

RichDom2185 (commented):
Wow, thanks a lot for this! I agree with the questions that are to be answered, let's wait for Prof to come back from his break.

I did a _really really quick_ run through and give my in-principle approval – let's merge this PR if possible, I don't see why the SICP issue needs to be resolved first, since they can always retain an older version of js-slang in their `package.json`.

leeyi45:
One more thing: I have no idea what `inspect()` or any of those related things do, so I haven't been able to remove things like the `Scheduler`

leeyi45:
Just bumping this so that we can try to get this stuff settled

RichDom2185:
> Just bumping this so that we can try to get this stuff settled

What's left of the PR?

leeyi45:
> This remains a draft because there are some critical questions that need to be answered:
> 
> 1. Since `svml-machine.ts` is removed, we need to specify the behaviour for when `runInContext` is called with `Variant.CONCURRENT`. The variant has been left in since there are parts of the `vm` that still rely on it.
> 2. String representations of functions in the `cse-machine` always default to arrow functions. Is this desired behaviour? Because currently the native runner and the cse-machine have different behaviour.
> 3. I am wholly unqualified to update `src/__tests__/environment.ts`. So the snapshot is failing here but I haven't just gone ahead and changed it.
> 4. Stepper doesn't actually do type checking: `0 / "a";` doesn't cause it to throw errors. Does something need to be done about this?



> One more thing: I have no idea what `inspect()` or any of those related things do, so I haven't been able to remove things like the `Scheduler`
> 
> Ideally I'd like to remove things like `manualToggleDebugger` since it doesn't seem like we're using it anymore

RichDom2185:
CC: Prof @martin-henz to answer the questions

leeyi45:
I went back to check my 4th point cause it did seem a bit weird to me. I think what I was trying to say is that the stepper will happily evaluate up to the point of a runtime error and `js-slang` returns a non-errored result value (because the error itself is a step)

leeyi45:
Superseded by #1745

PR #1711, opened 2024-07-17
Streamline CSE machine

s-kybound:
### Description

Streamlines the logic behind continuations, removing GENERATE_CONT and RESUME_CONT instructions. Instead, continuations are now generated immediately upon calling call/cc, and continuations are applied immediately upon being called, without need for other instructions. All logic has been moved to CALL.

Also changes the continuation implementation to be less hacky. Instead of appending continuation data (C, S and E) to a function, continuations now have their own class representation.

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Code quality improvements

### Checklist

- [x] I have tested this code


martin-henz (approved):
Looks good. Things become clearer and simpler.

PR #1717, opened 2024-09-09
1712-stepper-bug-that-causes-frontend-crash

hanscau:
### Description

Fix second statement being a ReturnStatement in a BlockStatement. The first statement should be reduced before going to the return statement. 

### Type of change

- [ ] Documentation update

### How to test

Tests already exists, the expected case was just wrong. 

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/stepper/__tests__/__snapshots__/stepper.ts.snap
+++ b/src/stepper/__tests__/__snapshots__/stepper.ts.snap
@@ -6320,6 +6320,894 @@ expmod(4, 3, 5);
     }
   }
 };
+
+{
+  {
+    {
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
hard to follow the complex logic of reduction of blocks. This PR is not making it worse, and is fixing the given issue. So: best we merge this. (Eventually we should consider simplifying the stepper design.)

FYL2003:
Just some ideas: maybe breaking the big chunk of code into small functional pieces(each function into separate ts files) and add more comment on how the code words would help people better understand the code base.

PR #1719, opened 2024-09-09
1718 stepper built in function errors not caught and crashing the frontend

hanscau:
### Description

Built In Function's Error handling throws Errors that the stepper cannot handle. Changed the errors to SourceErrors instead. 

### Type of change

- [ ] Documentation update

### How to test

Test exists, wrong expected answers

### Note

`BuiltInFunctionError` changed to only take in explanation without node to handle all built in function error. These errors without a node should be handled by the parser instead. 

### Checklist

- [x] I have tested this code


martin-henz (approved):
Looks good!

PR #1723, opened 2024-09-14
Fix faulty regex for syntax highlighter

notmeegoreng:
### Description

When a single | or single & (not || or &&) is used, not as the final character in the line nor in a multiline comment (but a single line comment still has the bug), the syntax highlighting fails.

Starting from 
- the end of a single numeral literal at the start of the line o,
- or the latest occurrence of the character doubled (| => ||, & => &&),
- otherwise the start of the line,

it highlights the entire line up to the | and one character after in pink, the colour used for `new` and `with`.

Characters past one after the single | or & would be highlighted normally even if they are part of a comment.

We can observe this on the playground here: https://share.sourceacademy.nus.edu.sg/syntaxhighlightingbug

I identified the issue as a faulty regex here: https://github.com/source-academy/js-slang/blob/f6ea5712d578b02685120b484a2f3bbc3b78055a/src/editors/ace/modes/source.ts#L116-L118
The cases for '|' and '&' capture other parts of the string in the attempt to exclude '||' and '&&'.
This PR has a fix for this, using lookahead and lookbehind assertions.
This PR also includes a few lines in the test for this file to cover the case of '|'.

### Type of change

- [ ] Documentation update

### How to test

I expanded the existing tests for this file to cover this change.
https://github.com/notmeegoreng/js-slang/blob/patch-1/src/__tests__/mode.ts#L82

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/__tests__/mode.ts
+++ b/src/__tests__/mode.ts
@@ -79,7 +79,7 @@ test('constants are not correctly loaded', () => {
 })
 
 test('operator syntax type error', () => {
-  const code = 'const num = 3; \nnum++; \nnum--; \nnum += 1;'
+  const code = 'const num = 3; \nnum++; \nnum--; \nnum += 1; \n5 + num |2;'
 
   setSession(Chapter.SOURCE_1, defaultVariant, defaultExternal, code)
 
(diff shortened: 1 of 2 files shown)


s-kybound (approved):
Looks good to me, tested the regex online. @martin-henz Prof, do you have any comments as well?

Issue #701, opened 2020-07-31
Source 3 Non-Det: Improve documentation for amb operator


anubh-v:
To aid individuals coming from Source 3, the docs should mention that `amb` is not intended to behave like ordinary Source functions.

Specifically, when we evaluate `amb(alpha(), bravo())`, the call is not evaluated "inside out".
The expression `bravo()` will only be evaluated upon backtracking.

Moreover, `amb` cannot be passed around as a value.

leeyi45:
Resolved by #1725

PR #1725, opened 2024-10-09
Remove Non-Det Interpreter

leeyi45:
Instead of doing the big jump and removing everything at once, I think it might be better if I broke down #1707 into several different PRs.

Will resolve:
- #1617 
- #1695 
- #701


RichDom2185:
Let me know when it's ready for review, thanks!

leeyi45:
> Let me know when it's ready for review, thanks!

It is ready for review I guess, if we don't mind the coverage drop

eslint.config.mjs
```diff
@@ -0,0 +1,75 @@
+import tseslint from 'typescript-eslint'
+import globals from 'globals'
+import importPlugin from 'eslint-plugin-import'
+
+/**
+ * @type {import('eslint').Linter.FlatConfig[]}
+ */
+export default [
```
RichDom2185, line 8: I think we can use the `tseslint.config` helper now that things are more stable, and also add a `@ts-check`

package.json
```diff
@@ -72,14 +72,12 @@
     "@types/lodash": "^4.14.202",
     "@types/node": "^20.0.0",
     "@types/offscreencanvas": "^2019.7.0",
-    "@typescript-eslint/eslint-plugin": "^7.3.1",
-    "@typescript-eslint/parser": "^7.3.1",
     "ace-builds": "^1.4.12",
     "coveralls": "^3.1.0",
     "escodegen": "^2.0.0",
     "eslint": "^8.57.0",
```
RichDom2185, line 78: Are we not bumping to v9 together with flat config?

RichDom2185: Though since technically flat config is supported since v8.something, it's not necessary

RichDom2185 (changes requested):
Just some minor nits :)

package.json
leeyi45: For that exact reason I left it alone

eslint.config.mjs
leeyi45: The issue with adding the `@ts-check` is that some of the plugins don't provide their own declaration files (and neither does ESLint), and I'd rather not add more packages

package.json
RichDom2185: Sounds good!

eslint.config.mjs
RichDom2185: It's a matter of preference, but personally I prefer having a `@ts-check` on by default and a `@ts-expect error incompatible/incorrect type definition` to silence the instances we know works fine, just to prevent accidental mistakes

leeyi45: Made the changes as requested

s-kybound:
Hi, please run the command `yarn upgrade string-width strip-ansi` to update these two packages before trying this again!

leeyi45:
> Hi, please run the command `yarn upgrade string-width strip-ansi` to update these two packages before trying this again!
i've done as requested, but I've now got a bunch of failing tests that I'm not sure what to do about

s-kybound (approved):
LGTM

PR #1731, opened 2025-01-22
bumping version; updating scm-slang

martin-henz:
Version bump requested by @s-kybound ; needed for newest changes in frontend


RichDom2185 (approved):

RichDom2185 (changes requested):
Prof, can you bump once more? So the frontend doesn't break due to the GL change

Issue #1736, opened 2025-02-18
Source Typed Variant CallExpressions not working with Custom Types


tohlh:
Consider the following program:

```typescript

type Rune = "Rune";
const heart : Rune = "Rune";

function beside(rune1: Rune, rune2: Rune) {
    return "Rune";
}

beside(heart, heart);

beside("heart", heart); // Shows this error: Line 12: Type 'Rune' is not assignable to type 'string'.

beside(heart, "heart"); // Expects a type error, but nothing was thrown

beside("heart", "heart"); // Expects a type error, but nothing was thrown

```

This is not the desired behaviour of type checker for Source Typed variant. Currently, this has only been observed when call expressions that contain parameters with custom types.

tohlh:
My current observation is the problem is caused by this loop: https://github.com/source-academy/js-slang/blob/master/src/typeChecker/typeErrorChecker.ts#L417C7-L433C8

Once there is a string (in this case, "heart") assigned to one of the parameters, the subsequent parameters' expected types are replaced with their actual types, i.e. string, resulting in the behavior described above.

PR #1739, opened 2025-02-24
Typed modules

tohlh:
This is an implementation of Typed Modules based on this [proof-of-concept](https://github.com/source-academy/js-slang/pull/1467) and this [proposal](https://github.com/source-academy/modules/issues/326) for the FYP titled "Improving Types in the Source Academy".

Previously, module members are mapped to the `any` type for all variables, function parameters and function return values. With the implementation of `type_map` in the modules repository: https://github.com/source-academy/modules/pull/331, changes are made on the `js-slang` repository to load type maps into the `context`, before type checking could be performed properly on module members.

This PR also fixes this newly discovered bug https://github.com/source-academy/js-slang/issues/1736 where `CallExpressions` are not being properly type checked.


src/typeChecker/__tests__/source4TypedModules.test.ts
```diff
@@ -0,0 +1,287 @@
...
+import { Chapter, Variant } from '../../types'
+import { parse } from '../../parser/parser'
+import { parseError } from '../../index'
+
+let context = mockContext(Chapter.SOURCE_4, Variant.TYPED)
+
+beforeEach(() => {
+  context = mockContext(Chapter.SOURCE_4, Variant.TYPED)
+  context.nativeStorage.loadedModules = {
+    module: {}
```
RichDom2185, line 11: Nit: let's name it `exampleModule` or some other thing so that first time developers don't get confused

src/typeChecker/__tests__/source4TypedModules.test.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+  }
+})
+
+describe('Typed module tests', () => {
+  /* VALID CASES */
+
+  it('should allow correct assignment of a string variable', () => {
+    const code = `
+      import { x } from 'module';
```
RichDom2185, line 39: Because although you can infer it this way that `module` is the name of the module, it's not immediately obvious unless you look through the code and are familiar with both the proof-of-concept and proposal.

src/typeChecker/typeErrorChecker.ts
```diff
@@ -10,6 +10,7 @@ import {
   InvalidIndexTypeError,
   InvalidNumberOfArgumentsTypeError,
   InvalidNumberOfTypeArgumentsForGenericTypeError,
+  // NameNotFoundInModuleError,
```
RichDom2185, line 13: Why is this commented?

src/createContext.ts
```diff
@@ -153,7 +153,8 @@ const createNativeStorage = (): NativeStorage => ({
   gpu: new Map(Object.entries(gpu_lib)),
   maxExecTime: JSSLANG_PROPERTIES.maxExecTime,
   evaller: null,
-  loadedModules: {}
+  loadedModules: {},
+  loadedModulesTypes: {}
```
RichDom2185, line 157: idk if it's my English but I think it would be `loadedModuleTypes` instead of `loadedModulesTypes`? What do you think?

RichDom2185 (approved):
LGTM, thanks a lot!

Issue #1650, opened 2024-04-08
CSE: spreading of arrays is not supported, propose new `spread` instruction


CZX123:
The below code runs correctly in the REPL tab, but fails in the CSE machine tab:
```js
const f = (...x) => x;
f(...[1, 2, 3]);
```
This is because the `SpreadElement` node is not handled inside the CSE machine interpreter.

### Proposal
A `SpreadElement` node, when evaluated, should result in two new items inside the control:
1. A `spread` instruction
2. The expression on the right of the ellipsis.

Once the expression on the right is fully evaluated, there should be a runtime check that the result is indeed an array, and then the `spread` instruction should unpack the array in the stash, by removing it and adding the individual elements back into the stash.

This new `spread` instruction could also potentially allow spreading of array elements inside arrays, e.g. `[...arr]`, instead of just being limited to the arguments of function calls.

### Update
It appears that the `call 1` instruction above the `SpreadElement` node would also be incorrect, and it should be `call 3` instead. Unfortunately, the correct number of arguments can only be known after the array expression is evaluated, which is after the `call` instruction has already been pushed.

We also have to take note of more complex situations like below, where there can be multiple spread elements inside a function call:
```js
const f = (...x) => x;
f(0, ...[1, 2], ...[3, 4]);
```

The simplest solution would be to increase the value in the `call` instruction by the array length - 1 every time a `spread` instruction is run, and keep the limitation where array spread can only be used inside function calls, so it is guaranteed that there will be a `call` instruction somewhere above the `spread` instruction.

martin-henz:
Yes, this looks like the correct specification for this feature.

PR #1740, opened 2025-02-25
Implements new spread instruction

alsonleej:
### Type of change

Deals with  #1650

### How to test

const f = (...x) => x;

// one spread
// f(...[1, 2, 3]);

// empty array spread
// f(...[]);

// literals, empty array, multiple spreads
// f(0, ...[1, 2], ...[3, 4], ...[], ...[6], 7);

// array reference
// const rr = [1,2,3];
// f(...rr, 3);

// function returning array
// const k = () => [1, 2, 3];
// const g = (...x) => x;
// g(...k()); // Expected: [1, 2, 3]

### Checklist

- [ ] I have tested this code

### DIFF
--- a/src/cse-machine/interpreter.ts
+++ b/src/cse-machine/interpreter.ts
@@ -584,8 +584,6 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
     // Refer to Source §3 specifications https://docs.sourceacademy.org/source_3.pdf
     if (init.type === 'VariableDeclaration' && init.kind === 'let') {
       const id = init.declarations[0].id as es.Identifier
-      // const valueExpression = init.declarations[0].init!
-
       control.push(
         ast.blockStatement(
           [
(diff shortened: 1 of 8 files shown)


martin-henz:
TO-DO: run time check if RHS of spread operator is indeed an array

What happens if the RHS of spread is not an array? Can you try:

math_max(... 1);

src/cse-machine/interpreter.ts
```diff
@@ -1310,5 +1319,31 @@ const cmdEvaluators: { [type: string]: CmdEvaluator } = {
...
+    // spread array
+    for (let i = 0; i < array.length; i++) {
+      stash.push(array[i])
+    }
+
+    // update call instr above
+    const cont = control.getStack()
+    const size = control.size()
+    for (let i = size - 1; i >= 0;  i--) {
+      // guaranteed at least one call instr above
```
martin-henz, line 1342: ...because spread in arrays is currently not allowed in Source.

martin-henz (approved):
Looks good!

PR #1741, opened 2025-02-25
Bump prettier, ace-builds, and misc clean up

RichDom2185:
### Description

To take advantage of prebuilds. Also major bump for Prettier to align with other repositories, even if their defaults differ from Source coding standard.

### Checklist

- [x] I have tested this code


src/__tests__/mode.ts
```diff
@@ -14,7 +14,7 @@ const defaultVariant: Variant = Variant.DEFAULT
 const defaultExternal: string = 'NONE'
 
 // define session
-const session = ace.createEditSession('', DefaultMode)
+const session = ace.createEditSession('', new DefaultMode.Mode())
```
tsammeow, line 17: what does this change do, doesn't seem to be building?

tsammeow (commented):
lgtm except that one comment

src/__tests__/mode.ts
RichDom2185: There's a TS type error otherwise, looking into it

tsammeow: i'd be happy to approve the other two upgrades if you want to get those through first

tsammeow (approved):
lgtm thanks!

PR #1752, opened 2025-03-18
Fix for JSON imports not working

sayomaki:
### Description

As per the original comment in the commit 80341a9, the 3 year long issue still holds right now as described by the comment itself. It appears that the compiled version (using `yarn build`) will append the `default` key to the imported JSON files, which does not exist and hence giving an error when used by other project (i.e. frontend). However, it does work correctly in testing/jest which appears to use `ts-jest` instead to transpile the typescript files.

This PR reverts the commit added here: https://github.com/source-academy/js-slang/pull/1741/commits/bca24234e63eccb3413f8185d154981cf95e74de

Some other possible fixes for this that can be considered (but more testing and discussion required):
1. Not using a JSON file (could be in a ts file instead, but since it is autogenerated there would be more steps to properly transform this)
2. Standardising the build procedure to be same as Jest (would be complicated to switch to a different packaging tool entirely)
3. Making jest use `tsc` instead of `ts-jest` (mentioned briefly on the jest docs, but this may affect coverage reporting/results)

### Type of change

- [ ] Documentation update

### Checklist

- [x] I have tested this code

### DIFF
--- a/src/__tests__/mode.ts
+++ b/src/__tests__/mode.ts
@@ -22,9 +22,9 @@ const CATEGORY = {
   types: /\bstorage.type\b/,
   forbidden: /\bvariable.language\b/,
   keywords: /\bkeyword\b/,
+  consts: /\bbuiltinconsts\b/,
   number: /\bconstant.numeric\b/,
-  bool: /\bconstant.language.boolean\b/,
-  identifier: /\bidentifier\b/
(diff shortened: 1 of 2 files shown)


martin-henz (approved):
Let's merge this. Seems like an urgent fix.

PR #1799, opened 2025-07-21
Migrate to TypeScript 5.9

RichDom2185:
### Description

Supersedes #1666.

### Type of change

- [ ] Documentation update

### Checklist

- [ ] I have tested this code


leeyi45:
> > Is the stepper officially deprecated (and replaced with the tracer)?
> 
> 
> 
> Is it not just a naming thing? I'm unaware of the developments; this PR tries to be as small as possible and not change any behavior (just trying to update TS)
> 
> 
> 
> I do have a subsequent PR that enables `verbatimModuleSyntax` and migrates the package to ESM, but I don't want to include it here because of scope creep

I ask because I briefly toyed with a similar migration and got tired of fixing the stepper but then I checked and nothing refers to the stepper outside of the stepper folder.

RichDom2185:
> > > Is the stepper officially deprecated (and replaced with the tracer)?
> > 
> > 
> > Is it not just a naming thing? I'm unaware of the developments; this PR tries to be as small as possible and not change any behavior (just trying to update TS)
> > I do have a subsequent PR that enables `verbatimModuleSyntax` and migrates the package to ESM, but I don't want to include it here because of scope creep
> 
> I ask because I briefly toyed with a similar migration and got tired of fixing the stepper but then I checked and nothing refers to the stepper outside of the stepper folder.

Prof @martin-henz is the stepper gone?

@leeyi45 Either way, I think we can just remove it (if we have to) in a future PR.

src/transpiler/transpiler.ts
```diff
@@ -260,8 +260,8 @@ function transformSomeExpressionsToCheckIfBoolean(program: es.Program, globalIds
     const { line, column } = (node.loc ?? UNKNOWN_LOCATION).start
     const source = node.loc?.source ?? null
     const test = node.type === 'LogicalExpression' ? 'left' : 'test'
-    node[test] = create.callExpression(globalIds.boolOrErr, [
-      node[test],
+    ;(node as any)[test] = create.callExpression(globalIds.boolOrErr, [
```
leeyi45, line 263: I'd rather have this written as 
```
const testNode = node.type === 'LogicalExpression' ? node.left : node.test
const callExpr = create.callExpression(... whatever)
if (node.type === 'LogicalExpression') {
  node.left = callExpr
} else {
  node.test = callExpr
}
```

Then we can avoid the type assertion

src/repl/utils.ts
```diff
@@ -23,7 +22,7 @@ export function chapterParser(str: string): Chapter {
   }
 
   if (foundChapter in Chapter) {
```
leeyi45, line 24: For these, I wrote a typeguard:
```ts
export function isKeyOf<T extends Record<string | number | symbol, unknown>>(obj: T, p: string | number | symbol): p is keyof T {
  return p in obj;
}
```

Then you can just use it like
```ts
if (isKeyOf(str, Chapter)) {
  return Chapter[str]
}
```

src/stepper/converter.ts
```diff
@@ -62,7 +62,7 @@ export function nodeToValueWithContext(node: substituterNodes, context: Context)
   return node.type === 'Literal'
     ? node.value
     : util.isBuiltinFunction(node)
-      ? builtin[(node as es.Identifier).name]
+      ? builtin[(node as es.Identifier).name as keyof typeof builtin]
```
leeyi45, line 65: If we change `isBuiltInFunction` to a proper type guard we won't need the type assertion:

```ts
type BuiltinIdentifier = Omit<es.Identifier, 'name'> & {
  name: keyof typeof builtin
}

export function isBuiltinFunction(node: SubstituterNodes): node is BuiltinIdentifier {
  if (!isIdentifier(node) || !isKeyType(builtin, node.name)) return false
  // ...the other checks
}
```

leeyi45 (commented):
In general LGTM, not sure if we want to pursue the path of fewer type assertions and more usage of typeguards and stuff

s-kybound:
> Prof @martin-henz is the stepper gone?

The tracer is the new stepper, and should completely replace it. I'm double checking with the ~stepperv2~ tracer team, and if all is good I intend to remove the old stepper code completely.

src/stdlib/pylib.ts
```diff
@@ -28,7 +28,8 @@ export function __py_adder(x: Value, y: Value) {
     return x + Number(y)
   }
   if (__is_numeric(x) && __is_numeric(y)) {
-    return x + y
+    // TODO: Shouldn't this be `Number(x) + Number(y)`?
```
s-kybound, line 31: @Fidget-Spinner ^ thoughts? :sweat_smile:

src/transpiler/transpiler.ts
s-kybound: ^^ we should do this, we should aim to reduce the amount of total type assertions/typeguards

src/repl/utils.ts
s-kybound: ^ stylistically I agree and would rather use a typeguard to avoid explicit type assertions

s-kybound (changes requested):
LGTM with a few nits. I am a bit concerned with the amount of additional type gymnastics we have to do, but that's more to do with the codebase quality than this PR. Additionally, I agree with @leeyi45 and would argue for typeguards > over type assertions, but this is more of a stylictic choice.

Should we draft out a style guide for Source Academy (and maybe heavily discourage `any`?)

src/infiniteLoops/state.ts
```diff
@@ -138,7 +138,9 @@ export class State {
     }
     const transitions = this.mixedStack[this.stackPointer].transitions
     for (let i = 0; i < transitions.length; i++) {
-      const transition = transitions[i]
+      // TODO: Something seems to be very wrong with the type definitions here
+      // Shall we just remove the infinite loop detector as per #1516?
```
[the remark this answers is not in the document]
RichDom2185: Normally, I try to fix type issues as I go (even if it means weird typecasts etc.) but this is beyond me; either it's some super old TS logic that predates my knowledge of the "modern" 4.x and 5.x versions or it's just (ab)use of TS that was somehow never caught.

Without looking at the code and figuring out how this all works I had to resort to just putting an `any` there.

I suspect the original type definitions may have not been entirely correct, but since TS implicitly gives an `any` for these kinds of invalid accesses (and they are accessed in a way consistent and expected of the actual contract/behaviour) then it all just magically seems to work.

src/transpiler/transpiler.ts
RichDom2185: Agreed, originally I didn't want this PR to introduce behavioural changes (and leave these cleanup tasks to the future instead) but on second thought it makes sense to do it as part of this PR (since the goal of migrating is for type safety anyway)

Although, having to have these awkward semicolon placements due to the fact the codebase doesn't use semicolons by default bugs me. It may work for plain JS but TS is not the place for no semicolons imo.

What do you think of adding semicolons into the codebase?

leeyi45:
> LGTM with a few nits. I am a bit concerned with the amount of additional type gymnastics we have to do, but that's more to do with the codebase quality than this PR. Additionally, I agree with @leeyi45 and would argue for typeguards > over raw type assertions, but this is more of a stylictic choice.
> 
> 
> 
> Should we draft out a style guide for Source Academy (and maybe recommend to write code in ways that avoid the need for playing with the typechecker?)

I think a big part of the reason is that a lot of Typescript tricks wouldn't be immediately obvious to someone that hasn't been working with the language for that long so things like Typeguards won't come to mind first. 

Like I was staring at the code for the cse machine and there's a ton of type assertions that don't need to be there

src/infiniteLoops/state.ts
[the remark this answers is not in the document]
martin-henz: I'm ok with removing the feature for now. Re-implementing it will lead to a cleaner implementation. 

The project is well-documented in a report and can be revived from there.

I don't think the system will be missed very much. It carries a bit of infrastructure with a server side, sentry support etc that was needed for carrying out the project.

So yea, please go ahead and remove it.

src/transpiler/transpiler.ts
leeyi45: Are we expanding the scope of the PR to include general type safety? I can include some changes I was working on for the cse machine

leeyi45:
I just noticed something really weird:

In the modules preprocessor `parseProgramsAndConstructImportGraph` is called twice?

PR #1811, opened 2025-08-16
[Stepper] Remove the obsolete stepper

s-kybound:
### Description

The tracer has completely superceded the original stepper. This PR removes the dead code from `js-slang`.

### Checklist

- [X] I have tested this code: all tests pass, no errors generated, `js-slang` as an executable works


RichDom2185 (approved):
LGTM, thanks!

PR #1893, opened 2026-02-20
Error Handling and Stringify Changes

leeyi45:
# Changes to Errors

To better unify the errors thrown by modules with those thrown by `js-slang`, I want to make the errors in `js-slang` available for modules to use. Part of that involves actually having Source errors inherit from `Error`.

There are a lot of legacy error classes that don't seem to be used for anything anymore (like those under `typeErrors.ts`), but I'm not sure what is used where (because the documentation is woefully inadequate)

This led me down a bit of a rabbit hole where I found that the new tracer doesn't actually make use of the existing functions for verifying types (like `rttc.checkBinaryExpression`), so I had to go through and fix those errors because otherwise they weren't properly following the Source error format.

I'm not sure if this will also need changes to how errors are displayed in the frontend.

When code from the prelude throws an error, the `prelude` file name is automatically attached to it to indicate to the user that the error was thrown in prelude code, thereby demystifying weird errors thrown at line 216 when the user's code clearly doesn't have 200 lines of code.

Where possible, `InternalRuntimeError` and `GeneralRuntimeError` should be favoured over throwing the raw `Error` type.

# Making better use of Typescript

I found a few instances of (especially when having a processing function for each Node type) code that looked like this:

```ts
const processors = new Map<Node, (node: Node) => string>([
  ['ImportDeclaration', node => 'An import declaration'],
  ['FunctionDeclaration', node => 'A function declaration'],
  // other node processors...
  ['ImportDeclaration', node => 'A declaration that imports'],
])
```

The code inadvertently defines two kinds of processors for the same type of node. I've replaced such instances with a much more straightforward, type safe, object based approach:

```ts
// NodeTypeToNode helper converts a string like 'ImportDeclaration' to the actual node type
const processors: { [K in Node['type']?: (node: NodeTypeToNode<K>) => string) } = {
  ImportDeclaration: node => 'An import declaration' // Type of `node` parameter is automatically inferred!
}
```

This avoids a lot of extraneous type annotations and helps increased type safety: you can't define multiple processors for the same node type and if you forget to define a processor for a certain node type Typescript will let you know.

# Stringify Changes

Previously, `toReplString` wasn't being prioritized over `toString` in some cases. With this change, so long as the `toReplString` property is defined and is a function the return value will be used. This should make it such that most objects from modules should have consistent and working REPL representations.

# Changes for Modules

I've added a way to change how Source modules are loaded (similar to how `FileGetter`s are used to control how local modules are loaded). In the future this should enable running `js-slang` from the REPL without having to spin up an instance of the modules server.

I've also begun some work into introducing the necessary functionality for modules to begin to control their execution environment. Modules will error out if they are loaded with too low a version of Source.

# Adding functions to `stdlib`

A lot of the functions available in the prelude weren't available in Javascript. By adding these functions modules can work with the list and stream abstractions more directly.

# Minor Change to `createContext`

You don't have to provide all `CustomBuiltins` at once. You can just provide the ones you want to override, and the rest will just use the default implementations provided by `js-slang`.


s-kybound (commented):
Overall, great improvement to the codebase, much more elegant. I've tried to check the changes to see if the changes have been propagated properly into the different sections of the PR.

src/stdlib/index.ts
```diff
@@ -7,7 +7,7 @@ import * as parser from './parser'
 import * as stream from './stream'
 
 export const chapter_1 = {
-  get_time: misc.error_message,
```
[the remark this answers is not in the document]
leeyi45: My recommendation is that we should extract `stdlib` into its own package. From there we can use things like API generator which generate API files that you then commit so that you get alerted an exposed API changes.

I guess it could still be done from within `js-slang`.

src/utils/stringify.ts
```diff
@@ -431,12 +433,16 @@ export function valueToStringDag(value: Value): StringDag {
     } else if (typeof v === 'string') {
       const str = JSON.stringify(v);
       return [{ type: 'terminal', str, length: str.length }, false];
+    } else if (typeof v.toReplString === 'function') {
+      // callIfFuncAndRight args is necessary because if we implement toReplString as a function
+      // in Source, it gets wrapped by the transpiler
+      // this allows object literals to implement toReplString
+      const reprValue = callIfFuncAndRightArgs(v.toReplString.bind(v), -1, -1, null, undefined);
```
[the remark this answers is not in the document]
leeyi45: Not an issue. If stringify is called in Source code, then it will be called through callIfFuncAndRightArgs.

src/runner/sourceRunner.ts
```diff
@@ -40,15 +39,11 @@ const runners = {
...
-    // a bunch of global identifiers to it. Once that happens, the infinite loop detection instrumentation
-    // ends up generating code that has syntax errors. As such, we need to make a deep copy here to preserve
-    // the original AST for future use, such as with the infinite loop detector.
-    const transpiledProgram = _.cloneDeep(program);
-    let transpiled;
     let sourceMapJson: RawSourceMap | undefined;
     try {
-      ({ transpiled, sourceMapJson } = transpile(transpiledProgram, context));
+      let transpiled: string;
+      ({ transpiled, sourceMapJson } = transpile(program, context, options.isPrelude));
```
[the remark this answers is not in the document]
leeyi45: This was necessary because the AST was being reused. However, the AST is no longer being reused. Regardless, the cloning of the AST should be located in the transpiler instead of the runner.

src/parser/source/rules/index.ts
```diff
@@ -26,7 +25,7 @@ import noUpdateAssignment from './noUpdateAssignment';
 import noVar from './noVar';
 import singleVariableDeclaration from './singleVariableDeclaration';
 
-const rules: Rule<Node>[] = [
+const rules: Rule<any>[] = [
```
s-kybound: Why was this changed from Node -> any?

src/stdlib/misc.ts
```diff
@@ -84,27 +88,29 @@ export function array_length(xs: Value[]) {
...
-    return parseInt(str, radix);
-  } else {
-    throw new Error(
-      'parse_int expects two arguments a string s, and a positive integer i between 2 and 36, inclusive.',
-    );
+  if (typeof str !== 'string') {
+    throw new InvalidParameterTypeError('string', str, parse_int.name, 'str');
   }
+
+  assertNumberWithinRange(radix, parse_int.name, 2, 36, true, 'radix');
```
s-kybound, line 95: nice

src/transpiler/transpiler.ts
```diff
@@ -1,7 +1,12 @@
+/**
+ * This whole transpiler includes many many many many hacks to get stuff working.
+ * Order in which certain functions are called matter as well.
+ * There should be an explanation on it coming up soon.
```
s-kybound, line 4: we should probably mark this as a TODO, or raise an issue on the code quality of the transpiler

s-kybound (approved):
I'm quite happy with the PR as is - it really improves the code quality in js-slang and heavily reduces code duplication. Considering that the PR has been held back for a while now, since the tests pass, I think the best call is to push it, and respond to any issues that come up.

Minor queries and requests in comments below.

@martin-henz @RichDom2185 bumping just in case you want to check the PR again before final merging.

src/parser/source/rules/index.ts
leeyi45: This is the definition of the `Rule` interface:

```ts
export interface Rule<T extends Node> {
  name: string;
  disableFromChapter?: Chapter;
  disableForVariants?: Variant[];
  checkers: {
    [K in T['type']]: (node: NodeTypeToNode<K>, ancestors: Node[]) => RuleError<Node>[];
  };
}
```

Every checker in the `checkers` property must be present and assigned (no question mark). The reason for this is so that Typescript will produce an error if you don't define a checker for all the node types that a given rule is supposed to cover.

An object assignable to  `Rule<Node>` must have a checker function for every possible node type, which of course most rules don't. So that is why it has to be `Rule<any>` here.

src/transpiler/transpiler.ts
leeyi45: My hope is that in the future we can have a solution similar to the modules developer documentation.

src/parser/source/rules/index.ts
leeyi45: Either ways I changed the interface, so this isn't an issue anymore

PR #1939, opened 2026-04-06
CSE machine: support for StreamViz drawing arrows

of89p:
### Description

The map streamLineage (Found in context object, types.ts) that maps the ID of nullary functions to the pairs it returns. 

Everytime a CALL0 is called in the control, we push a tuple (nullaryFnID, controlStackHeight) to the pendingStreamFnStack[]. Within generateCSEMachineStateStream, we check the whether the height of the control stack has reached the height of i-1, where i is the controlStackHeight when we pendingStreamFnStack.peek(). If so, we pop pendingStreamFnStack.pop() and we map (using streamLineage) the nullaryFnID to the pair that is at the top of the stash.

We store breakpoints in streamPointSteps[]. This array is calculated within envChangingStreams. The criteria for a breakpoint is exactly the same as updating streamLineage. 

See: https://github.com/source-academy/frontend/issues/3727#issuecomment-4188491044
and https://github.com/source-academy/frontend/pull/3747

This PR is blocking: https://github.com/source-academy/frontend/pull/3747

### Type of change

- [ ] Documentation update

### Checklist

- [ ] I have tested this code.
  - [ ] Tests pass with `yarn test`
  - [ ] `js-slang` executable is verified to work
  - [ ] Did I have to reduce the strength of some tests?
    - If so, why...


martin-henz:
/gemini review
Can you check that all review comments are properly handled?

martin-henz:
/gemini review
can you describe how the trace that the CSE machine of js-slang passes to the frontend represents the links between nullary functions and the pairs they return?

martin-henz (dismissed):
I think we should add regression tests for this change because it affects the core of the CSE machine: function calls

martin-henz (approved):
Looks good!

PR #1942, opened 2026-04-10
Removing py-slang submodule from js-slang

martin-henz:
Background: Source Academy is moving to conductor framework. Python evaluators will be loaded dynamically into the frontend, not hardcoded via js-slang.

This PR removes all Python-related functionality from js-slang.


.prettierignore
```diff
@@ -1,3 +1,2 @@
 /src/alt-langs/
```
RichDom2185: Please also remove alt-langs (missed in the Scheme PR)

eslint.config.mjs
```diff
@@ -8,7 +8,7 @@ import * as importPlugin from 'eslint-plugin-import'
 export default tseslint.config(
   {
     // global ignores
-    ignores: ['dist', 'src/alt-langs', 'src/py-slang', 'src/__tests__/sicp', '**/*.snap']
+    ignores: ['dist', 'src/alt-langs', 'src/__tests__/sicp', '**/*.snap']
```
RichDom2185: Same here

vitest.config.ts
```diff
@@ -14,7 +14,6 @@ export default defineConfig({
         "dist",
         '**/__mocks__/**',
         "src/alt-langs/scheme/scm-slang",
```
RichDom2185: Same here

RichDom2185 (changes requested):
Please also update the README to remove instructions pertaining to git submodules since we don't use them anymore

RichDom2185 (approved):
LGTM, thanks!

PR #1961, opened 2026-04-24
Add auto-publish workflow for npm

martin-henz:
## Summary
- Adds `.github/workflows/publish.yml` that auto-publishes `js-slang` to npm when `package.json` version changes on `master`.
- On every push to `master`, it reads the version from `package.json` and asks npm whether `js-slang@<version>` already exists. If yes, it exits cleanly; if no, it builds and publishes.
- After a successful publish, it pushes a `v<version>` tag so git tags track released versions.
- Fails fast with a clear message if `NPM_TOKEN` is not reachable from this repo (useful the first time we run it, since the token lives at the org level and needs the repo in its allowed-repos list).

## How releasing works after this
1. Bump `package.json` version in a PR (same flow as the existing `bumping version` commits).
2. Merge to `master`.
3. The workflow publishes to npm and pushes the matching tag. No manual `npm publish` needed.

If a push to `master` does not bump the version, the workflow is a no-op — duplicate version detection is based on what is already on npm, so re-running is safe.

## Notes
- `NODE_AUTH_TOKEN` is wired via `actions/setup-node`'s `registry-url` — the standard pattern for npm publishes from Actions.
- `permissions: contents: write` is scoped to this workflow so it can push the release tag.
- No changes to the existing `nodejs.yml` CI; that still runs on every PR and push and continues to gate merges.

## Test plan
- [ ] First push to `master` after merge: confirm the `Verify NPM_TOKEN is available` step passes (i.e. the org secret is scoped to this repo). If it fails, add `js-slang` to the org secret's allowed repos.
- [ ] Next version bump (1.0.93): confirm the workflow publishes to npm and that a `v1.0.93` tag appears on the repo.
- [ ] Subsequent unrelated push to `master` with no version change: confirm the workflow runs and exits cleanly without republishing.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,94 @@
+name: Publish to npm
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Verify NPM_TOKEN is available
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          if [ -z "$NPM_TOKEN" ]; then
+            echo "NPM_TOKEN secret is not reachable from this repo."
+            echo "If it is set as an organization secret, confirm this repo is in its allowed-repos list."
+            exit 1
+          fi
+          echo "NPM_TOKEN is present."
+
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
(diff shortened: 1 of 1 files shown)


RichDom2185 (changes requested):
Outdated way to publish

RichDom2185:
As mentioned in the past, we should not be using NPM token due to various security reasons. Not to mention the PR description seems to be incorrect.

RichDom2185:
This should also be gated, not freely on every push to master

PR #1977, opened 2026-05-28
fix(stepper): various stepper fixes

### LINKED ISSUE #1805: Stepper: Stepper Produces Long, Hard-to-Read Trace for Higher-Order Recursion Function
Consider this following snippet:
```js
function h(f, x) {
    function h(g, x) {
        return x <= 1 ? 1 : 3 * g(f, x - 1);
    }
    return x <= 1 ? 1 : 2 * f(h, x - 1);
}
h(h, 5);
```

Image

The reasonable explanation for this step should be `1 <= 1 ? 1 : 2 * h(h, 1 - 1)) returned`.

Shrey5132:
Resolves #1805

### Description
This PR fixes a bug where the stepper produced extremely long, deeply nested, and hard-to-read substitution traces/explanations for recursive or higher-order functions.

**Root Cause**:
Recursive and named functions are converted into `StepperArrowFunctionExpression` nodes with a custom `name` property. While the visualizer's top pane React renderer cleanly prints named lambdas by their `name` (with a hover tooltip), the bottom pane explanation strings are generated inside `js-slang` using standard `astring.generate()`. The standard generator has no knowledge of this custom `name` field, so it stringified the entire arrow function body recursively.

**Fix**:
Introduced a custom code generator wrapper around `astring` inside `utils.ts` that:
1. Intercepts `ArrowFunctionExpression` nodes and stringifies them to their assigned `name` if present.
2. Intercepts `CallExpression` nodes to prevent wrapping named arrow function calls in unnecessary parenthesis (e.g. producing `h(args)` instead of `(h)(args)`).

This makes the generated trace text perfectly consistent with the top visualizer pane formatting.

### Type of change
- [ ] Documentation update

### How to test
Added a dedicated unit test in `stepper_full.test.ts` representing the nested recursive function from the issue report:
```javascript
function h(f, x) {
    function h(g, x) {
        return x <= 1 ? 1 : 3 * g(f, x - 1);
    }
    return x <= 1 ? 1 : 2 * f(h, x - 1);
}
h(h, 5);

### CODE REVIEW

**Review by Akshay-2007-1 — CHANGES_REQUESTED** [review]
File: src/stepper/generator.ts Line: 101

---
Zero-param named block-body functions show a generic explanation inconsistent with the code pane

The CallExpression explainer hard-codes the string '() => {...} runs' for any zero-parameter block-body function call, regardless of whether the function has a name:

```typescript
} else if (func.body.type === 'BlockStatement') {
  if (func.params.length === 0) {
    return '() => {...}' + ' runs';   // ignores func.name entirely
  }
```
This PR's fix makes the code pane use func.name when rendering named functions — so after a function foo() { ... } declaration, the code pane correctly shows foo() at the call site. But the explanation pane at the same step still says () => {...} runs, creating a visible contradiction: the two panes describe different things for the same reduction step.

Image

---

Verified: running the stepper on a program with a named zero-param function (e.g., function a() { return () => 2; }) shows a() in the code pane and () => {...} runs in the explanation at the moment of the call.

src/stepper/utils.ts
```diff
@@ -34,3 +35,36 @@ export function assignMuTerms(
...
+        }
+      }, state);
+    } else {
+      GENERATOR.CallExpression!(node, state);
+    }
+  }
+};
+
+export function generate(node: any, options?: any) {
+  return astringGenerate(node, {
```
Akshay-2007-1, line 69: The spread order { generator: customGenerator, ...options } lets a caller's options.generator silently override customGenerator. Since generate is exported, this might cause a problem later. 

Flip the order to make the custom generator non-overridable. Just a thought!

Akshay-2007-1 (commented):
Just another comment as follows

PR #1984, opened 2026-05-31
fix(stepper): various stepper fixes

Shrey5132:
This PR contains multiple stepper fixes.
Check the commit descriptions under the "Commits" tab for more details for each commit.

NOTE: The "Files changed" tab does not show differences between individual commits, only the overall differences between the source-academy:master branch and my fork Shrey5132:stepper-fixes. To view differences from commit to commit, go to my fork Shrey5132:stepper-fixes and view the commits there.


Akshay-2007-1 (changes requested):
In BinaryExpression.ts file, there is this function thats called repeatedly
```ts
 public override isOneStepPossible(redex: RedexInfo): boolean {
    return (
      this.left.isOneStepPossible(redex) ||
      this.right.isOneStepPossible(redex) ||
      this.isContractible(redex)
    );
  }
```

the same this.left.isOneStepPossible() is called again inside the isContractible() function. Consider maybe caching the result to avoid double the work in complicated tasks.

```ts
public override isContractible(redex: RedexInfo): boolean {
    const leftFullyReduced = !this.left.isOneStepPossible(redex);
    const rightFullyReduced = !this.right.isOneStepPossible(redex);
.............
```

I am talking purely from a functional standpoint. If in practice, this code is used on such trivial and small tasks that it doesnt matter, ignore this comment!

The rest is all clean as far as I see it

Akshay-2007-1:
In BinaryExpression.ts:40, UnaryExpression.ts:65, both isContractible methods now call checkBinaryExpression/checkUnaryExpression which can throw RuntimeTypeError. This makes isContractible (and transitively isOneStepPossible) a method with side effects callers now need to know it can throw.

This works because steppers.ts:31-32 wraps the call in a try-catch:
```ts
  try {
    if (node.isOneStepPossible(redex)) {  // can now throw
      const newNode = node.oneStep(redex);
      ...
    }
  } catch (error) {
    if (!(error instanceof RuntimeSourceError)) throw error;
    ...show "Evaluation stuck"...
  }
```

But isContractible is called in several places beyond the stepper loop, for example in VariableDeclaration.ts:34,
BlockStatement.ts:38, ExpressionStatement.ts:32, and oneStep itself. These callers don't have try-catch, so the throw
propagates correctly by accident, it bubbles up to the stepper loop's catch.

Akshay-2007-1 (approved):

PR #1985, opened 2026-06-02
Update dependency @types/estree to v1.0.9 (with type adaptations)

martin-henz:
Updates `@types/estree` to **1.0.9**. Unlike a normal Renovate lockfile bump, this version's updated ESTree types model newer ECMAScript constructs, which broke compilation in 28 places. This PR includes the source adaptations so the build passes.

### Changes
- **Import/export specifier endpoints are now `Identifier | Literal`** — ES2022 allows arbitrary module namespace names (e.g. `export { x as "y" }`). Added a `getSpecifierName` helper in `utils/ast/helpers.ts` and routed `.name` accesses through it. Source only supports identifier names, so it falls back to the literal's value.
- **`ImportDeclaration` now requires `attributes`** (import attributes) — set to `[]` in the `importDeclaration` AST constructor.
- **`BinaryExpression.left` is now `Expression | PrivateIdentifier`** (private-in expressions) — narrowed with casts in the transpiler and stepper, since private fields are not valid in Source.

### Verification
- `yarn build` passes (0 type errors)
- eslint clean on changed files
- 885 tests pass across preprocessor / transpiler / stepper / stdlib / parser / cse-machine / finder / utils-ast

### Note
Supersedes the lockfile-only Renovate PR #1766, which fails CI on its own because it lacks these source changes. That PR can be closed once this merges.

### DIFF
--- a/src/cse-machine/interpreter.ts
+++ b/src/cse-machine/interpreter.ts
@@ -24,6 +24,7 @@ import * as ast from '../utils/ast/astCreator';
 import {
   filterImportDeclarations,
   getSourceVariableDeclaration,
+  getSpecifierName,
   hasNoDeclarations,
   hasNoImportDeclarations,
 } from '../utils/ast/helpers';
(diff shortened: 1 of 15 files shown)


RichDom2185 (approved):
Looks good, thanks!

PR #2004, opened 2026-06-22
feat(conductor): Source/JS CSE machine evaluators and plugin

Akshay-2007-1:
## Summary

Adds conductor-based evaluators for Source 1–4 and a JS CSE machine plugin, enabling the CSE machine visualiser to work with Source/JS programs via the Conductor framework.

### New files

- **`src/conductor/JSCseEvaluator.ts`**: Extends `BasicEvaluator` to run Source/JS code through the CSE engine. Registers a `CseMachinePlugin` (from `@sourceacademy/runner-cse-machine`) and sends a full batch of `CseSnapshot`s to the host after each evaluation.
- **`src/conductor/SourceEvaluator1.ts`**: Thin evaluator for Source 1 (non-CSE chapters) via conductor.
- **`src/conductor/evaluator.ts`** / **`index.ts`** / **`initialise.ts`**: Entry points and exports for the conductor evaluator bundle.
- **`rollup.config.evaluator.mjs`** / **`scripts/build-evaluators.mjs`**: Build configuration for producing the Web Worker evaluator bundles.


martin-henz:
Closing as superseded — with thanks, this PR got the Conductor migration moving.

Source now runs under Conductor via three slices that landed since:

- #2058 — transpiler evaluators for Source §1–§4
- #2061 — CSE machine evaluators for §3–§4
- #2064 — the substitution stepper for §1–§2

Two things from here were carried forward directly, and both were the non-obvious parts: the rollup browser shims (`path` → path-browserify, the `inspector` stub, the `global`/`process` banner) and the `rewriteStaticRequires` transform for `require('acorn-class-fields')`. I initially doubted that second one was necessary and tried to do without it — the first build put a literal `require(...)` in an IIFE bundle, which would have killed the worker at load before Conductor could report anything. It was necessary, and it is credited in `rollup.config.evaluator.mjs`.

The evaluators are deployed at https://source-academy.github.io/js-slang/ and Source §1–§4 are back in the language directory.

Issue #2057, opened 2026-09-21
Conductor: CSE machine visualisation for programs that import modules


martin-henz:
**Deliberately postponed — feature parity first.**

A synchronous CSE evaluator cannot call Conductor module closures, so programs that import a module will not be visualisable in the CSE tab. Recording the reasoning now so it does not have to be rediscovered.

## Why this is not urgent

In js-slang the CSE machine exists for *visualisation*; the workhorse is the transpiler. So unlike py-slang — whose CSE machine is its primary engine and is `async` throughout — js-slang's can stay synchronous, and the large async refactor of `src/cse-machine/` (~2,570 LOC) is off the critical path entirely.

## Why it is not as hard as it looks, when picked up

Conductor's module interop is async end-to-end: `ModuleLoaderRunnerPlugin.requestModule()` returns a promise, and `ExternCallable` is `(...args) => AsyncGenerator<...>`. There is a `.sync` escape hatch (see conductor's `BaseModulePlugin`, and `closure_call_sync` in py-slang's `GenericDataHandler`), and real modules do use it — `pix_n_flix` attaches 11 `.sync` twins, `sound` 7 — but only on hot inner-loop accessors and module→user callbacks. Top-level functions that do a genuine host round-trip (`show`, `play`) are inherently async and can never carry one.

The key observation is that **this probably does not require an async interpreter**. `generateCSEMachineStateStream` is already a `function*`, so the *driver* (`collectSnapshots`) can be async and `await` between `next()` calls, with the step handler pushing a continuation onto the control and the driver pushing the result onto the stash. That is exactly what py-slang's `src/engines/cse/modules.ts` does. Bounded work in the instruction set, not a rewrite.

## Related

Module support for the transpiler is the separate, larger problem: it needs dual-mode compilation (a sync body and an async body per function, gated on whether the chunk imports anything), which is what py-slang's py2js engine does. See its `src/engines/py2js/README.md`.

Issue #2053, opened 2026-09-21
Conductor: Source Typed (Variant.TYPED) evaluators


martin-henz:
Conductor evaluators for `Variant.TYPED` (Source Typed), §1–4.

Deferred from #2051, which covers `Variant.DEFAULT` only.

Notes for whoever picks this up:

- The typed parser (`src/parser/source/typed/typeParser.ts`) is already in the evaluator bundles — it is reached statically from `src/parser/parser.ts` — so no extra bundling work is expected. It is also the file whose `require('acorn-class-fields')` needs the build-time rewrite added in #2051.
- Error mapping can be improved here. `src/conductor/errors.ts` currently maps `ErrorType.TYPE` to the generic `EvaluatorError`, because a `SourceError` exposes no structured `expected`/`actual` — those are baked into `explain()`. The static type checker's errors (`src/errors/typeErrors.ts`) *do* carry that structure, so they can be upgraded to conductor's `EvaluatorTypeError`, which renders `(expected X, got Y)` itself.
- `docs/specs/source_{1,2,3,4}_typed_typing.tex` exist and give the typed variant its own typing rules. Worth checking whether the operator conformance sweep added in #2051 should be extended over them, or whether static typing makes the dynamic table moot for this variant.

Issue #2052, opened 2026-09-21
Module imports fail in a Web Worker: browser detection tests for `window`, which Workers lack


martin-henz:
## Summary

`getBundleAndTabImporter()` picks its dynamic-import strategy by testing for `window`:

https://github.com/source-academy/js-slang/blob/master/src/modules/loader/importers.ts#L85-L96

```ts
function getBundleAndTabImporter(): Importer<PartialSourceModule> {
  if (process.env.NODE_ENV === 'test') {
    return p => import(p);
  }

  if (typeof window !== 'undefined') {
    return new Function('path', 'return import(`${path}?q=${Date.now()}`)') as any;
  }

  // eslint-disable-next-line @typescript-eslint/no-require-imports
  return p => Promise.resolve(require(p));
}
```

A **Web Worker has no `window`** — it has `self`. So in a Worker this falls through to the Node branch and returns a closure calling `require`, which does not exist in a browser bundle. Any module import then fails with a `ReferenceError`.

The accompanying comment frames the choice as "browser vs Node", but the test used only distinguishes *the browser main thread* from everything else. A Worker is a browser environment that supports native `import()` and lands in the Node branch anyway.

## Why it matters now

The Conductor evaluators added in #2051 run inside exactly such a Worker. Confirmed present in the built bundle — after rollup, the surviving free `require` is this one:

```js
"undefined" != typeof window
  ? new Function("path", "return import(`${path}?q=${Date.now()}`)")
  : e => Promise.resolve(require(e))
```

It fails **lazily**, only when a program actually imports a Source module, so it does not affect the transpiler evaluators as shipped (module support is out of scope there). It blocks the module-support slice of the Conductor migration, and it is a latent bug for any other Worker-based embedding independent of Conductor.

## Suggested fix

Widen the browser test, e.g. `typeof window !== 'undefined' || typeof self !== 'undefined'`, or feature-detect support for dynamic `import()` rather than testing for a DOM global. The `new Function(...)` wrapper exists to hide the `import()` from bundlers (see the comment below the function) and is equally needed in a Worker, so the browser branch itself should be reusable as-is.

Worth checking the cache-busting query string (`?q=${Date.now()}`) still behaves sensibly in a Worker context while touching this.

## Reproducing

Build a Conductor evaluator (`yarn build:evaluators`), serve it, select Source in the Playground with the Conductor flag on, and run a program with an `import` from a Source module.

PR #2058, opened 2026-09-21
feat(conductor): Source §1–4 transpiler evaluators + operator conformance sweep

### LINKED ISSUE #2051: Conductor migration, slice 1: Source §1–4 transpiler evaluators
## Background

js-slang is the last major Source Academy language not running under [Conductor](https://github.com/source-academy/conductor). Source was removed from the exported language directory in [language-directory#44](https://github.com/source-academy/language-directory) ("Remove Source and Java from the exported language directory") because the only conductor evaluators that existed — in the `js-slang-lean` fork — were too limited. The frontend now defaults `conductor.enable` to **true** and carries an explicit stopgap for us: `selectConductorEnable` in `frontend/src/features/conductor/flagConductorEnable.ts` force-disables conductor on `/sicpjs` routes, with the comment *"Remove this override once Source is migrated into the Language Directory."*

py-slang completed the equivalent migration and is the reference implementation. The shared infrastructure it needed now exists and is stable: `@sourceacademy/conductor` 0.8.3 on npm, and the common/runner/web plugin triples in [`source-academy/plugins`](https://github.com/source-academy/plugins).

This issue is **slice 1**: the workhorse evaluator path only. It deliberately excludes the CSE machine, the stepper, modules, and every non-default variant, each of which gets its own issue.

Prior art, for reference but not to be rebased: #2004 (conductor 0.5.0, 49 commits behind, conflicting) and the `js-slang-lean` fork (js-slang v1.0.84, conductor 0.2.1). Both are superseded by this work. The genuinely useful part of #2004 is its rollup browser shims, called out below.

## Scope

**In scope**
- `SourceEvaluator1` … `SourceEvaluator4` — conductor evaluators for Source §1–4, `Variant.DEFAULT`, running on the **transpiler** path (`ExecutionMethod = 'native'`, `Variant.DEFAULT` — see the naming trap below)
- `display()` / `error()` output, error reporting, and `prompt()` wired to the conductor host
- A rollup build producing one IIFE bundle per evaluator, plus a build driver script
- Manual verification in the frontend via the `conductor.evaluator.url` override

**Explicitly out of scope** (separate issues)
- CSE machine evaluators and the `__cse` channel
- Stepper, data visualizer, autocomplete plugins
- Module support (`import ... from 'rune'`) — needs dual-mode transpilation
- `Variant.TYPED`, `Variant.EXPLICIT_CONTROL`, `Variant.WASM`
- `FULL_JS`, `FULL_TS`, `HTML` chapters
- CI deploy to GitHub Pages, and the `language-directory` cutover

A program that `import`s a Source module is expected to fail in this slice. That is acceptable and tracked separately.

## Deliverables

```
src/conductor/index.ts              exports SourceEvaluator1..4
src/conductor/initialise.ts         rollup entry: initialise(__EVALUATOR__)
src/conductor/SourceEvaluator.ts    abstract base + 4 concrete subclasses
src/conductor/errors.ts             SourceError -> ConductorError mapping
rollup.config.evaluator.mjs         per-evaluator IIFE bundle config
scripts/build-evaluators.ts         build driver (EVALUATOR env var per target)
scripts/evaluator-shims/path.mjs    path-browserify re-export
scripts/evaluator-shims/empty.mjs   stub for node-only imports
```

Plus these small changes to existing files:

- **`tsconfig.prod.json`** — add `src/conductor/**` to `exclude`. `@sourceacademy/conductor` is ESM-only (`"type": "module"`, `exports` declares only `import`); `build:slang` is `tsc` with `"module": "commonjs"`, so without this the published `dist/` gets `require()` calls that fail at runtime.
- ~~**`src/parser/parser.ts:7`** — make the `FullTSParser` import dynamic.~~ **Superseded during implementation:** aliasing `@ts-morph/bootstrap` itself to a throwing stub at build time achieves the same thing with *no* js-slang source change, and avoids making the synchronous `parse()` async (which a dynamic import would have forced, rippling through every caller). Only `FullTSParser` (`Chapter.FULL_TS`) reaches it, and these evaluators never select that chapter. Result: 1.4 MB bundles, zero TypeScript compiler.
- **`src/parser/source/typed/typeParser.ts:1093`** — replace the inline `require('acorn-class-fields')` with a static ESM import. `@rollup/plugin-commonjs` cannot rewrite it (it skips modules its acorn parser rejects as TS). #2004 worked around this with a custom transform; a static import removes the need. js-slang compiles to CJS anyway, so emit is unchanged.
- **`package.json`** — add `@sourceacademy/conductor@^0.8.3` and rollup devDependencies; add a `build:evaluators` script.

### Evaluator shape

Extend `BasicEvaluator` and implement `evaluateChunk` only. **Do not override `startEvaluator`.** As of conductor 0.8.3 the base class already runs the REPL loop and emits `RunnerStatus.EVAL_READY` / `RUNNING` / `ERROR` around each chunk, and calls `sendResult` with whatever `evaluateChunk` returns. (py-slang's `PyStepperEvaluator` overrides it only because Python's exec model wants a one-shot run; Source is expression-oriented and the default loop is correct for us.)

So `evaluateChunk` should **return** the program's value rather than calling `sendResult` itself — calling both produces two entries in the host's REPL.

### Execution method — and a naming trap

js-slang has **two unrelated things called "native"**. Getting them confused here would be costly, so, explicitly:

- **`ExecutionMethod = 'native'`** selects *which engine runs the program*: `runners.native` (transpile to JS, then `sandboxedEval`) as opposed to `runners['cse-machine']` (interpret). "Native" here means "runs on the browser's own JS engine". For a Source chapter on the default variant it still applies the **full** Source instrumentation — `transpileToSource` does tail calls, call-expression function checks, operators rewritten to function calls, boolean checks, property access/assignment guards, and infinite-loop protection.
- **`Variant.NATIVE`** is a *language variant* meaning "run the JavaScript essentially as written". `transpile()` dispatches on it (`src/transpiler/transpiler.ts:548-552`) to `transpileToFullJS`, which does almost nothing: import rewriting, an undefined-variable check, and the evaller replacer. It is not a selectable Source language — it is absent from `sourceLanguages` in `src/langs.ts` and appears in only three places repo-wide.

**This slice wants the first and not the second.**

Pass `executionMethod: 'native'` explicitly to `runFilesInContext`. The default `'auto'` silently falls back to the CSE machine when verbose errors are on or a `debugger;` statement is present (`src/runner/utils.ts:47-59`), which is out of scope here and would be confusing behaviour for this evaluator. Pinning it is a deliberate choice; revisit when the CSE evaluator lands.

⚠️ Keep the variant at **`Variant.DEFAULT`**.

What `Variant.NATIVE` actually changes is narrower than "no restrictions", and worth being precise about. Chapter *syntax* restrictions are enforced at parse time by `SourceParser.validate`, and `getDisallowedSyntaxes()` filters on `this.chapter` **only** — the variant is not consulted (`src/parser/source/index.ts:105-111`). So §1 + `Variant.NATIVE` still rejects `while` loops.

What it drops is everything `transpileToSource` adds and `transpileToFullJS` does not — i.e. the *runtime* guards:

- `transformCallExpressionsToCheckIfFunction` — calling a non-function
- `transformUnaryAndBinaryOperationsToFunctionCalls` — Source's operator type checks (`1 + "a"`)
- `transformSomeExpressionsToCheckIfBoolean` — `if (1)`
- `transformPropertyAccess` / `transformPropertyAssignment` — array access guards
- `transformReturnStatementsToAllowProperTailCalls` — **proper tail calls**. Not a safety guard but a *language semantics* guarantee: it reifies tail calls into `{isTail, function, arguments}` records that `callIfFuncAndRightArgs` (`src/utils/operators.ts:296-310`) trampolines in a loop, giving Source constant-stack tail recursion on JS engines that do not implement ES6 PTC (V8 and SpiderMonkey do not; JavaScriptCore does). Source specifies proper tail calls; without this transform a *correct* Source program that recurses deeply in tail position overflows the stack. Note the entanglement: the trampoline lives inside `callIfFuncAndRightArgs`, the same wrapper doing the arity and non-function-call checks, so "keep tail calls, drop the type checks" is not available as a middle option.
- `wrapArrowFunctionsToAllowNormalCallsAndNiceToString` — arity checks and function `toString`
- `addInfiniteLoopProtection` — the execution timeout

So the failure mode of getting this wrong is not "§1 accepts anything" but something subtler and harder to notice: §1 still *looks* correct, but `1 + "a"` quietly evaluates to `"1a"` instead of raising a type error, a runaway loop never times out, and — the one that is a genuine semantic divergence rather than a missing guard — deep tail recursion overflows the stack. See the acceptance criteria for the guard against this.

### Output, input, errors

Construct the context with `externalBuiltIns` (`createContext`'s 6th parameter):

| js-slang | conductor |
|---|---|
| `rawDisplay(value, str)` | `conductor.sendOutput(...)` — the `__stdio` channel |
| `prompt(value, str)` | `conductor.tryRequestInput()` |
| `alert(value, str)` | `conductor.sendOutput(...)` |
| `visualiseList` | throw — data visualizer is out of scope |

Note `IRunnerPlugin.tryRequestInput(): string | undefined` is **synchronous**, which matches js-slang's `CustomBuiltIns.prompt: (...) => string | null` exactly. No async plumbing is needed for `prompt` in this slice.

⚠️ This is the thing #2004 gets wrong and is worth being careful about: it never passes `externalBuiltIns`, so js-slang's default `rawDisplay` stays in place — which is `console.log` (`src/stdlib/misc.ts:16`), inside a Web Worker. `display()` output never reaches the REPL at all. `js-slang-lean` wired this correctly; #2004 regressed it.

### Error mapping (`src/conductor/errors.ts`)

Send **one `ConductorError` per `SourceError`** on the `__error` channel, rather than one bulk `parseError(...)` string. `EvaluatorError`'s constructor is `(message, line?, column?, fileName?)` and js-slang's `SourceError` carries `location.start.{line,column}` plus `explain()`, so the mapping is direct and the host gets real positions.

- `ErrorType.SYNTAX` → `EvaluatorSyntaxError`
- `ErrorType.TYPE` → `EvaluatorTypeError`
- `ErrorType.RUNTIME`, `ErrorType.IMPORT` → `EvaluatorRuntimeError`
- `ErrorSeverity.WARNING` → `sendOutput`, **not** `sendError` (warnings render red otherwise)

Clear `context.errors` after reporting, so a bad REPL line doesn't leak into the next one.

### Context lifetime

One `Context` per evaluator instance, created once and reused across chunks, so a later REPL line sees an earlier one's declarations. This matches `js-slang-lean`'s behaviour and the frontend's existing REPL semantics.

### Build

Follow py-slang's `scripts/build.ts`: one rollup child process per target, `EVALUATOR` env var selecting the target, `@rollup/plugin-replace` substituting `__EVALUATOR__`, IIFE output to `dist/<Name>.js`. Port its concurrency cap (`cpus().length`) — py-slang found that firing every target at once on a 2-vCPU CI runner caused sub-second builds to balloon past 5 minutes and eventually kill the runner.

Salvage from #2004: the `path` → `path-browserify` and `inspector` → empty-stub aliases, and the `global` / `process` banner. These are the non-obvious bits and they were correct.

## Acceptance criteria

**Gate (must pass):**
1. `yarn build:evaluators` produces `dist/SourceEvaluator{1,2,3,4}.js`
2. No bundle contains the TypeScript compiler (check: bundle size is sane, and `grep -c "createProjectSync" dist/SourceEvaluator1.js` is 0)
3. `yarn test` — the existing 82 test files — still passes
4. `yarn build:slang` produces a `dist/` with no `src/conductor` output and no `require("@sourceacademy/conductor...")`
5. In the frontend with `conductor.evaluator.url` pointed at a locally-served `SourceEvaluator1.js`:
   - `1 + 1;` → REPL shows `2`
   - `display("hi");` → `hi` appears in default colour, and the returned value is shown
   - `const x = 1;` then `x + 1;` **as two REPL entries** → `2`. Note this is specifically the REPL input box, not the Run button: pressing Run starts a fresh program (`evaluateFile`, new worker) and correctly does *not* see earlier state, while REPL entries are chunks against the same worker and do. `BasicEvaluator`'s loop implements exactly that split, so both behaviours are correct and only the REPL one tests context reuse.
   - `1 +;` → a **red** error naming the right line
   - `head(list(1,2));` in §1 → a red error about `head` not being declared, not a crash
   - `1 + "a";` in §1 → a red runtime type error ("Expected number on right hand side of operation, got string"). **This is the guard against the `Variant.NATIVE` trap above** — if it evaluates to `"1a"`, the variant is wrong. A syntax-level test such as `while (true) {}` does *not* work as a guard here: chapter syntax restrictions are variant-independent and would reject it either way.
   - the run spinner stops after each run

**Should pass, verify and report:**
6. §2 `head(list(1,2,3));` → `1`; §3 `const a = [1,2,3]; a[0];` → `1`; §4 `parse("1+1;");` returns a tagged list
7. `prompt("name?")` round-trips

## Manual test setup

Test through the **real** frontend path — language dropdown, evaluator selection, welcome text — rather than the single-evaluator `conductor.evaluator.url` override. No frontend source changes are needed, because `directory.language.url` is already a public feature flag (`frontend/src/commons/featureFlags/publicFlags.ts`), unlike `conductor.evaluator.url` which is not.

`LanguageDirectorySaga` fetches that URL and uses the returned JSON whenever the flag differs from its default (`LanguageDirectorySaga.ts:133-143`), so pointing it at a locally-served directory is enough to make Source appear in the dropdown.

1. **In `../language-directory`** (local only, do not commit):
   - re-add `...sourceLanguage` to the `languages` array in `src/languages.ts` — `src/languages/source/source{1..4}.ts` still exist and only need re-exporting
   - point each `path` at the locally-served bundle, e.g. `http://localhost:4001/SourceEvaluator1.js`
   - `yarn start` — this already runs `rollup -cw` plus `http-server ./dist -p8125 -c-1 --cors`, i.e. a CORS-enabled watch server on port 8125, purpose-built for this
2. **In js-slang**: `yarn build:evaluators`, then serve `dist/` on port 4001
3. **In the frontend**: Settings → Feature Flags → `conductor.enable` on (already the default), `directory.language.url` = `http://localhost:8125/directory.json`

Source §1–4 then appear in the language dropdown and run through the normal conductor path.

There is precedent for this already in the committed directory: the `python4Cse` entry points at `http://localhost:4001/PyCseEvaluator4.js`, i.e. someone's local py-slang build.

Driving the checks in the acceptance criteria through the Claude Chrome extension is fine and probably faster than doing them by hand.

## Risks

- **Bundle size / node-isms.** js-slang has more Node surface than py-slang (`@ts-morph/bootstrap`, `source-map`, `path` in the module preprocessor, `lodash`). The ts-morph pull-in is known and handled above; others may surface only at bundle time. This is the main reason slice 1 exists as a spike.
- **`executionMethod: 'native'` pinning** may surprise on `debugger;` — noted as a deliberate, revisitable choice.
- **The transpiler's `new Function` / `eval`** (`src/transpiler/evalContainer.ts`) must work inside a Web Worker under the frontend's CSP. Expected to be fine (py2js does the same), but unverified for js-slang specifically.

## Follow-ups

Filed:

- #2052 — module imports fail in a Web Worker (`window` detection), found during this slice
- #2053 — Source Typed (`Variant.TYPED`) evaluators
- #2054 — explicit-control variant evaluator
- #2055 — WASM / SVM evaluator
- #2056 — `FULL_JS`, `FULL_TS`, `HTML` chapters
- #2057 — CSE machine visualisation for programs importing modules (postponed)

Not yet filed, in planned order: CSE machine evaluators · stepper · data visualizer · autocomplete · module support (dual-mode transpilation) · CI deploy and the `language-directory` cutover.

---

## Implementation notes

Landed on `conductor-slice-1` as two commits: the evaluators and build, then the operator conformance sweep.

Three deviations from the plan above, all in the direction of touching less js-slang source:

1. **`@ts-morph/bootstrap` stubbed at build time** rather than lazily imported — see the struck-through item above.
2. **`require('acorn-class-fields')` kept as-is in source**, rewritten to an ESM import by a rollup transform (ported from #2004). I had expected running esbuild before `@rollup/plugin-commonjs` to let `transformMixedEsModules` handle it; it does not — the first build emitted a literal `require("acorn-class-fields")` at module top level, which in an IIFE is a free variable, so the worker would have died with a `ReferenceError` at load, before conductor could report anything. #2004's transform was necessary, not incidental. The source cannot simply switch to an `import`: the package is CJS (`module.exports = fn`, no `.default`) and this project compiles with `esModuleInterop` off, so a default import emits `.default` (undefined) in the published library, while a namespace import is what the CJS build needs but breaks under rollup interop.
3. **Operator conformance testing included** rather than deferred to its own PR — the spec tables (`docs/specs/source_typing{,_3}.tex`) already existed and the sweep is what makes the CSE evaluator checkable rather than eyeballed when it arrives.

Found along the way and filed separately: **#2052** — module imports fail in a Web Worker, because `src/modules/loader/importers.ts` detects the browser by testing for `window`, which Workers lack. Lazy failure, so it does not affect this slice, but it blocks module support.

Verified in the real frontend against a locally-served language directory: `1 + 1` → `2`; `display("hi")` on the stdout channel; `1 +;` red with position; `head(list(1,2))` and `while (true) {}` rejected at §1; `1 + "a"` rejected (the `Variant.NATIVE` guard); REPL context persistence; and 200,000-deep tail recursion returning rather than overflowing, confirming the trampoline is active.

Final state: 4 bundles at 1.4 MB, 83 test files / 3846 tests passing, `eslint` clean, `build:slang` free of conductor.

martin-henz:
Closes #2051.

First slice of the Conductor migration: the **workhorse evaluator path only**. No CSE machine, no stepper, no modules, no non-default variants — each of those is filed separately (#2053–#2057) and none is started.

A program that `import`s a Source module is expected to fail here. That is in scope to be broken, and is tracked separately.

## What's in it

**`SourceEvaluator1..4`** — Source §1–4, `Variant.DEFAULT`, on the transpiler. They extend `BasicEvaluator` and implement `evaluateChunk` alone: as of conductor 0.8.3 the base class already owns the REPL loop, emits `EVAL_READY`/`RUNNING`/`ERROR` around each chunk, and calls `sendResult` with whatever `evaluateChunk` returns. So the evaluator *returns* the program's value rather than sending it — doing both would show every run twice in the host's REPL.

**Output, input, errors.** `display()`/`prompt()`/`alert()` are wired through `externalBuiltIns` to `sendOutput`/`tryRequestInput`. `tryRequestInput` is conductor's *synchronous* counterpart to `requestInput`, which matches js-slang's synchronous `CustomBuiltIns.prompt` exactly, so `prompt` needs no async plumbing. Errors are sent one per `SourceError` rather than as a single bulk `parseError` string, so each keeps its own position; warnings go to stdout, since the host renders the error channel in red.

**Build.** One IIFE bundle per evaluator (`yarn build:evaluators`), 1.4 MB each, driven by `scripts/build-evaluators.mjs` with a concurrency cap.

**Operator conformance sweep** — a table-driven suite transcribed from `docs/specs/source_typing.tex` (§1/§2) and `docs/specs/source_typing_3.tex` (§3/§4). 1002 tests. See the second commit.

## Two naming traps worth knowing before reviewing

**`ExecutionMethod = 'native'` is not `Variant.NATIVE`.** The former picks *which engine runs the program* — `runners.native` (transpile to JS, then eval) as opposed to the CSE machine — and still applies the full Source instrumentation. The latter is a *language variant* that routes to `transpileToFullJS`, which skips it. This PR pins the first and keeps the variant at `DEFAULT`.

Getting that wrong fails subtly rather than loudly: chapter *syntax* restrictions are enforced at parse time and are variant-independent (`getDisallowedSyntaxes()` filters on chapter only), so §1 would still reject `while` loops and look fine — while `1 + "a"` quietly evaluated to `"1a"`, and deep tail recursion overflowed the stack. `1 + "a";` is in the acceptance checks precisely as the guard for this.

**Proper tail calls are semantics, not a safety net.** V8 and SpiderMonkey never shipped ES6 PTC, so `transformReturnStatementsToAllowProperTailCalls` — which reifies tail calls into records that `callIfFuncAndRightArgs` trampolines — is the only reason Source has constant-stack tail recursion in a browser. Verified live below.

## Three deviations from the issue's plan

All in the direction of touching less js-slang source:

1. **`@ts-morph/bootstrap` stubbed at build time** rather than `FullTSParser` lazily imported. It drags in the whole TypeScript compiler (~24 MB on disk) and is reached only via `Chapter.FULL_TS`, which these evaluators never select. A dynamic import would have forced the synchronous `parse()` to become async, rippling through every caller; a build-time alias needs no source change at all.

2. **`require('acorn-class-fields')` left as-is**, rewritten to an ESM import by a rollup transform (ported from #2004, credited in the config). I expected running esbuild before `@rollup/plugin-commonjs` to let `transformMixedEsModules` handle this and was wrong — the first build emitted a literal `require("acorn-class-fields")` at module top level, which in an IIFE is a free variable, so the worker would have died with a `ReferenceError` at load, before conductor could report anything. The source cannot simply switch to an `import`: the package is CJS with no `.default`, and this project compiles with `esModuleInterop` off, so a default import breaks the published library while a namespace import breaks the bundle. Build time is the right layer.

3. **Operator conformance testing included** rather than deferred. The spec tables already existed, and the sweep is what will make the CSE evaluator *checked* rather than eyeballed when it lands.

## Verification

Full suite: **83 files, 3846 tests pass** (was 82 / 2844). `eslint` clean. `yarn build:slang` produces a `dist/` with no `src/conductor` output and no `require("@sourceacademy/conductor")` — conductor is ESM-only and the library build emits CommonJS, hence the `tsconfig.prod.json` exclude.

Checked in the real frontend against a locally served language directory (Source re-added, evaluator paths repointed):

| input | result |
|---|---|
| `1 + 1;` | `2` |
| `display("hi");` | `"hi"` on the stdout channel, plus the returned value |
| `1 +;` | red, `1:3: SyntaxError: Unexpected token` |
| `head(list(1,2));` at §1 | red, `Name head not declared.` |
| `while (true) {}` at §1 | red, `While statements are not allowed` |
| `1 + "a";` | red, `Expected number on right hand side of operation, got string.` |
| `const y = 41;` then `y + 1;` as REPL chunks | `42` |
| `function f(x) { return x === 0 ? 0 : f(x - 1); } f(200000);` | `0`, not a stack overflow |

The conformance suite was checked to be non-vacuous by mutation: claiming `===` is `any × any` at §1/§2 fails 56 cases; claiming `string + string -> number` fails 4.

## Found on the way

**#2052** — module imports fail in a Web Worker: `src/modules/loader/importers.ts` detects the browser by testing for `window`, which a Worker lacks, so it falls through to a `require` that does not exist in the bundle. Lazy failure, so this PR is unaffected, but it blocks module support and is arguably a js-slang bug independent of Conductor.

## Not touched

#2004 is left alone. Its rollup browser shims were the genuinely useful part and are carried forward here with credit.


martin-henz:
@codex review

martin-henz:
Thanks — the finding is factually right, and I've addressed it, though not with the suggested change.

**Confirmed:** `runFilesInContext` defaults `shouldAddFileName` to `Object.keys(files).length > 1` ([`src/index.ts:258`](https://github.com/source-academy/js-slang/blob/master/src/index.ts#L258)), this evaluator always passes exactly one file, so the linker parses without `sourceFile`, `error.location.source` stays undefined, and `toConductorError` never sees a filename. Verified live in the playground: errors render as `1:3: SyntaxError: Unexpected token`, with no file prefix.

**But I don't think `shouldAddFileName: true` is the right fix here.** That default isn't an oversight — it exists so a single-file program doesn't carry a redundant filename on every diagnostic. Forcing it on would turn every error in the Playground into `/program.js:1:3: ...`, where the `/program.js` part disambiguates nothing: there is only one file, and the REPL is already scoped to it. Conductor's `EvaluatorError` renders `${fileName}:${line}:${column}: ` verbatim, so the noise would be unconditional.

The actual defect was the **comment**, which claimed the entrypoint capture made errors "carry the right file". That was false, and it's what made the code look like it had a bug. Fixed in 866f3dfc, which states what the capture really does (it's the file-map key and entrypoint argument — `validateFilePath` and the module preprocessor's resolution both key off it) and records the condition under which the suggestion becomes correct: when this evaluator grows real multi-file support (local imports, folder mode), `shouldAddFileName: true` should land in the same change, because then the filename genuinely disambiguates.

Multi-file support is out of scope for this PR by design — see the scope section in #2051.

PR #2061, opened 2026-09-21
feat(conductor): CSE machine evaluators for Source §3–§4

### LINKED ISSUE #2060: Conductor migration, slice 2: CSE machine evaluators for Source §3–§4
## Background

Slice 1 (#2051, merged in #2058) put Source §1–4 on the transpiler under Conductor. This slice adds the CSE machine evaluators, so the CSE tab works for Conductor-run Source.

In js-slang the CSE machine exists **for visualisation** — the transpiler is the workhorse. That shapes everything below: the evaluator can stay synchronous, and the bar is not "it renders something" but *parity with the CSE tab students use today*.

## Scope

**In scope**
- `SourceCseEvaluator3` and `SourceCseEvaluator4` — chapters §3 and §4, `Variant.DEFAULT`
- Serialisation of js-slang's control / stash / environments into `CseSnapshot`
- `RunConfig` (`stepLimit`, `breakpointLines`) via the host's `/__cse_config__`
- Conformance of this engine against the transpiler on the operator tables

**Out of scope**
- Modules in the CSE tab — deliberately postponed, see #2057
- Any async refactor of `src/cse-machine/` — not needed, see below
- §1/§2 CSE evaluators: the tab is gated at `chapter >= Chapter.SOURCE_3` in the frontend (`ApplicationTypes.ts:245`), matching `python3Cse`/`python4Cse` in the language directory
- Stepper, data visualizer, autocomplete, the `language-directory` cutover

## Why §3 and §4 only

`supportedFeatures.cseMachine = chapter >= Chapter.SOURCE_3` (`frontend/src/commons/application/ApplicationTypes.ts:245`). Under Conductor the tab shows when the language offers an evaluator carrying `EvaluatorCapability.CSE` (`Playground.tsx:815`), so §1/§2 should simply not have one — same as Python, which lists `python3Cse` and `python4Cse` and nothing below.

## The evaluator stays synchronous

py-slang's CSE machine is `async` throughout because it is that language's primary engine and must await module round-trips. js-slang's is not, and `generateCSEMachineStateStream` is a plain `function*` (`src/cse-machine/interpreter.ts:317`). Keep it that way.

If module support in the CSE tab is ever picked up (#2057), note that it still probably does not need an async interpreter: the *driver* can be async and `await` between `next()` calls, with the step handler pushing a continuation onto the control and the driver pushing the result onto the stash. That is what py-slang's `src/engines/cse/modules.ts` does. Not this issue's problem.

## What has to be written

Only the serialiser. The plugin trio in [`source-academy/plugins`](https://github.com/source-academy/plugins) is **264 LOC of pure transport** — `common-cse-machine` (the protocol), `runner-cse-machine` (`sendSnapshots`), `web-cse-machine` (an abstract `receiveSnapshots`). The renderer stays in the frontend, which `web-cse-machine` says explicitly: the visualisation layer is "tightly coupled to the existing CSE machine UI".

So the data path is:

```
js-slang Control/Stash/Environment
  -> CseSnapshot (plain JSON, structured-clone-able)
  -> __cse channel
  -> frontend CseSnapshotAdapter.ts, which rebuilds *fake js-slang objects*
  -> the existing CseMachine renderer
```

For Source specifically that is a round trip: serialise js-slang state, then reconstruct js-slang-shaped objects at the other end. **Every lossy shortcut in the serialiser shows up as a visible regression against the current CSE tab.** This is the main risk in this slice and the reason the acceptance bar below is what it is.

`frontend/src/features/cseMachine/CseSnapshotAdapter.ts` is the contract to write against — it documents exactly which duck-typed shapes the renderer's guards accept.

## Protocol

`@sourceacademy/common-cse-machine@^0.3.0`:

- `CseSnapshot` — `{ stepIndex, control[], stash[], environments[], currentLine? }`, control and stash ordered **top first**
- `CseSerializedValue` — `{ displayValue, label, tag?, metadata? }`
- `CseSerializedInstruction` — `{ displayText, tag?, metadata? }` (no `label`: instructions are operations, not typed values)
- `CseSerializedEnvFrame` — `{ id, name, parentId, closureFrameId?, bindings[], heapObjects?, isActive, isOnCallStack?, globalNames? }`
- `CseMachinePlugin.sendSnapshots(snapshots, breakpointSteps?)` — note the second argument, added in `runner-cse-machine@3.0.0`

## Run config

The host already serves it — no frontend work needed. `evalCode.ts:1023` writes `'/__cse_config__': JSON.stringify({ stepLimit, breakpointLines })`, fetched via `conductor.requestFile('/__cse_config__')`. Mirror py-slang's `src/conductor/runConfig.ts`: a typed `RunConfig`, with a missing or malformed file falling back to `{}` rather than failing the run.

`breakpointLines` are the editor's gutter breakpoints, to be resolved to their enclosing statement and reported back as `breakpointSteps` (0-based step indices) so the host's breakpoint navigation works.

## Acceptance criteria

**Parity, not plausibility.** #2004's serialiser is the cautionary example: it hid every global builtin, truncated arrays at depth 2, and deduplicated steps by display-text fingerprint. Each of those is a silent divergence from the tab it is replacing.

1. For a fixed corpus of Source §3/§4 programs, the snapshot sequence agrees with the raw machine: same step count (modulo a documented and deliberate rule), same control depth per step, same stash values under `stringify`, same environment chain shape
2. Closures carry their defining frame, so the renderer draws the right arrows
3. Arrays serialise to full depth, or the truncation rule is explicit, documented, and visible in the UI rather than silent
4. Global-frame builtins are handled the same way the current tab handles them — check what it does before choosing
5. `stepLimit` from `/__cse_config__` is honoured (py-slang had to fix a hardcoded 1000 — source-academy/py-slang#427)
6. `debugger;` produces `breakpointSteps` the host can navigate to
7. The operator conformance sweep (`src/__tests__/operator-conformance.test.ts`) is extended to this engine, taking its expected values by **running the transpiler fresh** rather than re-reading the spec table — pinning the two engines to each other rather than to two transcriptions that could drift. See that file's header.
8. Side-by-side against the legacy CSE tab on the corpus: same steps, same frames, same arrows

## Risks

- **Fidelity round-trip**, above. The single biggest one.
- **Snapshot volume.** A whole run's snapshots are sent as one `CseSnapshotMessage` and must be structured-clone-able. Worth measuring on a program at the step limit before assuming it is fine.
- **Snapshot volume** and **`debugger;` handling** — the latter is now decided, see below.

## `debugger;` on the transpiler evaluator — decided

Measured current behaviour, so the change is against a known baseline:

| `executionMethod` | outcome | resolved engine |
| --- | --- | --- |
| `native` (what slice 1 pins) | runs to completion, no error, no warning | `native` — the statement is silently ignored |
| `cse-machine` | runs | `cse-machine` |
| `auto` (the legacy default) | runs | **`cse-machine`** — silently switched by `determineExecutionMethod` |

So both existing behaviours are invisible to the user: the legacy path silently *changes engine* when it sees a `debugger;`, and slice 1 silently *ignores* it.

**Decision: raise a runtime warning that points at the CSE machine evaluator**, rather than switching engines behind the user's back or staying mute.

- Emit on the **stdout** channel (`sendOutput`), not `sendError`. The program is not wrong and runs fine; this is advice, and the host renders the error channel in red. This matches how `ErrorSeverity.WARNING` is already routed in `src/conductor/errors.ts`.
- Once per run, not once per `debugger;` statement.
- Word it so it names the evaluator as the dropdown shows it, so the user knows what to click.

**Gate it on there being something to point at.** The CSE evaluator exists only for §3/§4, so at §1/§2 the hint would send the user somewhere that does not exist. There, either say only that the statement has no effect, or say nothing — decide when writing the wording, but do not emit the same message at every chapter.

**Implementation note:** detection needs the parsed AST, and `runFilesInContext` does not return one. Mirror `determineExecutionMethod`'s own check (`src/runner/utils.ts:52-58`, a `simple` walk for `DebuggerStatement`) rather than matching source text, which would fire inside strings and comments. A separate parse pass for a playground-sized program is cheap; if it errors, skip the check silently and let the real run report the syntax error.

## Reference

- py-slang's `src/conductor/plugins/PyCseMachinePlugin.ts` (506 lines) — the same job for a different language
- #2004's `JSCseEvaluator.ts` — useful for *which fields the renderer actually reads*, but see the caveats above

martin-henz:
Closes #2060.

Adds `SourceCseEvaluator3` and `SourceCseEvaluator4`, so the CSE tab works for Conductor-run Source. Builds on #2058.

## Scope

§3/§4 only — the frontend gates the tab at `chapter >= Chapter.SOURCE_3` (`ApplicationTypes.ts:245`) and, under Conductor, shows it when the language offers an evaluator carrying `EvaluatorCapability.CSE`. The evaluator is hidden from the dropdown and auto-selected when the tab opens, like the stepper.

Out of scope, unchanged: modules in the CSE tab (#2057), stepper, data visualizer, autocomplete, the `language-directory` cutover.

## The machine stays synchronous

In js-slang the CSE machine is the *visualiser*; the transpiler is the workhorse. So unlike py-slang — whose CSE machine is its primary engine and is `async` throughout — nothing here needs to await the host. `generateCSEMachineStateStream` is a plain `function*` and is used as one.

## Only the serialiser is new

The plugin trio in [`source-academy/plugins`](https://github.com/source-academy/plugins) is 264 LOC of pure transport; the renderer stays in the frontend. For Source that makes this a **round trip** — serialise js-slang state, then have the host's `CseSnapshotAdapter` rebuild js-slang-shaped objects for the same renderer the non-conductor tab uses. So the adapter, not the protocol's type definitions, is the real contract. Three conventions it depends on, all easy to get silently wrong, are documented at the top of `serialize.ts`: control and stash are serialised **top-first**, `InstrType` is a *string* enum passed through untouched, and `ENVIRONMENT` instructions are matched on `displayText` rather than on `instrType`.

Deliberately unlike #2004's serialiser:

- **arrays serialise to full depth**, with cycle detection, instead of truncating at depth 2 — which silently flattens ordinary Source list programs
- **steps are not deduplicated** by rendered text. Two steps can look identical in control and stash while differing in the environments, and collapsing them also desynchronises the step counter the user scrubs through from the machine's own numbering
- **closures send the machine's own rendering** as `displayValue`, so the stash reads identically to the non-conductor tab

## `debugger;` now says something

Both previous behaviours were invisible: the legacy `'auto'` method silently *switched engines* on seeing `debugger;`, and slice 1's pinned `'native'` silently ignores it. It now emits a note on **stdout** — not the error channel, since the program is not wrong — pointing at the CSE evaluator. Only at §3/§4, where one exists. Detection walks the AST, so `debugger` inside a string does not trigger it.

## Tests

**86 files, 4481 tests** (from 83 / 3846). `eslint` clean.

`collectSnapshots.test.ts` compares the serialised sequence against a **raw CSE machine run** over the same program — step count, control depth per step, stash contents — rather than snapshotting the serialiser's own output, which would only confirm it still does whatever it did. It earned its keep while being written: closures were sending a truncated name where the machine renders full source. Fixed in the serialiser, not by relaxing the assertion.

`operator-conformance-cse.test.ts` (610 cases) sweeps operator × type × type through the CSE machine and requires it to match the transpiler, taking expected values by **running the transpiler fresh** rather than reading the spec tables a second time — pinning the two engines to each other instead of to two transcriptions that could drift while both still looked right. Measured spread at §3 is 211 errors to 64 successes, so both branches are genuinely exercised rather than everything erroring.

`SourceEvaluator.test.ts` gives slice 1's evaluator its first unit tests against a stub `IRunnerPlugin`.

## Verified live, and what is not

The CSE tab **renders, steps and scrubs** — 24 steps for a closure program (`adder`/`add2`), with the slider, environment frames and control all working.

Two divergences are documented in `SourceCseEvaluator.ts` rather than papered over:

1. **The empty list.** The adapter maps the label `null` to a Python `None` stand-in, which is wrong for Source and also suppresses the empty-list visual. Source's `null` is therefore labelled `empty_list`, which the adapter does not yet know, so it falls through to the string fallback. Rendering it correctly needs a small frontend change.
2. **The global frame.** js-slang puts every builtin in the global environment's `head`, so that frame legitimately carries ~40 bindings; py-slang's is empty (its builtins live in `nativeStorage`), and the adapter was written against *that* shape — it injects a `:::pre-declared names::` sentinel binding, which is spurious here.

**What I could not confirm:** whether the non-conductor tab renders the global frame identically. The deployment pins `conductor.enable`, so a side-by-side was not possible from this machine. Until someone compares the two directly, treat the global frame as the one part of this evaluator whose parity is *unverified* rather than established — everything else above was checked against the raw machine or seen working in the browser.


martin-henz:
@codex review

martin-henz:
Thanks — all five were valid, and two of them were genuinely broken rather than merely untidy. Fixed in 382c0059.

**1. `stepIndex` collision** — confirmed: `steps += 1` happens *before* the yield (`interpreter.ts:390`), so the first yield reports 1 and `steps - 1` gave 0, colliding with the step-0 snapshot pushed ahead of the loop. Every run had two snapshots numbered 0 and every breakpoint step pointed one position behind. Now uses `steps`, which also makes `stepIndex` equal the array position.

**2. Gutter breakpoints ignored** — correct, and an outright omission on my part: #2060 specified `breakpointLines` and I only implemented `stepLimit`. Now threaded through; any node starting on a breakpoint line is reported, deduplicated by node identity so a breakpoint that stays on top across several steps is reported once rather than at every step.

**3. Duplicate runtime errors** — confirmed: `handleRuntimeError` (`cse-machine/utils.ts:486`) both pushes onto `context.errors` and throws. The fallback now fires only when the thrown value was not already drained.

**4. Parse polluting the shared context** — correct. Worth noting my existing test asserted `errors.length > 0`, which passed happily with two copies; it now asserts exactly 1.

**5. Array reachability** — reasoning accepted and fixed: an array carries its own `environment` and may hold closures, and `serializeValue` emits both ids, so the walk skipping arrays could leave a referenced frame unresolvable. It now follows array environments and elements with cycle protection.

One caveat I want to be straight about on **5**: I could not construct a program where the old code demonstrably lost a frame. In the cases I tried — including `function f() { const n = 1; return [() => n]; } f();` — the frame stayed reachable through `context.runtime.environments` regardless. So the fix is defensive, and its test is an *invariant* check (every referenced frame id resolves within its snapshot) rather than a regression repro. The test comment says so rather than implying more coverage than exists.

Findings 1–4 are each covered by tests verified by mutation: reverting any one of those four fixes fails 9 tests between them.

Suite is 87 files / 4515 tests. One unrelated flake to flag: `cse-machine-errors.test.ts > Infinite recursion with function calls in argument` is timing-sensitive (20s budget, ~16.6s typical) and tipped over once under full-suite load. It passes in isolation both with and without this branch, so it is not a regression here, but the added tests may have nudged the contention — worth watching in CI.

Issue #2062, opened 2026-09-21
Remove non-Conductor legacy surface (after the frontend migration)


martin-henz:
## Goal

Step 3 of the legacy removal: frontend first (source-academy/frontend#4397), then py-slang, then here.

An important scoping point up front, because "remove non-Conductor legacy" reads more drastically than it is: **js-slang's core is not legacy.** The Conductor evaluators added in #2058 and #2061 are built directly on `runFilesInContext`, `createContext`, the transpiler and the CSE machine. None of that goes anywhere.

What is legacy is the surface that exists *because the frontend evaluated Source itself*, and the module system that predates Conductor.

## Prerequisite

frontend#4397 — until the frontend stops importing `js-slang`, none of this can be removed. It currently imports 91 files' worth, most of it exactly the surface below.

## Candidates

**The legacy module system.** `src/modules/loader/requireProvider.ts` hands Source modules a fake `require('js-slang')` exposing the interpreter, parser, stdlib, errors and `utils/operators` — a module can reach straight into js-slang's internals. Conductor modules use `IDataHandler` instead, with no access to the host language's guts. Once modules move (#2062 territory / the dual-mode transpilation work), this and the loader/importer machinery around it go.

Note #2052 lives here too: `importers.ts` picks its import strategy by testing for `window`, which a Worker lacks. If the legacy loader is being deleted anyway, that bug may resolve by deletion rather than by fix — worth checking before spending effort on it.

**Editor-support APIs.** `src/name-extractor`, `src/finder`, `src/scope-refactoring`, `src/editors/ace/*` and `SourceDocumentation` exist for the frontend's editor: autocomplete, go-to-declaration, scope refactoring, syntax highlighting. Under Conductor these are the autocomplete plugin's job (`@sourceacademy/runner-autocomplete`), served per-evaluator. The frontend imports `js-slang/dist/editors/ace/modes/source` in three places today.

**`Variant.NATIVE`.** Reachable from nothing but the REPL's `transpiler` subcommand, and only three references repo-wide. Its relationship to `Chapter.FULL_JS` is the naming trap documented at length in #2051. If the native/full-JS evaluator family (#2056) is built, this should collapse into it rather than survive as a second way to reach `transpileToFullJS`.

## Explicitly *not* legacy

- **The REPL** (`src/repl`). A genuinely useful CLI, unrelated to the frontend.
[...]

PR #2064, opened 2026-09-21
feat(conductor): substitution stepper for Source §1–§2

### LINKED ISSUE #2063: Conductor migration, slice 3: substitution stepper for Source §1–§2
## Background

Slice 3 of the Conductor migration, after #2051 (transpiler evaluators, merged) and #2060 (CSE machine, #2061). Adds the substitution stepper, so the Stepper tab works for Conductor-run Source.

This is the **cheapest** of the three visualizers, for a reason worth stating: js-slang's stepper already produces estree-shaped nodes, and `@sourceacademy/common-stepper`'s host renderer *defaults to the Source/JavaScript syntax profile*. So unlike py-slang — which had to ship a whole `SyntaxProfile` describing Python's surface syntax — js-slang needs none at all. `BaseStepperRunnerPlugin.getSyntaxProfile()` returns `undefined` by default, and that default is exactly right here.

## Scope

**In scope**
- `SourceStepperEvaluator1` and `SourceStepperEvaluator2`, carrying `EvaluatorCapability.STEPPER`
- Serialisation of js-slang's `IStepperPropContents[]` into `SerializedStepperStep[]`
- Honouring the host's `stepLimit`

**Out of scope**
- §3/§4: the stepper is the substitution model, gated at `chapter <= 2` in the frontend today (`Playground.tsx:521`), matching py-slang's `PyStepperEvaluator1`/`2`
- A `SyntaxProfile` — not needed, see above
- Modules, the `language-directory` cutover

## Chapters

`chapter <= 2`. §3 introduces mutation and the substitution model stops being the right explanation, which is why the legacy tab draws the line there too.

## What has to be written

`@sourceacademy/runner-stepper` does the protocol work: channel wiring, the host's replay request, error reporting, caching the last steps. A language extends `BaseStepperRunnerPlugin<TAst>` and implements one method, `getSteps(ast) => SerializedStepperStep[]`, then the evaluator calls `sendSteps(ast)`.

So the work is a **serialiser**, again:

- `IStepperPropContents { ast, markers? }` -> `SerializedStepperStep { ast, markers? }`
- js-slang's stepper nodes are class instances (`StepperBinaryExpression` and friends) whose data lives in own enumerable fields and whose methods live on the prototype — so the walk keeps own properties, recurses into child nodes and arrays, and drops nothing else
- every node needs a stable `nodeId` within its step's tree
- `Marker.redex` points at a node **by object identity**, which does not survive the channel. It has to become `redexId` (that node's assigned `nodeId`) plus `redexNodeType`, per `SerializedMarker`

## Acceptance criteria

1. Step count and per-step explanations match js-slang's own `getSteps` for a corpus of §1/§2 programs — same approach as the CSE parity suite: compare against the real thing, do not snapshot the serialiser's own output
2. Every `redexId` in a step resolves to a `nodeId` present in that step's tree
3. Serialised steps are structured-clone-able: no functions, no class instances, no cycles
4. The host's `stepLimit` is honoured (py-slang shipped a hardcoded 1000 and had to fix it — source-academy/py-slang#427)
5. The Stepper tab renders in the frontend, with explanations, for a §1 program
6. No `SyntaxProfile` is sent, and the host's default renderer produces correct Source syntax

## Risks

- **Marker resolution.** Markers reference nodes by identity; if a redex is not part of the step's own tree, that has to be handled deliberately rather than emitting a dangling id.
- **Bundle size.** The stepper pulls in its own node classes and `astring`; worth checking the bundle stays comparable to the others.

martin-henz:
Closes #2063.

Adds `SourceStepperEvaluator1` and `SourceStepperEvaluator2`, so the Stepper tab works for Conductor-run Source.

## §1/§2 only

The stepper explains evaluation *as substitution*, which stops being the right account once §3 introduces mutation. The legacy tab draws the same line (`chapter <= 2`, `Playground.tsx:521`), as does py-slang with `PyStepperEvaluator1`/`2`.

Carries `EvaluatorCapability.STEPPER`, so the host hides it from the evaluator dropdown and selects it automatically when the Stepper tab opens.

## No SyntaxProfile — and that is the whole point

`@sourceacademy/common-stepper`'s host renderer **defaults to the Source/JavaScript profile**. js-slang's stepper already produces estree-shaped nodes, so it gets correct rendering by supplying *nothing*: `BaseStepperRunnerPlugin.getSyntaxProfile()` returns `undefined` and that default is exactly right.

py-slang has to ship a full profile — templates for every node type, token classes, associativity hints — only because Python's surface syntax is not the host's default. This is the one place where being the host renderer's native language pays off directly, and it is why this slice is much smaller than the CSE one.

## What is actually new

A serialiser, again. `runner-stepper` owns the channel wiring, the host's replay request, error reporting and step caching; a language implements one method, `getSteps(ast)`.

Two things do not survive a `MessageChannel`, and are what the serialiser exists for:

- **Methods.** The stepper AST is class instances (`StepperBinaryExpression` and friends) whose data lives in own enumerable fields and whose behaviour lives on the prototype — so copying own properties keeps exactly the data and drops exactly the methods.
- **Object identity.** A `Marker` points at its redex *by reference*. The protocol replaces that with `redexId`, referring to a `nodeId` assigned during the walk. A redex that is not part of its step's own tree yields `redexId: null` rather than a dangling id the host cannot resolve.

The host's `stepLimit` is read per run rather than fixed at construction — py-slang shipped a hardcoded 1000 and had to fix it (source-academy/py-slang#427).

## Tests

25 cases over a six-program corpus (§1 arithmetic, conditional, function application, arrow functions, higher-order; §2 lists).

**Parity** against js-slang's own `getSteps` rather than a snapshot of the serialiser's output: step count, per-step node type, marker count, and explanations.

**Three protocol invariants** over the same corpus:
- every `redexId` resolves to a `nodeId` present in that step's tree
- every node carries a unique `nodeId`
- the steps survive `structuredClone` — no functions, no class instances, no cycles

Suite: **88 files, 4540 tests**. `eslint` clean. All eight evaluator bundles build (`SourceStepperEvaluator*` at ~1.7 MB, comparable to the others).

## Not yet done

**Not verified in the browser.** The CSE slice was; this one has only been checked against js-slang's own stepper output and the protocol invariants. Worth a look at the Stepper tab before merging, particularly that explanations render and the default renderer really does produce correct Source syntax (criteria 5 and 6 in #2063).


martin-henz:
@codex review

martin-henz:
## Verified in the browser

Criteria 5 and 6 of #2063 are now met — this was the gap I flagged when opening the PR.

Setup: stepper web plugin built from `source-academy/plugins` and served on :4002 (which is where `plugin-directory` already points it), evaluator bundles on :4001, a local language directory on :8125 with `source1Stepper`/`source2Stepper` carrying `EvaluatorCapability.STEPPER`, and `plugin-directory` on :8126.

Program:

```js
function adder(n) {
  return x => x + n;
}
const add2 = adder(2);
add2(5);
```

**What works:**

- The Stepper tab appears for Source §1 and the stepper evaluator is auto-selected — it is correctly absent from the evaluator dropdown, as `STEPPER` capability implies.
- **12 steps**, slider and the four navigation controls all functional.
- **Step 1** renders the whole program with correct Source syntax *and* syntax highlighting — `function adder(n) {`, `return x => x + n;`, `const add2 = adder(2);` — with the explanation "Start of evaluation". **This is the no-`SyntaxProfile` claim confirmed live**: the host's default Source/JavaScript renderer produces correct output with js-slang supplying nothing.
- **Step 7** shows `add2(5);` with the redex **highlighted**, explained as "Constant add2 declared and substituted into the rest of block". Redex highlighting is the part I was least sure of, since it depends on `redexId` resolving against a `nodeId` assigned during serialisation — it resolves end to end.
- **Step 12** reaches `7;` with "Evaluation complete". `adder(2)` then `add2(5)` is 7, so the substitution semantics are right, not just the rendering.

Also incidentally exercised: `BaseStepperRunnerPlugin`'s replay path. Editing the program without re-running leaves the previously computed steps displayed, which is the host requesting `__lastSteps` rather than a stale render.

Nothing left outstanding from the issue's criteria.

martin-henz:
Both valid, fixed in 00162055 — though they differ a lot in how badly they bite, and I want to be straight about which is which.

## 1. Imports — worse than reported

Confirmed, and it does not degrade into "incorrect or stuck steps": it **throws**. `getSteps` dies with `this.body[0].contractEmpty is not a function`, which my evaluator catches and forwards, so the user sees an internal-looking message with no hint that imports are the cause.

On the suggested fix: running `preprocessFileImports` is what the non-conductor path does, but doing it here would pull the whole module pipeline into this slice — including the loader that #2052 shows is broken in a Worker — and that is #2062's work, not stepping. So instead import/export declarations are detected up front and reported as plainly unsupported, with the offending line. A clear "not supported" beats a stack-trace artefact, and it does not pretend to a capability that is not there.

Mutation-verified: disabling the guard fails its test.

## 2. Shared node identity — real, but defensive

The premise checks out. `StepperIdentifier.substitute` does insert the *same* object at every occurrence: for `(x => x + x)(1)` one `Literal` sits in two positions, for `(x => x * x * x)(2)` in three. Keying the id map by object identity kept only the last, so a marker pointing at such a node would resolve to whichever occurrence was walked last.

Each object now maps to all of its ids, and a marker expands to one per occurrence — which is what the reference stepper achieves by holding a node reference the renderer matches by identity.

**But I could not reproduce the symptom.** I instrumented every marker across those programs and measured how many positions each marker's redex occupies: it is **always exactly one**. The shared objects are real; markers just never point at them. So while the `postRedex` reasoning is sound, the "all resulting after-markers resolve to the last occurrence" outcome does not occur in practice on anything I could construct.

I have kept the fix — it is strictly more faithful and changes nothing in the cases that actually arise — but its tests pin invariants (ids unique per position, marker count matching occurrence count, every `redexId` resolving) rather than demonstrating a failure of the old behaviour, and the test comment says exactly that rather than implying more.

If you have a program where a marker *does* target a multiply-substituted node, that would turn this from defensive into demonstrated, and I would add it as a regression test.

---

Suite: 89 files, 4552 tests. `eslint` clean. The browser verification in the comment above still holds — it was run against the same stepping path.

PR #2080, opened 2026-09-21
fix(conductor): don't crash the error reporter on a non-SourceError

martin-henz:
Found while surveying the ground for module support.

`context.errors` is typed `SourceError[]`, but js-slang does not always keep that promise. The CSE machine's `evaluateImports` (`src/cse-machine/interpreter.ts:227`) reads `context.nativeStorage.loadedModules[moduleName]` — `undefined` under Conductor, because nothing loads modules yet — and the resulting raw `TypeError` is pushed straight onto `context.errors`.

`toConductorError` then called `explain()` on it and threw **from inside the error reporter**. On `SourceCseEvaluator.evaluateChunk`'s catch path that replaced the student's diagnostic with `error.explain is not a function` and lost the original failure entirely, so a §3 program containing

```js
import { black } from "rune";
```

reported nothing useful at all.

## The fix

One guard, falling back to `unknownToConductorError`, which already handles an arbitrary thrown value:

```ts
if (typeof error?.explain !== 'function') {
  return unknownToConductorError(error);
}
```

This is **hardening, not module support** — with it, the same program now reports `Cannot read properties of undefined (reading 'black')`, which is honest but still not a good message. Making imports actually work is tracked separately.

## Testing

New `src/conductor/__tests__/errors.test.ts`. The two raw-value cases both fail without the guard (verified by stashing it), and the `SourceError` cases pin the existing position-preserving behaviour so the guard can't quietly swallow a real diagnostic.

I also tried adding `?.` to `isWarning` and **removed it again**: `reportErrors` calls `isWarning` first, but a raw `Error` passes through it without trouble (`severity` is just `undefined`), so the optional chaining would have been unverifiable churn. Only `toConductorError` actually crashed.

Note for whoever runs `yarn tsc --noEmit -p tsconfig.json` locally: `src/utils/__tests__/rttc.test.ts:399` has a pre-existing type error on `master`, unrelated to this change. The prod build (`tsconfig.prod.json`) excludes tests and is clean.


martin-henz:
@codex review

Issue #2081, opened 2026-09-21
Conductor: module support for Source (design)


martin-henz:
Source programs cannot import modules under Conductor. This is the last large gap in the migration, and the one thing a §2 student hits immediately — `rune`, `curve`, `sound` and `pix_n_flix` are the whole point of SICP §2 in this curriculum.

This issue is the design, for review before any of it is built, following the same pattern as #2051. **No decision here is settled**; two of them I'd particularly like a second opinion on, marked **Decision** below.

## Where we actually are

Measured, not assumed — I ran each evaluator against `import { black } from "rune"; black;`:

| Evaluator | Today |
| --- | --- |
| `SourceEvaluator` (transpiler) | `EvaluatorError: Unable to get modules.` — the legacy loader tries to fetch the bundle over HTTP and fails |
| `SourceCseEvaluator` | raw `TypeError: Cannot read properties of undefined (reading 'black')` — `evaluateImports` reads `nativeStorage.loadedModules['rune']`, which nothing populates. The reporter used to crash on this too; #2080 fixes that half |
| `SourceStepperEvaluator` | rejected up front with an `EvaluatorSyntaxError` — deliberate, and honest |

**The host side is done.** The frontend already resolves and loads Conductor module plugins (`src/features/conductor/createConductor.ts`, `flagDirectoryModulesUrl.ts`), and py-slang already imports from them in production. Twelve bundles are ported (`conductor-modules.json`). So everything below is evaluator-side work in this repo, with no frontend changes and no protocol changes.

## What the protocol requires

A module arrives as an `IModulePlugin` from `ModuleLoaderRunnerPlugin.requestModule(name)`, carrying `exports: IModuleExport[]` — `{ symbol, value: TypedValue, signature? }`. To register that plugin the evaluator must be an `IInterfacableEvaluator`, i.e. `IEvaluator & IDataHandler`.

Two facts about `IDataHandler` set the whole shape of the work:

1. **Compound values cross the boundary as opaque identifiers, not live references.** `TypedValue<PAIR>.value` is a `PairIdentifier`; closures and arrays likewise. A module never touches a js-slang value directly — it asks the evaluator, through `pair_head`, `array_get`, `closure_call`.
[...]

PR #2082, opened 2026-09-21
feat(conductor): add SourceDataHandler, js-slang's module data interface

martin-henz:
First slice of #2081, following the decision there to use **private identifier tables** rather than backing handles with live Source values.

A Conductor module never holds a live reference into the evaluator. Pairs, arrays, closures and opaques all cross the boundary as opaque identifiers, and the module comes back through `IDataHandler` to read or write them. `SourceDataHandler` is the table those identifiers index into.

**Nothing registers it yet**, so the eight evaluator bundles are byte-for-byte unchanged (verified). The async spine that lets a Source program actually *call* a module function is the next slice, and the hard one.

## Three decisions worth reviewing

**Every method is `async`, including the ones that compute synchronously.** `IDataHandler` declares them `Promise`-returning, and without `async` a failed lookup throws *synchronously* out of a call the caller is entitled to hold as a promise — `handler.pair_head(p).catch(...)` would never see it. I hit this in the tests, where `.rejects` didn't fire. `require-await` is disabled for the file with that rationale inline; writing `Promise.reject()` by hand at each throw site would satisfy the rule and lose the guarantee at the first site anyone forgets.

**The accepted shapes are more forgiving than the types suggest, on purpose.** The module bundles are *shared with py-slang* — the same `rune`, `sound` and `midi` builds serve both languages — so this handler has to accept what py-slang's accepts or a bundle that works from Python breaks from Source. Concretely: `pair_head`/`pair_tail`/`pair_sethead`/`pair_settail`/`pair_assert` work on a two-element `ARRAY` as well as a genuine `PAIR`, the list helpers accept a flat `ARRAY` as well as a `PAIR`/`EMPTY_LIST` chain, and a declared `PAIR` parameter is satisfied by an `ARRAY`. Both tolerances are copied from py-slang rather than re-derived.

**`tie`/`untie` are no-ops rather than throws.** Our tables hold strong references for the handler's lifetime, so a tie is already guaranteed and an untie cannot be honoured. py-slang throws `Method not implemented`, which would abort a student's program over a call asking for something we already provide. The honest cost is that identifiers are never reclaimed within a run — hence `reset()`.

## One divergence from py-slang that is not a style choice

`accumulate` here applies **right-to-left with the element first**, matching Conductor's own reference implementation (`conductor/src/conductor/stdlib/list/accumulate.ts`):

```ts
for (let i = vec.length - 1; i >= 0; --i) {
    result = yield* this.closure_call(op, [vec[i], result], resultType);
}
```

py-slang's goes left-to-right and passes `[acc, element]`. For a commutative op that is invisible; for anything else the two languages give different answers from the same shared bundle. I'll file that against py-slang separately — flagging it here so this file isn't "corrected" into agreement with the wrong one.

## Errors

`src/conductor/modules/errors.ts` — `RuntimeSourceError` subclasses, so they reach the host through `toConductorError` with the rest of js-slang's diagnostics rather than being flattened by `unknownToConductorError`. They report in Source's vocabulary (`a pair`, `a function`, `null`), not Conductor's. None carry a node: the transpiled program has no AST at runtime, so there is no call site to point at, and an absent position beats a wrong one.

This is also why I'd now push back on extracting a shared handler package: py-slang's `GenericDataHandler` calls itself engine-agnostic, but it carries Python type names (`'str'`, `'NoneType'`), a `variant >= 3` gate and py-slang's own `ExprNS.Call`. The identifier bookkeeping is shareable; the diagnostics are not. Worth revisiting once both implementations exist and the genuinely common part is visible — the way the CSE and stepper plugins were extracted *after* two implementations, not before.

## Testing

32 tests. I mutation-tested them — ten mutants, all killed:

| Mutant | |
|---|---|
| `accumulate` left-to-right | killed |
| no PAIR↔ARRAY tolerance in signatures | killed |
| proxy methods returned unbound | killed |
| no array-as-pair view | killed |
| `array_set` skips its type check | killed |
| list helpers reject arrays | killed |
| closure arity unchecked | killed |
| immutable opaque becomes updatable | killed |
| `reset` keeps tables | killed |
| array bounds unchecked | killed |

The proxy-binding one is the subtle case: a bare `Reflect.get` hands back an unbound method, so calling it through `asInterfacableEvaluator` would set `this` to the proxy and `this.uniqueId++` would write to the *evaluator*, silently issuing duplicate identifiers that alias each other.

Full suite: 4,296 passing. All 8 bundles build.


martin-henz:
@codex review

martin-henz:
Addressed all three Codex findings in 77861f94. Each confirmed real by mutation testing (the corresponding new/changed test fails without its fix):

**P1 — identifier recycling across `reset()`.** Correct: `reset()` cleared every table but also reset `uniqueId` to 0, so a module holding a stale handle across a reset (the "still-playing sound" case the method's own doc comment says must survive) could silently alias the new run's unrelated value instead of raising `InvalidIdentifierError`. Fixed by never resetting the counter — it now stays monotonic for the handler's whole lifetime.

**P2 — `pair_sethead`/`pair_settail` bypassing `array_set`'s type check on an array-backed pair.** Correct: these were a second, unchecked write path into a typed array. Fixed by enforcing the same `array.type` check `array_set` already does.

**P2 — `ARRAY` accepting a genuine `PAIR` in `typesCompatible`.** Correct: `resolveArray`/`array_get`/`array_set` only ever consult `arrayMap`, so a non-array-backed `PAIR` would pass the signature check and then throw `InvalidIdentifierError` on the first actual array operation, deep in the module's own code. Removed that direction; the still-valid `PAIR`/`LIST`-accepts-`ARRAY` direction (needed for py-slang-style bundles) is untouched.

Full suite passing (4,301 tests), all 8 evaluator bundles build.

PR #2083, opened 2026-09-21
feat(conductor): module support for Source — the async spine

martin-henz:
Stacked on #2082 (`SourceDataHandler`) — base is `feat/module-data-handler`, not `master`, so review here is just this slice's diff. Second slice of #2081's design.

This is the hard one from that design doc: it makes `import` actually work in `SourceEvaluator`. A Source program can now import a Conductor module, call its exported functions, and have a module call back into a Source-defined closure — with proper tail calls preserved across the await boundary.

## The async trampoline (`src/utils/operators.ts`)

A module closure crosses the boundary as an `ExternCallable` — always Promise-returning by protocol, whether or not a given call actually suspends. `callIfFuncAndRightArgsAsync` mirrors `callIfFuncAndRightArgs` line for line (same argument validation, same trampoline, same error wrapping) with one addition: a call whose result is *actually* a thenable gets awaited, and the wall-clock time spent suspended is excluded from the infinite-recursion budget — real host latency isn't the student's program "recursing".

It also adds `MAX_ASYNC_CALL_DEPTH`, a stand-in for the native call-stack limit a sync non-tail recursion gets for free (a clean, catchable `RangeError`). An async non-tail recursion has no such limit — it grows a chain of heap-allocated promise reaction records instead — and this **crashed the whole test worker outright** during development, well before any timeout fired. Not hypothetical; it's how I found the need for the guard.

## Dual-mode transpilation (`src/transpiler/transpiler.ts`)

`transpile(..., isAsync)`: every call routes through `await callIfFuncAndRightArgsAsync`, every function becomes `async`, and the chunk runs inside an async IIFE. `hasImports(program)` detects the direct case; `SourceEvaluator` forces it on for every subsequent chunk once any chunk has ever imported.

The genuinely hard part: the sync engine gets "the value of the program" for free from `eval()`'s own completion-value tracking, which an async IIFE's ordinary `return` doesn't provide. `transformStatementsToTrackCompletionValue` reproduces that propagation by hand — and it is **not** "carry the previous value through unless you see an expression statement". Checked against real `eval()`, this caught two genuine spec subtleties I'd have gotten wrong by intuition:

- a `for`/`while` loop's completion resets to `undefined` on zero iterations, not "keep whatever ran before it"
- an `if` whose taken branch itself has an empty completion (ends in a bare declaration) **also** resets to `undefined` — it does not carry the value from before the `if`, even though a bare declaration on its own, outside any `if`, does carry the previous value through

Both are easy to get backwards since they look like the same "empty completion, do nothing" case at a glance. The test suite (`completionValue.test.ts`) checks the transform's output against `eval()` directly for exactly this reason.

Import bindings are deliberately kept **outside** the async IIFE, in the same outer eval-chainable scope sync mode already uses — they're plain property reads that never need `await`, since the module they name already finished loading before the chunk was transpiled. This is specifically what lets a later chunk with no import of its own still use a name an earlier chunk imported (verified in `SourceEvaluator.modules.test.ts`).

## Module interop (`src/conductor/modules/`)

- **`moduleInterop.ts`** — `sourceToModule`/`moduleToSource`. Considerably more direct than py-slang's equivalent: Conductor's data model *is* Source's data model (designed from SICP), so a pair is already a two-element array and the empty list is already `null` — no tagged union to build or reinterpret. The one real subtlety: a proper cons chain (`list(1,2,3)` is `[1,[2,[3,null]]]`) has to be **flattened** before crossing, because `SourceDataHandler` reads a `DataType.ARRAY`'s elements flat with no recursion. Mirrors py-slang's `isProperList`/`flattenProperList` — I found this the hard way, via a failing test, before adding it.
- **`conductorBundleImporter.ts`** — plugs into the *existing* legacy preprocessing pipeline (`preprocessFileImports`) as its `sourceBundleImporter`, reusing import-graph parsing, export-name validation and chapter gating unmodified. Also supplies a permissive `manifestImporter`: the legacy resolver needs a name→metadata manifest fetched over HTTP before it'll recognize *any* import as valid, and Conductor has no equivalent static manifest — so every name is treated as valid and metadata-free, deferring real validation to where it belongs, `ModuleLoaderRunnerPlugin.requestModule` itself. I found this the hard way too: `sourceBundleImporter` alone produced "Unable to get modules." because the resolver never got that far.

## SourceEvaluator wiring

Registers `SourceDataHandler` + `ModuleLoaderRunnerPlugin` per instance, wires the importers into `runFilesInContext`, and tracks `hasEverLoadedAModule` to force async mode on every chunk from the first import onward — deliberately coarser than a real reachability analysis, traded for simplicity (see #2081).

**Known limitation**, documented in the class doc and #2081: a name an async-mode chunk's *own code* declares — not an import, those persist fine — does not survive into a later, separate chunk, because that code runs inside the async IIFE's own function scope, invisible to a later chunk's separate `eval()` call.

**Not in this slice**: the CSE machine (#2057, tracked separately — harder, since it's otherwise synchronous), and the `.sync` escape hatch for hot per-call module closures (sound, pix_n_flix) — `SourceDataHandler.closure_call_sync` already supports it generically, nothing here populates it yet, every module call in this slice goes through the full async path.

## Testing

126 new tests across 6 files, several rounds of mutation testing. Two things worth knowing about, since both were caught by the tests rather than by review:

- An early version of the depth guard had `nativeStorage.asyncCallDepth++` with no fallback. `undefined + 1` is `NaN`, and every comparison against `NaN` is `false` — so a `NativeStorage`-shaped test double missing that field silently defeated the guard forever, and the resulting unbounded recursion crashed the test worker. Fixed with `?? 0` in the *real* code, not just the test double.
- `SourceEvaluator.modules.test.ts` is genuinely end-to-end: a fake `ModuleLoaderRunnerPlugin.instance` drives a real evaluator through an import, a module closure call, a module calling back into a Source closure, and the cross-chunk persistence case. Built by capturing the `IDataHandler` proxy the evaluator registers with the fake conductor, since a `TypedValue` identifier is only valid against the exact handler instance that allocated it — a value built by an unrelated `SourceDataHandler` would be rejected as dangling.

Full suite: 4,296+ tests passing, verified in isolation (the shared machine was under heavy, unrelated load during part of this session — several worker-pool timeouts on files I never touched, e.g. `cse-machine-utils.test.ts`, turned out to be pure resource contention, confirmed by clean fast reruns in isolation). All 8 evaluator bundles build. One snapshot diff: the new `callIfFuncAndRightArgsAsync` operator declaration, added unconditionally to every chunk's transpiled output (same pattern every other transpiler-internal binding here already follows) regardless of whether that specific chunk uses it.


martin-henz:
@codex review

PR #2088, opened 2026-09-22
fix: resolve all 13 tsc --noEmit errors (#2086)

### LINKED ISSUE #2086: tsc --noEmit reports 14 type errors in src/conductor + tests, invisible to CI
## Summary

`yarn tsc --noEmit -p .` (the full `tsconfig.json`, no excludes) currently reports 13 type errors across 6 files on `master`. None of them are caught by any CI step today, so they've been silently accumulating:

- `yarn build` runs `tsc --project tsconfig.prod.json`, but that config excludes both `src/conductor/**` and every `src/**/__tests__/**` — exactly where all of these live.
- `yarn build:evaluators` bundles the conductor evaluators (which do include `src/conductor/modules/moduleInterop.ts`) via `rollup-plugin-esbuild`, which transpiles but never type-checks.
- `yarn test-coverage` also runs through Vite/esbuild, with no type-checking.
- `yarn eslint` isn't a substitute — it doesn't surface general TS assignability errors like these.

Found while reviewing #2083; unrelated to that PR's own changes (confirmed present identically with and without it). Filing as its own issue since fixing it is out of scope there.

## Repro

```
yarn tsc --noEmit -p .
```

### A. `wrap()`/`wrapUnsafe()` overload resolution fails for some callback shapes

`HasCorrectParameters`/`CheckTupleHelper` in `src/utils/typeUtils.ts` infer `never` for certain valid callbacks, breaking `wrap()`'s overloads. Not conductor-specific — it also fires in `src/utils/__tests__/rttc.test.ts`, unrelated to conductor entirely.

- `src/utils/__tests__/callIfFuncAndRightArgsAsync.test.ts:35` — TS2345, `Argument of type '[]' is not assignable to parameter of type 'never'`
- `src/utils/__tests__/callIfFuncAndRightArgsAsync.test.ts:57` — TS2769, no overload matches
- `src/utils/__tests__/callIfFuncAndRightArgsAsync.test.ts:75` — TS2769, no overload matches
- `src/utils/__tests__/rttc.test.ts:399` — TS2344, `Type '(a: unknown, ...args: unknown[]) => unknown' does not satisfy the constraint '"Expected: function, Actual: never"'`

### B. `SourceDataHandler.array_set`/`array_get`'s untyped (`DataType.VOID`) overload doesn't unify against a specifically-typed array argument

This one also hits **production code**, not just tests:

- `src/conductor/modules/moduleInterop.ts:118` — TS2352, `Conversion of type 'ITypedValue<DataType.ARRAY, DataType.ANY>' to type 'ITypedValue<DataType.ARRAY, DataType.VOID>' may be a mistake`
- `src/conductor/modules/__tests__/moduleInterop.test.ts:153,154` — TS2345, `ITypedValue<DataType.ARRAY, DataType.NUMBER>` not assignable to `ITypedValue<DataType.ARRAY, DataType.VOID>`
- `src/conductor/modules/__tests__/SourceDataHandler.test.ts:62,114,120` — same TS2345 shape

### C. `InvalidArrayCreationError` (`src/conductor/modules/errors.ts:120-121`) has a real naming collision, not just a lint nit

```ts
export class InvalidArrayCreationError extends ModuleInterfaceError {
  constructor(private readonly type: string) {
```

`ModuleInterfaceError` → `RuntimeSourceError` → `SourceError` already declares a public `readonly type: ErrorType` (`src/errors/base.ts:19,61`) for error classification. This constructor parameter shadows it with an unrelated `string`, which is almost certainly a naming accident (it's meant to hold the `DataType` name for the message text) rather than an intentional override:

- `src/conductor/modules/errors.ts:120` — TS2415, `Property 'type' is private in type 'InvalidArrayCreationError' but not in type 'ModuleInterfaceError'`
- `src/conductor/modules/errors.ts:121` — TS4115, missing `override` modifier

Likely fix: rename the constructor parameter (e.g. `dataTypeName`) rather than adding `override`.

### D. One test-only cast

- `src/conductor/modules/__tests__/moduleInterop.test.ts:202` — TS2352, `ITypedValue<DataType.ARRAY, DataType>` not comparable to `ITypedValue<DataType.PAIR, DataType> | ITypedValue<DataType.EMPTY_LIST, DataType>`

## Suggested remediation

Two independent things worth doing:
1. Fix the underlying type issues (C in particular looks like a genuine bug, not just a suppressible type error).
2. Add a CI step that actually runs `tsc --noEmit` over the full project (conductor + tests included), so a regression like this gets caught going forward instead of silently landing.

martin-henz:
Closes #2086.

`yarn tsc --noEmit -p .` (the full `tsconfig.json`, no excludes) currently reports 13 type errors on `master`, none of which are caught by any CI step:
- `yarn build` runs `tsc --project tsconfig.prod.json`, which excludes both `src/conductor/**` and every `src/**/__tests__/**` — exactly where all of these live.
- `yarn build:evaluators` bundles the conductor evaluators via `rollup-plugin-esbuild`, which transpiles but never type-checks.
- `yarn test-coverage` also runs through Vite/esbuild, with no type-checking.
- `yarn eslint` isn't a substitute — it doesn't surface general TS assignability errors like these.

So they'd been accumulating silently. This PR fixes all 13, grouped by root cause:

## `SourceDataHandler.array_set`'s overloads were declared in the wrong order

The non-generic, `DataType.VOID`-only signature was declared *first* (making it the only one TypeScript exposes to callers), while the actually-generic, per-element-type signature was the *implementation* signature — never itself part of the public overload set. `array_get` right above it has the correct order (generic overload declared, VOID-specific shape folded into the implementation); `array_set` had them backwards. Swapped it to expose both the generic overload and a VOID-permissive one, matching `array_get`'s working pattern. This also let a same-line `as TypedValue<DataType.ARRAY, DataType.VOID>` cast in `moduleInterop.ts` be removed entirely — the generic overload now infers correctly without it.

## `InvalidArrayCreationError`'s constructor parameter shadowed the base class's `type` field

```ts
export class InvalidArrayCreationError extends ModuleInterfaceError {
  constructor(private readonly type: string) {
```

`ModuleInterfaceError` → `RuntimeSourceError` → `SourceError` already declares a public `readonly type: ErrorType` for error classification (`src/errors/base.ts`). This constructor parameter shadowed it with an unrelated string — a real naming collision, not just a suppressible type error. Renamed to `dataTypeName`.

## `rttc.ts`'s `FunctionOfLength<T>` used the wrong tuple shape

`isFunctionOfLength`'s runtime check is `f.length === l`, and JS's own `Function.length` already excludes a trailing rest parameter — so a value that passes this guard may genuinely accept *more* than `l` arguments. The file already had the correct rest-args type definition, just commented out with a `// TODO: Should it be the "proper" signature?`, while the stricter, wrong one was active — and the file's own test (`expectTypeOf`) was written against the commented-out one. Swapped it in.

## Three `wrap()` test call sites were themselves wrong, not `HasCorrectParameters`

- `callWithoutMetadataAsync(42 as never)` — casting the deliberately-wrong value to `never` (rather than `any`) makes the generic `Parameters<T>` also collapse to `never`, so TS rejects the call's own argument list against that — a generic-inference artifact of testing a non-function value, not a real bug in the wrapper. Changed to `as any`.
- Two `wrap(fn, 1, name)` calls where `fn` has exactly one *required* parameter and zero optional ones — `optArgCount` counts optional parameters, not total parameter count (see `wrap`'s own doc comment). Should be `0`.

## Two remaining test-only mismatches

- `SourceDataHandler.test.ts`: a test deliberately calls `array_set` with a mismatched element type to exercise the runtime `InvalidTypeError` check — now that `array_set`'s overloads are correct, TS correctly refuses this call at compile time too, so the test needs an explicit `as any` bypass (same pattern the `42 as never` fix above uses, just `any` this time since a concrete `TypedValue` value is still needed for the runtime assertion).
- `moduleInterop.test.ts`: `list_to_vec`'s declared parameter is `TypedValue<DataType.LIST>`, but it also accepts a flat `ARRAY` at runtime (documented on the function itself) — exactly what a flattened proper list crosses as. Fixed with the double-cast-through-`unknown` TS itself suggests.

## Testing

- `yarn tsc --noEmit -p .` — clean, 0 errors (was 13)
- `yarn vitest run` — 4,359 tests passing, 0 regressions
- `yarn eslint` — 0 errors (same pre-existing warning set as before)
- `yarn format:ci` — clean
- `yarn build` and `yarn build:evaluators` — both succeed


martin-henz:
@codex review

PR #2089, opened 2026-09-22
feat(conductor): wire up the data visualizer, so draw_data works again

### LINKED ISSUE #2078: Conductor: wire up the data visualizer, so `draw_data` works again
`draw_data` does not work in any of the Conductor evaluators. All three wire the `visualiseList` hook to a throw:

```ts
// src/conductor/SourceEvaluator.ts:49, SourceCseEvaluator.ts:77, SourceStepperEvaluator.ts:52
visualiseList: () => {
  throw new Error('draw_data is not supported by this evaluator.');
},
```

So a §2 program calling `draw_data(list(1, 2, 3))` — which every SICP §2 exercise set reaches for — fails at runtime. This was a known gap when slice 1 landed (#2051); I listed it as a follow-up there and never filed the issue. This is that issue.

## What it takes

py-slang already did this, and the host half is shared and published, so js-slang's share is genuinely small — one adapter class and one value converter.

Add to `package.json`:

```
"@sourceacademy/common-data-visualizer": "^0.0.1",
"@sourceacademy/runner-data-visualizer": "^0.0.1",
```

Then, mirroring `PyDataVisualizerRunnerPlugin.ts` / `toDataVisualizerNode.ts`:

- **`src/conductor/dataVisualizer/SourceDataVisualizerRunnerPlugin.ts`** — extends `BaseDataVisualizerRunnerPlugin<Value>`, implements the single abstract `toNode(value, refs)`.
- **`src/conductor/dataVisualizer/toDataVisualizerNode.ts`** — the whole language-specific part.
- In each evaluator's constructor, replacing the throw:

```ts
this.dataVisualizerPlugin = conductor.registerPlugin(SourceDataVisualizerRunnerPlugin);
conductor.hostLoadPlugin(DATA_VISUALIZER_DIRECTORY_ID);
```

and `visualiseList` becomes `(...values) => this.dataVisualizerPlugin?.sendDrawing(values)`.

Cycle detection, pair/list/tree classification and rendering are **not** our problem — they live host-side in `web-data-visualizer`, shared across every language, because pairs and lists do share structure across Source Academy's SICP-descended languages (unlike the stepper, whose per-language `getSteps` necessarily does the whole job). The adapter is a mechanical dispatch on value shape.

## The value mapping

js-slang's runtime values are plain JavaScript, so this is more direct than py-slang's tagged `Value` union:

| js-slang value | node |
|---|---|
| `null` (the empty list) | `{ type: 'empty' }` |
| any array | `{ type: 'array', refId, children }` |
| `Closure`, or a JS function (builtin) | `{ type: 'function', refId, displayValue: stringify(v) }` |
| everything else | `{ type: 'leaf', displayValue: stringify(v), label }` |

`refs.get(value)` returns `alreadySeen` on the second encounter of a reference; emit `{ type: 'ref', refId }` and stop recursing. That is what makes cyclic structures terminate.

Two details:

- **A pair is a 2-element array**, and so is a §3 `[1, 2]`. The host distinguishes "pair" from "list" by `children.length`, not by tag, so a two-element §3 array will be drawn as a pair. The pre-Conductor frontend did exactly the same thing, so this is fidelity, not a regression — but it should be stated in a comment so nobody "fixes" it later.
- **`displayValue` should come from `stringify`**, matching what `src/conductor/cse/serialize.ts` already does for closures in the CSE snapshot, so the same value reads the same way in both tabs.

## A real bug to fix on the way

`draw_data` is declared variadic but only ever forwards its first argument:

```ts
// src/createContext.ts:283
const visualise_list = (v0: Value, ..._v: Value[]) => {
  (externalBuiltIns.visualiseList ?? defaultBuiltIns.visualiseList)(v0, context.externalContext);
  return v0;
};
```

`draw_data(xs, ys)` silently drops `ys`. The plugin's `sendDrawing(values: TValue[])` takes the whole argument list and the host renders each argument as its own Konva `Stage`, so the fix is to stop discarding `..._v`. Note the `RefIdAllocator` is created **per argument**, not per call — a value shared between two arguments is serialized in full in each, and a `refId` must never be compared across arguments.

## Gating

`draw_data` is defined at `chapter >= 2` (`src/createContext.ts:345`), so register the plugin for §2+ only. `SourceStepperEvaluator` covers §1 and §2, so it needs the same conditional, not an unconditional registration.

Also call `resetRun()` at the start of each chunk, alongside the evaluator's other per-run resets — the Conductor-era equivalent of the old frontend's `DataVisualizer.clearWithData()` on every Run press. Without it, rows accumulate across REPL entries.

## Testing

py-slang has `src/conductor/dataVisualizer/__tests__/toDataVisualizerNode.test.ts`; the same shape applies here, and it is the part worth testing, since the plugin class is a two-line adapter. Cover at minimum: the empty list, a pair, a long list, a cyclic structure (`const x = pair(1, null); set_tail(x, x);`), a structure shared between two arguments, and a closure.

Then check it in the playground against the frontend's data visualizer tab.

## Out of scope

The frontend's renderer. `plugin-directory` already resolves `data-visualizer` to `https://source-academy.github.io/plugins/web/data-visualizer/index.mjs`, and py-slang already exercises it, so nothing should be needed there — but that assumption is worth confirming early rather than at the end.

martin-henz:
Closes #2078.

All three Conductor evaluators wired `visualiseList` to a throw (`"draw_data is not supported by this evaluator."`), so any §2+ program calling `draw_data` — which every SICP §2 exercise set reaches for — failed at runtime.

## What this adds

Following py-slang's `PyDataVisualizerRunnerPlugin`/`toDataVisualizerNode` shape, and the host-side plugin pair (`@sourceacademy/runner-data-visualizer` + `@sourceacademy/common-data-visualizer`), which is already published and shared across languages:

- **`src/conductor/dataVisualizer/toDataVisualizerNode.ts`** — converts a js-slang runtime value into the shared `SerializedDataVisualizerNode` wire format. Purely mechanical: `null` → `empty`, an `Array` → an `array` node recursing into elements, `typeof value === 'function'` → a `function` node, everything else → a `leaf`. No cycle-detection or pair/list classification here — that's entirely the host's job. A CSE machine `Closure` is itself `typeof 'function'` (its constructor swaps `Closure.prototype` onto a real `Function` object — see `cse-machine/closure.ts`'s `Callable`), so the same branch covers both the transpiler's bare functions and the CSE machine's `Closure`s with no `instanceof` needed — considerably more direct than py-slang's tagged-union dispatch.

- **`src/conductor/dataVisualizer/SourceDataVisualizerRunnerPlugin.ts`** — the two-line adapter `BaseDataVisualizerRunnerPlugin` expects.

- **`SourceEvaluator`/`SourceCseEvaluator`/`SourceStepperEvaluator`** now register the plugin (§2+ only — `SourceCseEvaluator` unconditionally, since it only ever instantiates for §3/§4; the other two gate on chapter), call `hostLoadPlugin(DATA_VISUALIZER_DIRECTORY_ID)`, route `visualiseList` through `sendDrawing`, and call `resetRun()` at the start of every chunk so rows don't accumulate across REPL entries — the Conductor-era equivalent of the old frontend's `DataVisualizer.clearWithData()` on every Run press.

  `SourceStepperEvaluator`'s wiring is present for parity with the other two, but currently **inert**: the stepper has its own, entirely separate symbolic `draw_data` (`src/stepper/builtins/lists.ts`) that just returns its first argument as an AST node and never calls `visualiseList` at all — the substitution model has no notion of "the host's data visualizer tab" to step through. Documented in a comment on the field rather than silently left looking functional; fixing it for real is a stepper-engine question, not evaluator wiring.

## A real bug fixed on the way

`draw_data` is declared variadic (`draw_data(x1, ...xs)`), but `visualise_list` only ever forwarded its first argument:

```ts
const visualise_list = (v0: Value, ..._v: Value[]) => {
  (externalBuiltIns.visualiseList ?? defaultBuiltIns.visualiseList)(v0, context.externalContext);
  ...
```

`draw_data(xs, ys)` silently dropped `ys`. `CustomBuiltIns.visualiseList` now receives the whole argument list (`Value[]`), and the plugin's `sendDrawing` renders each argument as its own Konva `Stage` on the host side.

## Testing

- `toDataVisualizerNode`'s own unit tests: leaves, the empty list, a pair, a longer array, a self-referential cycle, shared-but-acyclic structure, functions, and a plain-object fallback.
- An end-to-end test per evaluator (`SourceEvaluator`, `SourceCseEvaluator`, `SourceStepperEvaluator`), each driving a *real* `SourceDataVisualizerRunnerPlugin` instance over a minimal fake channel, verifying `draw_data(1, 2, 3)` actually reaches the host with every argument and that `resetRun()` prevents rows accumulating across chunks.
- A regression test for the dropped-argument bug in `testing.test.ts`.

Full suite passes (4,375 tests), `yarn tsc --noEmit -p .` is clean, 0 eslint errors, prettier clean, and both `yarn build` and `yarn build:evaluators` succeed.

## Not covered

The frontend's renderer — per #2078, `plugin-directory` already resolves `data-visualizer` and py-slang already exercises it, so nothing should be needed there, but I haven't verified it live in the playground (no browser available from here). Worth a manual check before merging.


martin-henz:
@codex review

PR #2090, opened 2026-09-22
feat(conductor): wire up the autocomplete plugin, generated from docs/lib jsdoc

### LINKED ISSUE #2079: Conductor: wire up the autocomplete plugin
There is no autocomplete in the Conductor Source evaluators. Typing `acc` in the §2 playground offers nothing, where the pre-Conductor editor offered `accumulate` with its documentation. Like #2078, this was a follow-up I listed in #2051 and never filed.

Unlike the data visualizer, this one has a **genuine open question**, because the data it needs was deleted.

## The plugin half is easy

py-slang's `src/conductor/plugins/autocomplete/` is the template:

```
index.ts             AutoCompletePlugin extends BaseAutoCompleteRunnerPlugin
                     + registerAutoCompletePlugin(conductor, variant)
keywords.ts          per-chapter keyword lists
mode.ts              SyntaxHighlightData for the editor
highlight-rules.ts
resolver.ts          getNames(tree, code, row, column, variant)
builtins/*.json      committed, one file per library
```

Registration is two lines, in every evaluator (py-slang calls it from all six of its own):

```ts
conductor.registerPlugin(AutoCompletePlugin, variant);
void conductor.hostLoadPlugin(WEB_PLUGIN_ID);
```

with `"@sourceacademy/common-autocomplete": "^0.0.1"` and `"@sourceacademy/runner-autocomplete": "^0.0.2"` added to `package.json`. js-slang already has acorn, so `resolver.ts` can walk a real ESTree program rather than py-slang's `@lezer/python` tree — scope-aware resolution is strictly easier here.

`mode.ts` / `highlight-rules.ts` may well be unnecessary. The stepper runner plugin already skips `getSyntaxProfile()` because the host's default renderer *is* the Source/JavaScript one; if the same holds for the autocomplete plugin's `mode`, js-slang gets correct highlighting by supplying nothing. Worth checking before porting 886 lines of Ace mode.

## The open question: where does the builtin data come from?

#2070 deleted exactly the machinery that produced it:

- `src/name-extractor/index.ts` (492 lines) — `getProgramNames`, `getKeywords`, `docsToHtml`, with 892 lines of tests
- `src/editors/ace/docTooltip/index.ts` — `SourceDocumentation`, importing eleven generated JSON files (`source_1.json`, … `External libraries.json`)
- `src/editors/ace/modes/source.ts` (886 lines) — the Ace mode
- `scripts/autocomplete.mjs` — the generator, which used **JSDOM to scrape `docs/source/*/global.html`**, i.e. the *rendered* jsdoc output, and wrote the JSON into `src/editors/ace/docTooltip/`

That last one is the awkward part, and it is why the deletion was right at the time: the generator only worked after `yarn docs` had run, which is what broke CI in #2068. Three candidates:

1. **Restore the scraper.** Cheapest, and reproduces the old tooltips exactly. But it keeps a build-order dependency on generated HTML that has already cost us one red CI, and the JSON would have to be committed (as py-slang commits `builtins/*.json`) rather than generated at build time.
2. **Generate from the jsdoc *sources*** in `docs/` instead of the rendered HTML. Removes the `yarn docs` ordering constraint and the `jsdom` dependency. More work up front; the parsing is against doclet comments we control.
3. **Move the data to the `plugins` repo**, next to `common-autocomplete`. Tempting for symmetry, but Source's builtin documentation belongs with Source, and per-chapter gating is js-slang's knowledge.

My inclination is **2** — the scrape was always the weak link, and this is the moment we are paying to rebuild the pipeline anyway. But I'd like a second opinion before committing to it, since 1 is a fraction of the work and the output is identical.

Whichever we pick, the JSON should be **committed**, matching py-slang, so a plain `yarn build:evaluators` on a clean checkout produces working bundles.

## Per-chapter gating

The old `SourceDocumentation.builtins` was keyed `'1'`, `'2'`, `'3'`, `'4'`, plus `_typed` and `4_explicit-control` variants. The typed variants are gone with `Chapter.FULL_TS` (#2070) and the variants are deferred (#2053–#2056), so §1–§4 is the whole matrix for now.

## Out of scope

**Module autocomplete.** The deleted `src/name-extractor/__tests__/modules.test.ts` covered completing names from `import { black } from "rune"`. That depends on module support landing first (#2062, source-academy/modules#975), so it should be a follow-up rather than a blocker here.

**Go-to-declaration and scope refactoring** (`src/finder.ts`, `src/scope-refactoring.ts`, also removed in #2070) are separate editor features with no plugin counterpart yet. Not this issue.

martin-henz:
Closes #2079.

There was no autocomplete in any Conductor Source evaluator: typing `acc` in the playground offered nothing, where the pre-Conductor editor offered `accumulate` with its documentation.

## Builtin data: mechanically generated from `docs/lib/*.js`, not scraped

Per the discussion on #2079, this uses **option 2**: generate from jsdoc *source* comments, not rendered HTML.

The old (deleted in #2070) pipeline scraped autocomplete tooltips from *rendered* jsdoc HTML, which only worked after `yarn docs` had already run — a build-order dependency that broke CI once already (#2068).

Following py-slang's own solution to the identical problem: `scripts/autocomplete.mjs` asks jsdoc for its own **parsed doclet AST** directly (`-X`, jsdoc's "explain" mode — a pure parse step, no rendering, no ordering requirement) against `docs/lib/*.js`, and writes the simplified `{name, title, description, meta}` shape straight to `src/conductor/plugins/autocomplete/builtins/*.json`. Output is committed (matching py-slang's `builtins/*.json`), so a clean `yarn build:evaluators` doesn't depend on this script having run — but CI (`nodejs.yml`) now runs `yarn autocomplete` and diffs the result, so a `docs/lib` edit that forgot to regenerate fails loudly instead of silently going stale.

Filters out `auxiliary.js` (internal `__access_export__`-style loader plumbing, never student-facing) and anything with no jsdoc comment (`list.js`'s own undocumented `$accumulate`/`$append`/`$build_list` helpers that the real, documented functions wrap).

## The plugin itself

- **`resolver.ts`** — walks a real ESTree program (via `acorn-walk`'s `fullAncestor`) to build the scope chain around the cursor, subsequence-matches against it plus chapter-gated keywords and the jsdoc-derived builtins, and scores inner scopes ahead of outer ones. Parses with `acorn-loose`, not the real parser: the buffer is normally *invalid* Source while a student is mid-keystroke, and completions need to survive that.
- **`highlight-rules.ts`** — adapted from ace-builds' own `javascript_highlight_rules.js` (by way of the pre-Conductor `src/editors/ace/modes/source.ts`), trimmed to what Source's smaller grammar actually needs: no regex literals (`no-unspecified-literal` rejects them, so `/` is always division and the `regex`/`no_regex` state split ace needs for that ambiguity collapses away), no template literals, JSX, classes, generators or `var`. The keyword-mapper's `support.function`/`constant.language` buckets are derived from the same builtins JSON the autocomplete entries use — one source of truth, not a hand-maintained duplicate list (which is exactly how the pre-Conductor version of this file drifted enough to need per-chapter manual upkeep).
- **`mode.ts`** — hooks `foldingRules`/`indents`/`outdents`/`autoOutdent` into ace's own existing `javascript`/`cstyle` modules (Source's block structure is JavaScript's, unchanged) rather than reimplementing them; only `highlightRules` is genuinely Source-specific data.
- **`index.ts`** — the two-line-registration adapter, mirroring py-slang's `AutoCompletePlugin`/`registerAutoCompletePlugin` shape.

Registered unconditionally in all three evaluators (`SourceEvaluator`, `SourceCseEvaluator`, `SourceStepperEvaluator`) — unlike the data visualizer (#2089), autocomplete is useful at every chapter including §1.

## Testing

- `resolver.ts`: local-scope resolution (inner/outer/sibling visibility, scoring, parameter scopes), chapter-gated keywords and builtins, subsequence matching, and tolerance for invalid mid-edit syntax.
- `highlight-rules.ts`: every produced regex compiles, every `next` target state exists, the keyword mapper is chapter-gated.
- `index.ts`: an end-to-end `AutoCompletePlugin` driving the real `mode`/`autocomplete` methods.
- A registration-reaches-every-evaluator check per evaluator test file.

Full suite passes (4,402 tests), 0 eslint errors, prettier clean, and both `yarn build` and `yarn build:evaluators` succeed — with the plugin and its builtin data confirmed present in the bundle.

## Not covered

The frontend's rendering of the highlight rules / autocomplete dropdown itself — no browser available from here, same caveat as #2089's data visualizer. Worth a manual check in the playground before merging.


martin-henz:
@codex review

Issue #250, opened 2019-06-12
Primitive function to be added to Source §3: set_timeout


martin-henz:
As in:
```quests/AY1819S1/quest-Q5B.xml:`set_timeout(f, t)`: Calls the function f with no arguments after waiting for t milliseconds.```

This is needed for textbook section 3.3.4.

martin-henz:
No need for set_timeout in Section 3.3.4. The section uses simulated time.

PR #2091, opened 2026-09-22
feat: add set_timeout(f, t) / clear_all_timeout() builtins

### LINKED ISSUE #2025: Add set_timeout(f, t) / clear_all_timeout() as language builtins (mirroring py-slang#311)
## Summary

Add `set_timeout(f, t)` / `clear_all_timeout()` as language builtins, mirroring what `py-slang` already did for Python (source-academy/py-slang#311). This is what unblocks retiring the equivalent hand-rolled reimplementation in the Source Academy `modules` repo's `matrix` bundle (formerly `sound_matrix`, Quest Q5B - "The Magical Tone Matrix") - see source-academy/modules#831, which found a real bug in that reimplementation that py-slang's own builtin already handles correctly (see below).

There's a related, older, *closed* issue - #250, "Primitive function to be added to Source §3: set_timeout" - from 2019, for the same quest. That one was closed because the specific textbook section it was raised for ended up using simulated time instead and didn't need it. This request is different: it's not textbook-driven, it's about consolidating a capability that already needs to exist at the language level (a module already reimplements it today, with a bug), the same call Prof Martin made when this got moved out of `sound_matrix` and into `py-slang` itself.

## Why this belongs in the language, not a module

`sound_matrix`/`matrix`'s own module-level `set_timeout` reimplementation (`source-academy/modules`, `src/bundles/matrix/src/index.ts`) has a real bug: its scheduled callback does `void drainGenerator(evaluator.closure_call_unchecked(f, []))` with no `.catch`. If the scheduled closure throws a genuine runtime error, that becomes a silently unhandled promise rejection inside a bare `setTimeout` callback, instead of being reported to the student the way a normal runtime error would be.

py-slang's own implementation (added for exactly this reason - see its `set_timeout` doc comment: *"moved into the language itself rather than staying module-gated, since it's a capability of the evaluator, not of any one module"*) already gets this right:

```ts
// py-slang, src/engines/py2js/runtime.ts
set_timeout: this.builtin("set_timeout", 2, (f, delay) => {
  ...
  const id = setTimeout(() => {
    this.pendingTimeouts.delete(id);
    this.acall(f, [])
      .catch((e: unknown) => {
        const line = e instanceof Py2JsRuntimeError ? `${e.pyKind}: ${e.message}`
          : e instanceof Error ? `${e.name}: ${e.message}`
          : String(e);
        this.output.push(line + "\n");
        this.onOutput?.(line);
      });
  }, Number(delay));
  this.pendingTimeouts.add(id);
  this.onPendingWorkChange?.(1);
  return null;
}),
```

- Errors from `f` are caught and reported through `onOutput`, "matching how a real browser reports an uncaught async error to the console rather than nowhere" (the implementation's own words) - not silently lost.
- `onPendingWorkChange`/`beginPendingWork()`/`endPendingWork()` keep the host from tearing down the runtime while a timer is still pending (source-academy/py-slang#329 documents what goes wrong without this: the callback gets silently killed mid-flight).
- `clear_all_timeout()` cancels every still-pending timer, matching the counts back down.

A single, correct implementation living in the language once (like this) is strictly better than every module that wants scheduling reimplementing it - and getting it wrong - on its own.

## Scope: the transpiler evaluator only, for now

**Update:** an earlier version of this issue claimed "js-slang has only the CSE machine as its evaluator - no Py2JS-style 'compile to a real JS closure' engine to fall back on the way py-slang does." That's no longer accurate (it may have been written before this distinction was as clear-cut as it is now). js-slang currently ships three evaluator families under Conductor:

- **`SourceEvaluator1..4`** (the transpiler) - the primary, default evaluator for all of Source §1-4. It compiles Source straight to real JS and runs it via `eval()` (`src/transpiler`). A Source closure here already *is* a plain, always-callable JS function once defined - exactly the same shape py-slang's Py2JS closures have, and exactly what makes py-slang's implementation above straightforward. `f` stays callable from a real `setTimeout` firing later, on the real JS event loop, with no re-entry machinery needed - the same `callWithoutMetadata`/`callWithoutMetadataAsync` trampolines already used elsewhere in this codebase (`src/utils/operators.ts`) can call it directly.
- **`SourceCseEvaluator3/4`** - a separate, secondary evaluator that only exists for the §3-4 CSE-machine visualization tab. This is the one with the actual "resume a stepped closure call from cold" problem the rest of this issue originally described (py-slang's own CSE machine has the identical gap, undocumented as a known limitation in its stdlib stub).
- **`SourceStepperEvaluator1/2`** - the §1-2 substitution-model stepper. Likely moot regardless, since `set_timeout` would presumably gate at §3+ (matching the historical context of #250).

**Decision: this issue is scoped to the transpiler evaluator only.** The CSE machine's version of this problem is real but is explicitly *not* part of this issue - leave `SourceCseEvaluator` unsupported (a clear, documented stub, matching py-slang's own precedent for its CSE machine) rather than blocking this on solving cold-resume in general. That remains a follow-up for whoever is closest to the CSE machine's architecture, potentially shared with py-slang's identical gap.

One thing worth keeping in mind for whoever implements this: builtins are registered once, in `createContext.ts`, shared by every evaluator - so `set_timeout` becoming globally registered would also make it *callable* from the CSE machine and stepper evaluators, where the scheduled closure would not actually work correctly once that run's own state is gone. The implementation needs some way to keep it transpiler-only (or otherwise be explicit that calling it elsewhere is unsupported), not just register it unconditionally the way most other builtins are.

## Suggested shape

- `set_timeout(f, t)`: schedules `f` (a no-argument closure) to be called after `t` milliseconds, without blocking the calling code. Returns nothing (`undefined`).
- `clear_all_timeout()`: cancels every timer scheduled by `set_timeout` that hasn't fired yet, no arguments, returns nothing.
- Builtins would register the same way existing ones do (`src/stdlib/misc.ts` for the implementation, `src/stdlib/index.ts` to expose the name), with `f` invoked later via `callWithoutMetadata`/`callWithoutMetadataAsync` (`src/utils/operators.ts`) rather than a raw call, so it goes through the same trampoline/error-wrapping every other Source call does.
- Errors thrown by `f` should be reported the way an uncaught runtime error normally is (console/REPL output), not silently swallowed - matching py-slang's `onOutput` approach.
- Some way to track pending timers and keep the host (`SourceEvaluator`/Conductor) from tearing down mid-flight - the js-slang equivalent of py-slang's `onPendingWorkChange`/`beginPendingWork()`/`endPendingWork()` (source-academy/py-slang#329 documents what goes wrong without this).

## References

- source-academy/py-slang#311 - the original py-slang feature request/implementation
- source-academy/py-slang#329 - the host-teardown race this needs to guard against
- source-academy/modules#831 - the bug in `matrix`'s own reimplementation this is meant to retire
- js-slang#250 (closed 2019) - an earlier, differently-motivated request for the same primitive

martin-henz:
## Summary

Closes #2025.

Adds `set_timeout(f, t)` / `clear_all_timeout()` as §3+ language builtins, mirroring py-slang's Py2JS implementation (source-academy/py-slang#311). This is what unblocks retiring the `matrix` module's own hand-rolled reimplementation (source-academy/modules#831), which drops errors from its scheduled callback on the floor instead of reporting them.

## Why this works cleanly here

A compiled Source function is already a plain JS closure (the transpiler evaluator, `SourceEvaluator`, compiles Source to real JS via `eval()`) — exactly like py-slang's Py2JS. So a real `setTimeout` firing later can just call it directly on the real event loop; there's no cold-re-entry problem to solve, unlike an interpreter loop that would need to resume from a torn-down state.

This PR corrected #2025's original scoping claim ("js-slang has only the CSE machine") before implementing — js-slang's *primary* evaluator is actually the transpiler, and the issue is scoped to it accordingly. See the updated issue body for the corrected evaluator inventory.

## Implementation

- `src/types.ts`: adds `setTimeout`/`clearAllTimeout` to `CustomBuiltIns`, following the same per-evaluator-hook pattern already used for `visualiseList` (#2078).
- `src/createContext.ts`: registers `set_timeout(f, t)`/`clear_all_timeout()` at chapter ≥ 3 (matching the historical closed issue #250), validates `f` as a zero-argument function via `assertFunctionOfLength`/`assertNumberWithinRange`, and delegates to `externalBuiltIns.setTimeout ?? defaultBuiltIns.setTimeout`. `defaultBuiltIns.setTimeout`/`clearAllTimeout` throw a clear "not supported by this evaluator" error — the fallback every evaluator gets unless it wires a real implementation.
- `src/conductor/SourceEvaluator.ts`: the real implementation. Schedules `f` with a real `setTimeout`, calls it through `callIfFuncAndRightArgsAsync` (so normal recursion/time-budget protection applies), reports a thrown error via `conductor.sendError` instead of dropping it, and wraps each scheduled call in `beginPendingWork()`/`endPendingWork()` — conductor's existing mechanism (its own doc comment cites "Python's set_timeout" as the motivating case) that keeps the host from tearing the evaluator down while a timer is still outstanding (source-academy/py-slang#329 documents the failure mode this prevents). `clear_all_timeout()` cancels every outstanding timer and matches the pending-work count back down.
- The CSE machine and stepper evaluators need **no changes** — they don't wire a `setTimeout` hook, so they automatically fall through to `defaultBuiltIns`'s throw.
- `src/utils/testing/index.ts`: the generic (non-Conductor) test harness records `set_timeout` calls into a new `setTimeoutResult` array rather than actually scheduling `f`, consistent with how the other hooks there work; a test that needs `f` to actually run can override via `testBuiltins`.

## Test plan

- [x] New tests in `SourceEvaluator.test.ts`: `f` runs after the delay without blocking `evaluateChunk`; an error thrown by `f` is reported (not swallowed); `clear_all_timeout()` cancels a pending timer; `set_timeout` is not declared below §3.
- [x] New test in `SourceCseEvaluator.test.ts`: `set_timeout` throws a clear "not supported" error on the CSE machine.
- [x] `yarn tsc --noEmit -p .`, `yarn vitest run` (full suite, snapshots updated for the two new global builtins), `yarn eslint`, `yarn format:ci`, `yarn build`, `yarn build:evaluators` all pass.


martin-henz:
@codex review

